Issue #2, opened 2022-01-31
Conda Package


madsbk:
We should create Conda packages, both regular and nightly and releases.

@quasiben and @jakirkham I am hope this is something you guys can help with? :)

quasiben:
@madsbk do you think it's necessary to build a libkvikio conda package as well as the Python package or is one large package enough ?

madsbk:
One conda package of the Python library should be enough. 
Notice, the C++ library is header-only so there isn't any compiled library such as `libkvikio.so`

jakirkham:
We build a C++ & Python package for RMM, which is also header only.

madsbk:
> We build a C++ & Python package for RMM, which is also header only.

You are right, we need a Conda package for the C++ header files to make it easy for projects like cuCIM and cuDF to use the C++ API.

PR #5, opened 2022-02-01
Add build scripts for gpuCI

charlesbluca:
This PR, along with some internal setup, should set up PR testing and nightly / release conda package builds on this repo:

- [x] Add conda build / upload scripts
- [x] Add conda recipe(s)
  - [x] kvikio
  - [x] libkvikio 
- [x] Add testing script

Fixes https://github.com/rapidsai/kvikio/issues/2

### DIFF
--- a/ci/cpu/build.sh
+++ b/ci/cpu/build.sh
@@ -64,8 +64,8 @@ conda config --set ssl_verify False
 # BUILD - Conda package builds
 ################################################################################
 
-gpuci_logger "Build conda pkg for libkvikio"
-gpuci_conda_retry build conda/recipes/libkvikio
+# gpuci_logger "Build conda pkg for libkvikio"
+# gpuci_conda_retry build conda/recipes/libkvikio
 
 gpuci_logger "Build conda pkg for kvikio"
 gpuci_conda_retry build conda/recipes/kvikio
--- a/ci/cpu/build.sh
+++ b/ci/cpu/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2018-2022, NVIDIA CORPORATION.
+# Copyright (c) 2022, NVIDIA CORPORATION.
 ##############################################
 # kvikIO CPU conda build script for CI       #
 ##############################################
--- a/ci/cpu/build.sh
+++ b/ci/cpu/build.sh
@@ -60,15 +60,18 @@ conda list --show-channel-urls
 # FIX Added to deal with Anancoda SSL verification issues during conda builds
 conda config --set ssl_verify False
 
+# FIXME: Remove
+gpuci_mamba_retry install -c conda-forge boa
+
 ################################################################################
 # BUILD - Conda package builds
 ################################################################################
 
 # gpuci_logger "Build conda pkg for libkvikio"
-# gpuci_conda_retry build conda/recipes/libkvikio
+# gpuci_conda_retry mambabuild conda/recipes/libkvikio
 
 gpuci_logger "Build conda pkg for kvikio"
-gpuci_conda_retry build conda/recipes/kvikio
+gpuci_conda_retry mambabuild conda/recipes/kvikio
 
 ################################################################################
 # UPLOAD - Conda packages
--- a/ci/cpu/build.sh
+++ b/ci/cpu/build.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+# Copyright (c) 2018-2022, NVIDIA CORPORATION.
+##############################################
+# kvikIO CPU conda build script for CI       #
+##############################################
+set -e
+
+# Set path and build parallel level
+export PATH=/opt/conda/bin:/usr/local/cuda/bin:$PATH
+export PARALLEL_LEVEL=${PARALLEL_LEVEL:-4}
+
+# Set home to the job's workspace
+export HOME="$WORKSPACE"
+
+# Determine CUDA release version
+export CUDA_REL=${CUDA_VERSION%.*}
+
+# Setup 'gpuci_conda_retry' for build retries (results in 2 total attempts)
+export GPUCI_CONDA_RETRY_MAX=1
+export GPUCI_CONDA_RETRY_SLEEP=30
+
+# Use Ninja to build, setup Conda Build Dir
+export CMAKE_GENERATOR="Ninja"
+export CONDA_BLD_DIR="$WORKSPACE/.conda-bld"
+
+# Switch to project root; also root of repo checkout
+cd "$WORKSPACE"
+
+# If nightly build, append current YYMMDD to version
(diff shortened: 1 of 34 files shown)


ci/cpu/build.sh
```diff
@@ -0,0 +1,75 @@
...
+export CMAKE_GENERATOR="Ninja"
+export CONDA_BLD_DIR="$WORKSPACE/.conda-bld"
+
+# Switch to project root; also root of repo checkout
+cd "$WORKSPACE"
+
+# If nightly build, append current YYMMDD to version
+if [[ "$BUILD_MODE" = "branch" && "$SOURCE_BRANCH" = branch-* ]] ; then
+  export VERSION_SUFFIX=`date +%y%m%d`
+fi
```
charlesbluca, line 32: @madsbk do we intend to do releases on RAPIDS' release cycle, or more/less frequently? Depending on that answer, we might want to either

- rename the main dev branch of this repo to `branch-22.04` and leave these lines
- keep the main branch as is and remove these lines along with some below

ci/cpu/build.sh
```diff
@@ -0,0 +1,75 @@
...
+env
+
+gpuci_logger "Activate conda env"
+. /opt/conda/etc/profile.d/conda.sh
+conda activate rapids
+
+# Remove rapidsai-nightly channel if we are building main branch
+if [ "$SOURCE_BRANCH" = "main" ]; then
+  conda config --system --remove channels rapidsai-nightly
+fi
```
charlesbluca, line 48: Could also remove this if we keep the `main` branch as is

ci/cpu/build.sh
madsbk: Yes, I think we should follow the RAPIDS' release cycle in addition to the nightly releases.

conda/recipes/kvikio/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  name: kvikio
+  version: {{ version }}
+
+source:
+  git_url: ../../..
+
+build:
+  number: {{ GIT_DESCRIBE_NUMBER }}
+  noarch: python
+  string: py_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```
charlesbluca, line 15: Think we need to create a tag (`v22.04.00a` perhaps?) for these variables to be populated

quasiben:
Thank you so much @charlesbluca!

ajschmidt8 (approved):

madsbk:
@charlesbluca It would be great if we can have a style check. The following checks everything:
```
python scripts/format-all.py --check
```

conda/recipes/kvikio/meta.yaml
quasiben: I think this has been addressed now

charlesbluca:
@madsbk sure! Would you mind if I shift over to doing the non-clang format checks using pre-commit, so we can have a style check similar to [that of cuDF](https://github.com/rapidsai/cudf/blob/branch-22.04/ci/checks/style.sh)?

madsbk:
Yes that would be great.

quasiben:
If we are following RAPIDS release schedules should we create a 22.04 branch now instead of main?

charlesbluca:
Yeah that makes sense - were we also planning to do an initial release of this package for 22.02? If so would it also make sense once we get packages working to make a `branch-22.02` branch / tag?

madsbk:
No need to rush, let 22.04 be our first release

ajschmidt8:
rerun tests

ajschmidt8:
rerun tests

python/benchmarks/single-node-io.py
```diff
@@ -361,7 +361,7 @@ def parse_directory(x):
         args.api = tuple(API.keys())
 
     # Create a temporary directory if user didn't specify a directory
-    temp_dir: Union[tempfile.TemporaryDirectory, contextlib.nullcontext]
+    temp_dir: Union[tempfile.TemporaryDirectory, ContextManager]
```
charlesbluca, line 364: mypy views `contextlib.nullcontext` as a function instead of a class and throws an error here - opted to use `ContextManager` instead to circumvent this, not sure if we can get more specific here

scripts/run-clang-format.py
```diff
@@ -22,7 +22,7 @@
 import sys
 import tempfile
 
-EXPECTED_VERSION = "10.0.0"
+EXPECTED_VERSION = "11.1.0"
```
charlesbluca, line 25: Noting this change because the version of `clang-format` in the gpuCI images is 11.1.0; are we okay to change this?

python/benchmarks/single-node-io.py
madsbk: `ContextManager` is fine, thanks

scripts/run-clang-format.py
madsbk: Yes, that is perfectly fine

madsbk (approved):

python/setup.cfg
```diff
@@ -48,4 +48,4 @@ skip=
 packages = find:
 install_requires =
     numpy
-python_requires = >=3.7,<3.9
+python_requires = >=3.8,<3.10
```
charlesbluca, line 51: gpuCI tests 3.8, 3.9 now - are we okay with bumping the `python_requires` of this package?

madsbk: Yes, I think 3.8+ is acceptable.

madsbk: Also, for now let's require CUDA v11.5+

quasiben: yes, i think that would be good

ci/gpu/build.sh
```diff
@@ -0,0 +1,82 @@
...
+################################################################################
+
+gpuci_logger "Check environment variables"
+env
+
+gpuci_logger "Check GPU usage"
+nvidia-smi
+
+gpuci_logger "Check GDS status"
+/usr/local/cuda/gds/tools/gdscheck.py -p
```
quasiben, line 44: I think this can be removed.  GDS won't be installed on the gpuCI VM

charlesbluca: Yeah I added this as a means of verifying that GDS wasn't installed on the system - will remove now.

> Everything else should pass because of the compatibility mode for cuFile

If that's the case then we might be running into some other issue to cause the failures being seen here:

https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/kvikio/job/prb/job/kvikio-gpu-test/CUDA=11.5,GPU_LABEL=driver-495,LINUX_VER=ubuntu20.04,PYTHON=3.9/2/

madsbk:
rerun tests

madsbk:
@charlesbluca thanks for working on this! With https://github.com/rapidsai/kvikio/pull/13 in, the CI now passes :)

There are still a lot to do regarding CI, but I think this is a great start!

PR #17, opened 2022-02-25
Don't use `CONDA_CHANNEL` for package builds

charlesbluca:
We aren't able to use `CONDA_CHANNEL` for kvikIO builds yet, so removed that variable from the builds scripts.

### DIFF
--- a/ci/cpu/upload.sh
+++ b/ci/cpu/upload.sh
@@ -14,9 +14,6 @@ export GPUCI_RETRY_SLEEP=30
 # Set default label options if they are not defined elsewhere
 export LABEL_OPTION=${LABEL_OPTION:-"--label main"}
 
-# Setup conda build dir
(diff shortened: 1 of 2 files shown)


ci/cpu/upload.sh
```diff
@@ -14,6 +14,9 @@ export GPUCI_RETRY_SLEEP=30
 # Set default label options if they are not defined elsewhere
 export LABEL_OPTION=${LABEL_OPTION:-"--label main"}
 
+# Setup conda build dir
+export CONDA_BLD_DIR="$WORKSPACE/.conda-bld"
+
```
ajschmidt8, line 19:
```suggestion
```

These lines can be removed. Since `CONDA_BLD_DIR` is `export`ed in the parent `ci/cpu/build.sh` script, it shouldn't be necessary to redefine it here.

Issue #11, opened 2022-02-23
[Bug] GDS doesn't work on WSL


gigony:
GDS doesn't work on WSL even for compatibility mode.

test.py
```python
import cupy
import kvikio

a = cupy.arange(100)
f = kvikio.CuFile("test-file", "w")
# Write whole array to file
f.write(a)
f.close()

b = cupy.empty_like(a)
f = kvikio.CuFile("test-file", "r")
# Read whole array from file
f.read(b)
assert all(a == b)
```

```bash
❯ python test.py
Assertion failure, file index :cufio-udev  line :143
[1]    13856 abort      python test.py
```

FYI, cuCIM handles the issue by checking platform and skip executing `cuFileDriverOpen()`.
- https://github.com/rapidsai/cucim/pull/106
- https://github.com/rapidsai/cucim/pull/106/files#diff-d82fd952d7f60aaef86557b8a3bf0ebf63e623abe68613996489a9503628bf11
- https://github.com/rapidsai/cucim/pull/106/files#diff-e884f8a3a339aca81d1fc593161b73f4ffa5c000778fd6cf2eff840830315dc8

## Related Issues or Pull Requests
- https://github.com/rapidsai/cudf/issues/9861
- https://github.com/rapidsai/cucim/pull/106

PR #21, opened 2022-02-28
Enable compat mode when runnning in WSL

madsbk:
Fixes https://github.com/rapidsai/kvikio/issues/11

@gigony can you confirm that this PR fixes the WSL issue?


gigony:
Hi @madsbk , I confirmed that this fix works on WSL!
Thank you for the fix!

madsbk:
@gigony thanks for checking!

PR #24, opened 2022-03-09
[Feature] Add Nvcomp python bindings to kvikio.

thomcom:
This PR adds the basic bindings for Cascaded and Lz4, and a broken Snappy binding. I need to figure out how to import nvcomp properly before this will work with kvikio.


quasiben:
Thanks @thomcom !  As part of this PR will you also include example notebooks and the benchmark plots you built earlier. The plots don't need to be committed in the PR but uploaded in the PR comments

thomcom:
Thanks @quasiben I've added the files, I'll move them around later.

thomcom:
This is basically ready for review. Tomorrow I will add the notebook, benchmarks, and get the style updated correctly.

quasiben:
> Do we need to do something to make sure nvCOMP is available on CI?

I don't believe so as nvcomp tooling is tested as part of cuDF/libcuDF

@thomcom looks like there are some style check failures

jakirkham:
> I don't believe so as nvcomp tooling is tested as part of cuDF/libcuDF

How is the nvCOMP binding code here tested?

jakirkham:
Thanks. Was more asking are we running those on gpuCI here? Relatedly are there any gpuCI changes we need to include here to run the tests.

thomcom:
Oh of course, I'll be looking into that today if I can affect it.

thomcom:
These CI failures are mysterious to me.

jakirkham:
Maybe try merging in `branch-22.04` or pushing another commit? Seems like CI failed to checkout the PR. So would just give it another try. If it still fails, we can ask for help :)

jakirkham:
Think [this CI error]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/kvikio/job/prb/job/kvikio-cpu-cuda-build/CUDA=11.5/48/console ) is capturing the issue that I'm wondering about

```
      building 'kvikio._lib.nvcomp' extension
      /usr/local/gcc9/bin/gcc -Wsign-compare -DNDEBUG -fwrapv -O3 -Wall -fPIC -O3 -I/workspace/.conda-bld/kvikio_1647375755170/work/cpp/include -I/workspace/.conda-bld/kvikio_1647375755170/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/include -I/usr/local/cuda/include -I/workspace/.conda-bld/kvikio_1647375755170/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/include/python3.8 -c kvikio/_lib/nvcomp.cpp -o build/temp.linux-x86_64-3.8/kvikio/_lib/nvcomp.o -std=c++17
      kvikio/_lib/nvcomp.cpp:776:10: fatal error: nvcomp.h: No such file or directory
        776 | #include "nvcomp.h"
            |          ^~~~~~~~~~
      compilation terminated.
```

Do we need to be cloning or downloading nvCOMP somewhere on CI first?

jakirkham:
Thinking we need to add [this code]( https://github.com/rapidsai/cudf/blob/cf936b6a199b7e91f398932ced6bcdc285f061d2/cpp/cmake/thirdparty/get_nvcomp.cmake ) and then call it [somewhere like this]( https://github.com/rapidsai/kvikio/blob/4d9f905aa3dbdad5f2e8e78f6305ef40d2791ba2/cpp/CMakeLists.txt#L84 ).

thomcom:
I think we're already depending on cudf in this, are we not? There's a piece in `setup.py` about loading the headers from NVCOMP_HOME if they aren't installed. I will look again.

jakirkham:
I don't think so. This is lower level than cuDF.

Thanks Thomson! 😄

jakirkham:
Glad to hear you discussed this.

Yeah was wondering about this. RMM recently added support for [scikit-build]( https://github.com/scikit-build/scikit-build ) ( https://github.com/rapidsai/rmm/pull/976 ), which might be one approach to this problem.

Another might be an `(lib)nvcomp` package, which we depend on.

vyasr:
I think Github is lying. I definitely beat @jakirkham to posting that :upside_down_face:

jakirkham:
Interesting what are the main challenges to implementing a scikit-build based process?

Yeah think I agree that a Conda package would be more medium term. So makes sense to table that for now.

Also curious how either of these compare to what you (Thomson) and Mads discussed earlier

vyasr:
@jakirkham my biggest concerns with scikit-build are the unknowns. It is still very much a beta project, and I found a number of issues while implementing it for RMM. While I would be hopeful that we have ironed out enough of them, RMM is header-only so the need to (statically) link to an additional project could potentially uncover additional complications or limitations that we would need to address. I am happy to give this a shot, and if it turns out to "just work" then that would be fantastic. I just don't want that adoption to become a blocker for this release since we're so close.

jakirkham:
Sure and to be clear I'm not pushing that or any of the other approaches. Am just trying to determine, which of these is most viable in the near term so we can proceed on next steps here.

vyasr:
I took a stab at trying to answer that question in #32. I can't seem to get builds of kvikio (C++ or Python) to run correctly (on my machine or a DGX), so I'm going to let CI tell us how viable scikit-build is going to be in the short term.

vyasr:
So... I guess it just worked :smiley: :tada: I guess we handled enough issues with the RMM conversion to make this relatively painless. So maybe we just roll with scikit-build (assuming everyone involved is on board).

thomcom:
What is the current status on you work @vyasr ? Do you have ideas for resolving the building of `nvcomp`, making its header available, and adding it to the Cython static library? Thanks!

thomcom:
To my great despair I found out that Nvcomp released v2.2 on Feb 7. They modified the API that is used in the nvcomp bindings thoroughly and a major refactor is required. I'm planning on scrambling to push this refactor as fast and hard as I can. 

Should we drop the idea of having this ready for 22.04 (since we're in burndown) or should we release with v2.1?

quasiben:
> Should we drop the idea of having this ready for 22.04 (since we're in burndown) or should we release with v2.1?

Do you know what version of nvComp cuDF is using for 22.04 and 22.06 ?  My guess is that we should follow cuDF

vyasr:
> What is the current status on your work @vyasr ? Do you have ideas for resolving the building of `nvcomp`, making its header available, and adding it to the Cython static library? Thanks!

I have ideas, yes. I think we would basically do what @jakirkham suggested here:

> Thinking we need to add [this code](https://github.com/rapidsai/cudf/blob/cf936b6a199b7e91f398932ced6bcdc285f061d2/cpp/cmake/thirdparty/get_nvcomp.cmake) and then call it [somewhere like this](https://github.com/rapidsai/kvikio/blob/4d9f905aa3dbdad5f2e8e78f6305ef40d2791ba2/cpp/CMakeLists.txt#L84).

but now that we're using scikit-build we could restrict that line of code to just the Python kvikio package rather than the libkvikio build. I'll test that out once #32 is completed (I'll ping you before I push any changes to this branch though).

jakirkham:
When you have a moment, can you please fix the merge conflicts here @thomcom? 🙂

python/setup.py
```diff
@@ -1,11 +1,155 @@
 # Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
-from setuptools import find_packages
-from skbuild import setup
```
jakirkham, line 5: Would try to prefer the changes in `branch-22.04` when resolving the conflicts here. In particular we need Vyas' changes ( https://github.com/rapidsai/kvikio/pull/32 ) like the lines here

vyasr: Honestly I would suggest checking out the version of `setup.py` from `branch-22.04` and building on that directly. AFAICT the only changes that should be made in this PR are to add nvcomp as a dependency for the Python library and to create the CMake files for the new nvcomp subpackage. @thomcom let me know if you'd like me to try to do that on this branch.

cpp/cmake/thirdparty/get_nvcomp.cmake
```diff
@@ -0,0 +1,40 @@
+# =============================================================================
```
jakirkham, line 1: To Vyas' point above, we can move this to the `python/cmake/thirdparty` directory (may need to create the `thirdparty` portion of that path)

Then we can add a similar `include` line here:

https://github.com/rapidsai/kvikio/blob/1ab41d6fe06f18e0e8996a4451307201bee4e787/python/cmake/CMakeLists.txt#L15

python/setup.py
thomcom: Grr I intended to take only the branch-22.04 version of the file, not sure what happened. Retry coming up.

cpp/cmake/thirdparty/get_nvcomp.cmake
thomcom: Great, will do shortly.

vyasr: Yup. Then I think the only piece that we'll need to add is a `target_link_libraries` call telling CMake to link `kvikio::nvcomp` against the C++ `nvcomp` library.

thomcom: :+1:

thomcom: I've got all of `skbuild` working locally, up to running the python and finding out that cuda has not been linked in properly to the libkvikio.so library. Trying to figure out which CMakeLists.txt and syntax to add it with.

vyasr: You need to add `nvcomp` to the [list of cython modules in `_lib/CMakeLists.txt`](https://github.com/rapidsai/kvikio/blob/branch-22.04/python/kvikio/_lib/CMakeLists.txt#L16) then I think you should be good.

python/kvikio/_lib/nvcomp Cascade Benchmarks.ipynb
```diff
@@ -0,0 +1,3672 @@
+{
```
jakirkham, line 1: Should we move this to `python/examples`?

thomcom: I was leaning toward moving it to `$KVIKIO_HOME/notebooks`, but I'm easily swayed.

jakirkham: Sure that could make sense

thomcom:
:+1: Today I'm updating the notebook once more and moving it to where it belongs, then producing benchmarks and sharing them here, then hoping this can fly like a phoenix.

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,235 @@
+import collections
+
+import cupy as cp
+import Enum
```
jakirkham, line 4: Appears to be causing [a CI build failure]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/kvikio/job/prb/job/kvikio-cpu-cuda-build/CUDA=11.5/95/console )

```suggestion
from enum import Enum
```

python/kvikio/_lib/pynvcomp.pyx
```diff
@@ -0,0 +1,429 @@
...
+# could be either __array_interface__ or __cuda_array_interface__
+cpdef __get_ptr(a):
+    # this has to be slow though... is there a better way? try/catch?
+    d = a.__dir__()
+    if '__cuda_array_interface__' in d:
+        return __get_cuda_array_interface_ptr(a)
+    elif '__array_interface__' in d:
+        return __get_array_interface_ptr(a)
+    else:
+        raise AttributeError('Argument does not implement __cuda_array_interface__ or __array_interface__')  # NOQA: E501
```
jakirkham, line 62: We could use [`Array`]( https://github.com/rapidsai/kvikio/blob/4a60dc0bf046023e8cbc7d9c39afc663bd407588/python/kvikio/_lib/arr.pyx#L66 ) to handle this. Would also handle computing `nbytes`. Though this could be follow on work after this PR

python/kvikio/__init__.py
```diff
@@ -1,8 +1,7 @@
 # Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
-
-from . import _version
+from . import _version, nvcomp  # noqa: F401
```
madsbk, line 4:
```suggestion
from . import _version
```
No need to import the module always.

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,235 @@
+import collections
```
madsbk, line 1:
```suggestion
# Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
# See file LICENSE for terms.

import collections
```

python/tests/test_nvcomp.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
+
```
madsbk, line 3:
```suggestion
```

python/tests/test_nvcomp.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.mark.parametrize("dtype", cudf.utils.dtypes.INTEGER_TYPES)
+@pytest.mark.parametrize("size", [int(1e6), int(1e7)])
+def test_cascade_lib_vs_module(dtype, size):
+    dtype = cupy.dtype(dtype)
+    data = cupy.array(np.arange(0, (size / dtype.itemsize) - 1), dtype=dtype)
+    compressor = nvcomp.CascadedCompressor(dtype)
+    compressed = compressor.compress(data)
+    print("module:", compressor.compress_out_size)
+    print("actual:", compressed.size)
```
madsbk, line 41:
```suggestion
```

madsbk (changes requested):
Thanks @thomcom, overall it looks great!

I only have some minor suggestions: 
- It would be good with some docstrings for all public facing functions and classes. It would also be good for non-trivial internal functions and classes. Use the NumPy docstring style like: https://github.com/rapidsai/kvikio/blob/branch-22.04/python/kvikio/cufile.py
- Remove the spaces in the notebook name

python/kvikio/_lib/pynvcomp.pyx
```diff
@@ -252,9 +237,9 @@ cdef class _LZ4Decompressor:
         del self.d
 
     cpdef configure(self, in_arr, in_bytes, temp_bytes, out_bytes, uintptr_t stream=0):
```
jakirkham, line 239:
```suggestion
    cpdef configure(self, in, temp_bytes, out_bytes, uintptr_t stream=0):
```

jakirkham (commented):
Just a suggestion, we might be able to simplify some of the interface code and handling within functions by tweaking how we use `Array`. Added a little example below with `in` in this `configure` function. Though we could do similar things with other arguments and functions

python/kvikio/_lib/pynvcomp.pyx
```diff
@@ -252,9 +237,9 @@ cdef class _LZ4Decompressor:
...
-        cdef uintptr_t in_ptr = __get_ptr(in_arr)
-        cdef uintptr_t temp_bytes_ptr = __get_ptr(temp_bytes)
-        cdef uintptr_t out_bytes_ptr = __get_ptr(out_bytes)
+        cdef uintptr_t in_ptr = Array(in_arr).ptr
+        cdef uintptr_t temp_bytes_ptr = Array(temp_bytes).ptr
+        cdef uintptr_t out_bytes_ptr = Array(out_bytes).ptr
         self.d.configure(
             <void*>in_ptr,
             <size_t>in_bytes,
```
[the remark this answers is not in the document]
thomcom: I'll try this out.

python/kvikio/_lib/pynvcomp.pyx
```diff
@@ -252,9 +237,9 @@ cdef class _LZ4Decompressor:
         del self.d
 
     cpdef configure(self, in_arr, in_bytes, temp_bytes, out_bytes, uintptr_t stream=0):
-        cdef uintptr_t in_ptr = __get_ptr(in_arr)
-        cdef uintptr_t temp_bytes_ptr = __get_ptr(temp_bytes)
-        cdef uintptr_t out_bytes_ptr = __get_ptr(out_bytes)
+        cdef uintptr_t in_ptr = Array(in_arr).ptr
```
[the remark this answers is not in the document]
thomcom: I'll check that too.

python/kvikio/_lib/pynvcomp.pyx
thomcom: This existing signature matches nvcomp 2.1, I wouldn't want to make them out of sync. However, Nvcomp 2.2 is considerably different.

python/kvikio/_lib/pynvcomp.pyx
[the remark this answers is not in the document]
jakirkham: Probably need this line

https://github.com/rapidsai/kvikio/blob/4d9f905aa3dbdad5f2e8e78f6305ef40d2791ba2/python/kvikio/_lib/libkvikio.pyx#L15

python/kvikio/_lib/pynvcomp.pyx
jakirkham: Gotcha. Perhaps this isn't where things would be coerced to `Array`. Instead maybe that happens in a Python layer then we pass the pointer and size from that layer to here

thomcom:
acceleration_ascending
compression_ascending
acceleration_random
compression_random
buffer_size_3x_and_1 5x

thomcom:
This is done unless there are more changes via review! Thanks!

python/kvikio/__init__.py
madsbk: @thomcom I insist :) 
Let's keep implicit imports to a minimum.

python/kvikio/_lib/pynvcomp.pyx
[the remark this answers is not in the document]
jakirkham: Is that including the `cimport` line above?

python/kvikio/_lib/pynvcomp.pyx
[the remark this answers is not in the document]
jakirkham: Ok

python/kvikio/_lib/pynvcomp.pyx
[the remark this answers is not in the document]
thomcom: Yes

python/kvikio/__init__.py
thomcom: Haha! It's your library!

python/kvikio/_lib/pynvcomp.pyx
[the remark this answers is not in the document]
jakirkham: Interesting we're probably missing something, but let's leave that for future work

jakirkham (approved):
LGTM. Thanks Thomson! 😄

madsbk (changes requested):
Sorry for spamming the same suggestion, but I think some docstrings would be good. I doesn't have to anything lengthy. Just a short description, list of arguments/types, and maybe an url to the nvcomp doc when relevant. 

Otherwise it looks good!

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,238 @@
...
+from enum import Enum
+
+import cupy as cp
+import numpy as np
+
+import kvikio._lib.pynvcomp as _lib
+
+
+def make_ptr_collection(data: collections.abc.Collection) -> list:
+    """ Returns a list of ndarray pointers.
```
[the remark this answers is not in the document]
thomcom: I'm sorry, I forgot! I will add docstrings tonight and tomorrow.

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,238 @@
...
+            self.decompress_temp_size,
+            self.decompress_out_buffer,
+            self.decompress_out_size,
+            self.s.ptr,
+        )
+        return self.decompress_out_buffer.view(self.dtype)
+
+
+class SnappyCompressor:
+    def __init__(self):
```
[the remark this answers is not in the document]
thomcom: Snappy isn't quite ready yet.

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,238 @@
...
+        self.compressor = _lib.LibSnappyCompressor()
+
+    def compress(self, data: cp.ndarray) -> cp.ndarray:
+        # convert data to pointers
```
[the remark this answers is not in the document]
thomcom: "

thomcom:
@madsbk I added the docs that I forgot this morning. Thanks! I also dropped BatchedSnappy which has some issues I won't be working on presently.

madsbk (approved):
Thanks @thomcom, great work!

I took the liberty to rename `nvcomp Cascade Benchmarks.ipynb -> nvcomp_cascade_benchmarks.ipynb`, hope that is okay.

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,224 @@
...
+import cupy as cp
+import numpy as np
+
+import kvikio._lib.pynvcomp as _lib
+
+
+def cp_to_nvcomp_dtype(in_type: cp.dtype) -> Enum:
+    """ Convert np/cp dtypes to nvcomp integral dtypes.
+
+    Returns
```
vyasr, line 15: Missing a parameters section, which would answer my above query.

python/kvikio/_lib/pynvcomp.pxd
```diff
@@ -0,0 +1,210 @@
+# Copyright (c) 2022 Carson Swope
```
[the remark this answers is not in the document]
thomcom: Heya, yeah the bindings started out from Carson, then I modified them and also got a SWIPAT.

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,224 @@
...
+
+from enum import Enum
+
+import cupy as cp
+import numpy as np
+
+import kvikio._lib.pynvcomp as _lib
+
+
+def cp_to_nvcomp_dtype(in_type: cp.dtype) -> Enum:
```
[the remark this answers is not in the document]
thomcom: It was a coercion function until I added mypy notations. Now it is only called by `CascadedCompressor` and the type has already been coerced, so it just converts to the proper NVCOMP_TYPE. I'll drop the coercion.

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,224 @@
...
+        self.dtype = dtype
+        self.config = config
+        self.compressor = _lib._CascadedCompressor(
+            cp_to_nvcomp_dtype(self.dtype).value,
+            config.num_RLEs,
+            config.num_deltas,
+            config.use_bp,
+        )
+        self.decompressor = _lib._CascadedDecompressor()
+        self.s = cp.cuda.Stream()
```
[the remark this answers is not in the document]
thomcom: I think immutable is a good idea here, the fact that they are properties of the class is really just making them available as queryable parameters. They've been set in the underlying C++ class and are no longer accessible by any API, so I keep a record here.

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
thomcom: So, coming back to this I'm reminded that there are many representations of valid cp.dtypes. The coercion is because, for example, `cp.int8` defined during runtime is actually `<class 'numpy.int8'>`, which doesn't seem to be caught by `mypy` in this context. There are many more valid dtype specifications that won't be caught by `mypy` such as passing one of the string variations that should be permitted. I'm going back to the coercion form. `mypy` annotations are just suggestions, after all. :)

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,224 @@
...
+    Cascaded Compressor.
+    """
+
+    def __init__(self, num_RLEs: int = 1, num_deltas: int = 1, use_bp: bool = True):
+        self.num_RLEs = num_RLEs
+        self.num_deltas = num_deltas
+        self.use_bp = use_bp
+
+
+class CascadedCompressor:
```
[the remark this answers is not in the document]
thomcom: Indeed, I'd like to push it till later. :)

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,224 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
```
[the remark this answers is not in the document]
thomcom: I'm just copying the pythonic convention here in `kvikio`.

python/kvikio/nvcomp.py
thomcom: I added one that might satisfy your curiosity.

python/tests/test_nvcomp.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
+
+import pytest
+
+cudf = pytest.importorskip("cudf")
+cupy = pytest.importorskip("cupy")
+np = pytest.importorskip("numpy")
```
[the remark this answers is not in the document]
thomcom: done

python/kvikio/nvcomp.py
```diff
@@ -0,0 +1,224 @@
...
+        cp.dtype("int16"): _lib.pyNvcompType_t.pyNVCOMP_TYPE_SHORT,
+        cp.dtype("uint16"): _lib.pyNvcompType_t.pyNVCOMP_TYPE_USHORT,
+        cp.dtype("int32"): _lib.pyNvcompType_t.pyNVCOMP_TYPE_INT,
+        cp.dtype("uint32"): _lib.pyNvcompType_t.pyNVCOMP_TYPE_UINT,
+        cp.dtype("int64"): _lib.pyNvcompType_t.pyNVCOMP_TYPE_LONGLONG,
+        cp.dtype("uint64"): _lib.pyNvcompType_t.pyNVCOMP_TYPE_ULONGLONG,
+    }[cp_type]
+
+
+class CascadedOptions:
```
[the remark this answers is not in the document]
thomcom: I felt like this configuration made the parameters the least visible to end users, who I presume have no use for them unless they're power users with very specific data distributions that could profit from tweaking these parameters. Having anything in the default constructor signature other than "CascadedOptions" makes it seem like the user of `CascadedCompressor` needs to know about them. That's my theory, anyway, it could be more cognitive load as you suggest.

python/kvikio/_lib/pynvcomp.pxd
```diff
@@ -0,0 +1,210 @@
...
+            void* out_ptr,
+            const size_t out_bytes,
+            cudaStream_t stream) except+
+
+# Low-level LZ4 API
+cdef extern from "nvcomp/lz4.h":
+
+    # _nvcompError_t
+    cdef nvcompError_t _nvcompBatchedLZ4CompressGetTempSize \
+        "nvcompBatchedLZ4CompressGetTempSize" (
```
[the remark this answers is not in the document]
thomcom: Well, there's a C++ class, a cython class, and a python class. The C++ class and the Cython class are identical, but a wrapper of course. The python class implements a slightly different API than the lower level classes.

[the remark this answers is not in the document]
vyasr: Yeah we can worry about this naming later.

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
vyasr: @madsbk @jakirkham what do you think? IMO keyword arguments are much more Pythonic than config objects, which I don't expect users to be familiar with. I think users are accustomed to ignoring keyword arguments that they don't understand or need to know about.

vyasr:
@thomcom my last request here would be to move `pynvcomp.pxd` to something like `libnvcomp.pxd`, then cimport it explicitly in `pynvcomp.pyx`. In Cython `pxd` and `pyx` files with the same name are interpreted in a very specific way, and in this case that relationship isn't accurate. In particular, there's no pure Cython in pynvcomp.pyx that you are trying to make `cimport`able in other Cython files, you're exclusively creating objects that are visible in pure Python. `pynvcomp.pxd` is not defining an interface for `pynvcomp.pyx`, it's importing the "symbols" from the C++ library. As such, it's misleading to have the names aligned.

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
madsbk: I agree, keyword arguments with useful default values are more Pythonic. 
I think a `CascadedOptions` would make more sense if there are multiple sets of default options the user should choose between.

thomcom:
Hey @vyasr, your comment is interesting to me. My approach with .pxd and .pyx naming has always been that the .pxd file defines the interface to the existing C/C++ code, and the pyx implements the python bindings for that interface. In this case, the pure cython in `pynvcomp.pyx` is only the typecastings that convert between python objects and c pointers. It's not intended for cimport into other cython files, only to expose the C++ interface to python. Would you put the C++ interface and the python visible bindings into the `.pyx` file and not have a pxd file at all? I see that in `_lib` here there's a `kvikio_cxx_api.pxd` and a `libkvikio.pyx`, how about I rename into that pattern?

vyasr:
You might be interested in looking at [the first section on this page](https://cython.readthedocs.io/en/latest/src/tutorial/pxd_files.html). Specifically, it's valuable to distinguish between use-cases 1 and 3. Use-case 1 is really what the pxd file in this PR is doing (exposing C++ code). In RAPIDS we rarely write pure Cython, it's almost always just a wrapper around C++, so we don't run into use-case 3 as much, but there are some good examples in cudf like [column.pxd](https://github.com/rapidsai/cudf/blob/branch-22.06/python/cudf/cudf/_lib/column.pxd) and [copying.pxd](https://github.com/rapidsai/cudf/blob/branch-22.06/python/cudf/cudf/_lib/copying.pxd). 

Regarding the naming, I suggested the `libFoo.pxd` for C++ bindings and `pyFoo.pyx` or just `Foo.pyx` for the Cython since I think that's the common convention that I have observed across most other Cython packages that I've worked with. I'm fine with your proposal too since it maintains internal consistency in kvikio. Feel free to check with other devs too in case they express different preferences from mine.

madsbk:
> I see that in `_lib` here there's a `kvikio_cxx_api.pxd` and a `libkvikio.pyx`, how about I rename into that pattern?

I think this is a good idea. Then, if we want to change to another naming scheme like @vyasr suggest, we can change the name of both libraries at the same time.

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
thomcom: Ok, I changed it to named defaults.

thomcom:
All requests have been answered.

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
vyasr: Note for future work: Depending on whether users might actually need to access these properties, I would recommend either renaming them with underscores (e.g. `self._decompressor`) or turning them into properties instead of raw attributes.

vyasr (approved):
@thomcom I've unresolved a pair of conversations to track potential future work, but otherwise looks good to go to me now!

madsbk (approved):

madsbk:
Thanks @thomcom, it is merged.

PR #30, opened 2022-03-17
Rename macros to avoid name collisions

madsbk:


madsbk:
@gpucibot rerun tests

madsbk:
rerun tests

PR #27, opened 2022-03-17
Build improvements

jakirkham:
Still WIP. Includes a couple changes to the recipes and CI scripts.

PR #32, opened 2022-03-17
Replace setuptools with scikit-build in the Python build

vyasr:
This PR rewrites the Python build system to use scikit-build, enabling easier interfacing with CMake and therefore integration with C++ builds. This change should help facilitate more complicated Python/C++ build interactions like what we might need for #24.

### DIFF
--- a/conda/recipes/kvikio/meta.yaml
+++ b/conda/recipes/kvikio/meta.yaml
@@ -28,8 +28,8 @@ requirements:
     - setuptools
     - cython >=0.29,<0.30
     - cudatoolkit {{ cuda_version }}.*
-    - cmake>=3.20.1
-    - scikit-build>=0.13.1
+    - cmake >=3.20.1
+    - scikit-build >=0.13.1
     - libkvikio {{ version }}
   run:
     - cupy
--- a/conda/recipes/kvikio/meta.yaml
+++ b/conda/recipes/kvikio/meta.yaml
@@ -23,12 +23,14 @@ build:
     - CUDAHOSTCXX
 
 requirements:
+  build:
+    - cmake >=3.20.1
+    - ninja
   host:
     - python
     - setuptools
     - cython >=0.29,<0.30
     - cudatoolkit {{ cuda_version }}.*
-    - cmake >=3.20.1
     - scikit-build >=0.13.1
     - libkvikio {{ version }}
   run:
--- a/conda/recipes/kvikio/meta.yaml
+++ b/conda/recipes/kvikio/meta.yaml
@@ -28,6 +28,8 @@ requirements:
     - setuptools
     - cython >=0.29,<0.30
     - cudatoolkit {{ cuda_version }}.*
+    - cmake>=3.20.1
+    - scikit-build>=0.13.1
     # - libkvikio {{ version }}
   run:
     - cupy
--- a/python/cmake/kvikio_python_helpers.cmake
+++ b/python/cmake/kvikio_python_helpers.cmake
@@ -41,7 +41,7 @@ function(add_cython_modules cython_modules)
         PROPERTIES PREFIX ""
                    CXX_STANDARD 17
     )
-    target_link_libraries(${cython_module} kvikio::kvikio)
+    target_link_libraries(${cython_module} kvikio)
 
     # Compute the install directory relative to the source and rely on installs being relative to
     # the CMAKE_PREFIX_PATH for e.g. editable installs.
--- a/python/cmake/kvikio_python_helpers.cmake
+++ b/python/cmake/kvikio_python_helpers.cmake
@@ -36,7 +36,11 @@ function(add_cython_modules cython_modules)
     python_extension_module(${cython_module})
 
     # To avoid libraries being prefixed with "lib".
-    set_target_properties(${cython_module} PROPERTIES PREFIX "")
+    set_target_properties(
+        ${cython_module}
+        PROPERTIES PREFIX ""
+                   CXX_STANDARD 17
+    )
     target_link_libraries(${cython_module} kvikio::kvikio)
 
     # Compute the install directory relative to the source and rely on installs being relative to
--- a/python/cmake/kvikio_python_helpers.cmake
+++ b/python/cmake/kvikio_python_helpers.cmake
@@ -41,7 +41,10 @@ function(add_cython_modules cython_modules)
         PROPERTIES PREFIX ""
                    CXX_STANDARD 17
     )
+    # Link to the C++ library.
     target_link_libraries(${cython_module} kvikio)
+    # Treat warnings as errors when compiling.
+	target_compile_options(${cython_module} PRIVATE -Werror)
 
     # Compute the install directory relative to the source and rely on installs being relative to
     # the CMAKE_PREFIX_PATH for e.g. editable installs.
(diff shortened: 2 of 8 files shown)


jakirkham:
Thanks Vyas! 😄 This is great to see.

FWIW there were some smaller build issues that I was cleaning up in PR ( https://github.com/rapidsai/kvikio/pull/27 ).

vyasr:
Cool thanks for the heads up. Those look orthogonal so hopefully the changes won't end up conflicting at all.

vyasr:
@jakirkham any idea what might be off with the build here? I see scikit-build in the environment on [line 798 of this build](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/kvikio/job/prb/job/kvikio-cpu-cuda-build/CUDA=11.5/62/console), but then it's not actually finding the module when it runs the mambabuild (error on line 1130). Does that use a different environment?

jakirkham:
Agreed. Was more curious if they would be needed to solve some of the build errors that you were seeing

vyasr:
So far nope, just oversights on my part. But let's see what happens when CI gets around to actually running tests on the build...

quasiben:
WOOT!

vyasr:
I need some guidance on how we want to be handling the `CUFILE_HOME` and `CUDA_HOME` environment variables in the CMake builds, but aside from that this seems to be working just fine!

jakirkham:
We have [`FindcuFile.cmake`]( https://github.com/rapidsai/kvikio/blob/branch-22.04/cpp/cmake/Modules/FindcuFile.cmake ), which should handle CMake detection. AIUI CMake has builtin CUDA Toolkit detection. So perhaps this is already largely taken care of for us?

vyasr:
Ah yes OK. So I probably want to load `FindcuFile.cmake` in the python CMakeLists.txt, then we should be set.

conda/recipes/kvikio/meta.yaml
```diff
@@ -28,6 +28,8 @@ requirements:
     - setuptools
     - cython >=0.29,<0.30
     - cudatoolkit {{ cuda_version }}.*
+    - cmake>=3.20.1
+    - scikit-build>=0.13.1
```
jakirkham, line 32: Minor suggestion. Would space out the version constraints. This is the norm in recipes

```suggestion
    - cmake >=3.20.1
    - scikit-build >=0.13.1
```

python/setup.py
```diff
@@ -168,12 +25,16 @@ def finalize_options(self):
...
-    install_requires=install_requires,
+    package_data={
+        # Note: A dict comprehension with an explicit copy is necessary (rather
+        # than something simpler like a dict.fromkeys) because otherwise every
+        # package will refer to the same list and skbuild modifies it in place.
+        key: ["*.pyi", "*.pxd"]
+        for key in find_packages(include=["kvikio._lib"])
+    },
+    cmdclass=versioneer.get_cmdclass(),
+    install_requires=["Cython>=0.29,<0.30"],
```
jakirkham, line 38: Is this just how `scikit-build` handles build dependencies?

Also is this still needed given it is in `pyproject.toml` above?

python/cmake/kvikio_python_helpers.cmake
```diff
@@ -0,0 +1,56 @@
...
+  foreach(cython_module ${cython_modules})
+    add_cython_target(${cython_module} CXX PY3)
+    add_library(${cython_module} MODULE ${cython_module})
+    python_extension_module(${cython_module})
+
+    # To avoid libraries being prefixed with "lib".
+    set_target_properties(
+        ${cython_module}
+        PROPERTIES PREFIX ""
+                   CXX_STANDARD 17
```
jakirkham, line 42: Do we need C++17 here?

jakirkham (commented):
Thanks Vyas! 😄

Had a couple questions below 🙂

python/setup.py
vyasr: What do you mean by "this"?

No we don't need this. I just copied it in without thinking. Cython is only a build-time requirement, not an install-time requirement, so it should only have been in pyproject.toml to begin with.

python/cmake/kvikio_python_helpers.cmake
vyasr: My assumption was yes because we used to specify it in the [`extra_compile_args`](https://github.com/rapidsai/kvikio/pull/32/files#diff-eb8b42d9346d0a5d371facf21a8bfa2d16fb49e213ae7c21f03863accebe0fcfL93). I assumed that it would be necessary because we're compiling against headers that require C++17 (and therefore may use C++17 features).

jakirkham:
rerun tests

(one of [the CI jobs had a download error]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/kvikio/job/prb/job/kvikio-cpu-python-build/PYTHON=3.9/52/console ))

python/setup.py
jakirkham: Sorry by this just meant using `install_requires` to handle build dependencies. Didn't know if `scikit-build` was reinterpreting `install_requires` in some way. In any event sound like no and it has since been dropped.

python/pyproject.toml
```diff
@@ -7,6 +7,9 @@ requires = [
     "wheel",
     "setuptools",
     "Cython>=0.29,<0.30",
+    "scikit-build>=0.13.1",
+    "cmake>=3.20.1",
+    "ninja",
```
jakirkham, line 12: Should we add `ninja` to the Conda recipe `requirements/build` as well?

conda/recipes/kvikio/meta.yaml
```diff
@@ -28,6 +28,8 @@ requirements:
     - setuptools
     - cython >=0.29,<0.30
     - cudatoolkit {{ cuda_version }}.*
+    - cmake >=3.20.1
```
jakirkham, line 31: Should `cmake` be in `build`? Typically that is where build tools (like `cmake`) are placed. Though please let me know if there is a reason scikit-build prefers `cmake` in `host`

jakirkham (commented):
Thanks Vyas! 😄

Generally looks good. Spotted a couple things I'd overlooked before. Have commented on them, but think that should be it.

conda/recipes/kvikio/meta.yaml
vyasr: Hmm that is a good question. I think `build` makes the most sense, but I put it in `host` because that's where all the build tools were previously placed (`setuptools`, `cython`). Should I move all of them over?

vyasr: Looking at the current list, I think _technically_ we would want
```
  build:
    - setuptools
    - cython >=0.29,<0.30
    - cmake >=3.20.1
    - scikit-build >=0.13.1
    - ninja
  host:
    - cudatoolkit {{ cuda_version }}.*
    - python
    - libkvikio {{ version }}
```
But in practice we create separate packages for every platform so everything can safely go in `build`. Does that sound right to you? Also is `python` necessary at all? I don't see `Python.h` anywhere, so I assume we don't actually need to link to the Python libraries (except as handled by Cython), but I guess no harm in including it.

jakirkham: Only command line tools (like `cmake` and `ninja`) would move. Things that are linked against or imported (like `python`, `cython`, and `setuptools`) would stay in `host`.

jakirkham (approved):

PR #34, opened 2022-03-18
Auto detect running inside docker

### LINKED ISSUE #33: Unable to create a working build inside a container
I am running inside a Docker container, specifically one created using [rapids-compose](https://github.com/trxcllnt/rapids-compose). I am able to successfully build both the Python and C++ kvikio libraries, but both of them fail at runtime with a cuFile internal error. For example, using the [exact C++ building instructions here](https://github.com/rapidsai/kvikio#c-build-from-source) appears to succeed, but then running the `basic_io` executable gives 
```
(rapids) rapids@compose:~/kvikio/cpp/build$ ./examples/basic_io 
KvikIO config: 
  Compatibility mode: disabled
DriverProperties: 
  Version: 0.0
  Allow compatibility mode: true
  Pool mode - enabled: false, threshold: 4 kb
  Max pinned memory: 33554432 kb
terminate called after throwing an instance of 'kvikio::CUfileException'
  what():  cuFile error at: /home/vyasr/local/rapids/kvikio/cpp/examples/../include/kvikio/file_handle.hpp:117: internal error
Aborted (core dumped)
```
I get the same `cuFile... internal error` if I try to run `pytest` on anything in the `python/tests` directory.

madsbk:
Automatically enable KvikIO compatibility mode when `/run/udev` isn't readable. This typically happens when running inside a docker image not launched with `--volume /run/udev:/run/udev:ro`.

Closes  #33

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -66,9 +66,6 @@ conda list --show-channel-urls
 
 cd "$WORKSPACE/python"
 
-# For now, we have to run KvikIO in compatibility mode
-export KVIKIO_COMPAT_MODE=true
-
 gpuci_logger "Build kvikio from source"
 python -m pip install .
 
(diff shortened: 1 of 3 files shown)


quasiben:
@vyasr can you confirm this fixes things when you run in docker ?

vyasr:
Yup, with this PR I am able to generate builds that work correctly!

PR #43, opened 2022-03-22
Fallback to Compat mode when `cufile.h` isn't found

madsbk:
Use `__has_include(<cufile.h>)` to fallback to compat mode.

The motivation here is to make KvikIO work in all circumstances so that libraries doesn't have to implement there own fallback solutions.


madsbk:
rerun tests

PR #53, opened 2022-04-04
Add new files to update version script

charlesbluca:
Adds `python/CMakeLists.txt` and `scripts/format-all.py` to the version updating script, so that the RAPIDS versions in these files are automatically bumped around release time.

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -33,7 +33,7 @@ function sed_runner() {
 sed_runner 's/'"VERSION ${CURRENT_SHORT_TAG}.*"'/'"VERSION ${NEXT_FULL_TAG}"'/g' cpp/CMakeLists.txt
 
 # python update
-sed_runner 's/kvikio_version '${CURRENT_SHORT_TAG}'.*/kvikio_version '${NEXT_FULL_TAG}'\)/g' python/CMakeLists.txt
+sed_runner 's/set(kvikio_version.*)/set(kvikio_version '${NEXT_FULL_TAG}')/g' python/CMakeLists.txt
 
 # script update
 sed_runner 's/version=.*/version="'${NEXT_SHORT_TAG}'"\):/g' scripts/format-all.py
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -30,13 +30,13 @@ function sed_runner() {
 }
 
 # cpp update
-sed_runner 's/'"VERSION ${CURRENT_SHORT_TAG}.*"'/'"VERSION ${NEXT_FULL_TAG}"'/g' cpp/CMakeLists.txt
+sed_runner "/project(/,/)/s/VERSION.*/VERSION ${NEXT_FULL_TAG}/" cpp/CMakeLists.txt
 
 # python update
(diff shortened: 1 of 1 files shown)


madsbk (approved):
LGTM, thanks @charlesbluca

ci/release/update-version.sh
```diff
@@ -32,6 +32,12 @@ function sed_runner() {
 # cpp update
 sed_runner 's/'"VERSION ${CURRENT_SHORT_TAG}.*"'/'"VERSION ${NEXT_FULL_TAG}"'/g' cpp/CMakeLists.txt
 
+# python update
+sed_runner 's/kvikio_version '${CURRENT_SHORT_TAG}'.*/kvikio_version '${NEXT_FULL_TAG}'\)/g' python/CMakeLists.txt
```
ajschmidt8, line 36:
```suggestion
sed_runner 's/set(kvikio_version.*)/set(kvikio_version '${NEXT_FULL_TAG}')/g' python/CMakeLists.txt
```

I think it's best to create these expressions so that the search term in the `sed` expression doesn't include the current version. This can cause issues if the lines that we're trying to change get out-of-sync with the current versions (which unfortunately does happen from time to time).

Btw, you can test this file locally with `bash ci/release/update-version.sh 22.06.00`. When I ran it with the original changes here, it didn't update the `python/CMakeLists.txt` file correctly. After making the suggested changes here, it does seem to be working.

ajschmidt8 (changes requested):

ci/release/update-version.sh
charlesbluca: That makes sense to me 🙂 out of interest, is there any way we can do the same for the expression updating our cpp/CMakeLists.txt?

https://github.com/rapidsai/kvikio/blob/c95d34c8fbae3227f7a99e3afb10c9b0b7b67539/ci/release/update-version.sh#L32-L33

Seems like it might be a little more complicated since we have a few `VERSION`s throughout the file

ajschmidt8: @charlesbluca, yes. I tested the line below which seemed to work.


```sh
sed_runner "/project(/,/)/s/VERSION.*/VERSION ${NEXT_FULL_TAG}/" cpp/CMakeLists.txt
```

The `,` separated values `/project(/` and `/)/` are a [sed address range](https://www.gnu.org/software/sed/manual/html_node/Range-Addresses.html) which basically says to only perform the following `s///` operation on the lines between `project(` and the first `)` that comes after it.

ajschmidt8: You can test these locally by alternating between the values below to make sure they all change correctly:

```sh
bash ci/release/update-version.sh 22.04.00
bash ci/release/update-version.sh 22.06.00
```

charlesbluca: Thanks for the help here! I'll go ahead and commit these changes

ajschmidt8 (approved):
:fire: :fire: :fire:

charlesbluca:
@gpucibot merge

jakirkham:
Thanks all! 😄

PR #54, opened 2022-04-05
compat_mode per `FileHandle`

madsbk:
KvikIO's compatibility mode is now per `FileHandle` thus changing compatibility mode, effects all new `FileHandles` that doesn't sets the `compat_mode` argument explicitly but it never effect existing `FileHandles`.

### DIFF
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -130,7 +130,7 @@ class FileHandle {
   int _fd_direct_on{-1};
   int _fd_direct_off{-1};
   bool _initialized{false};
-  bool _combat_mode{false};
+  bool _compat_mode{false};
   mutable std::size_t _nbytes{0};  // The size of the underlying file, zero means unknown.
   CUfileHandle_t _handle{};
 
@@ -161,9 +161,9 @@ class FileHandle {
     : _fd_direct_on{open_fd(file_path, flags, true, mode)},
       _fd_direct_off{open_fd(file_path, flags, false, mode)},
       _initialized{true},
-      _combat_mode{compat_mode}
+      _compat_mode{compat_mode}
   {
-    if (_combat_mode) { return; }
+    if (_compat_mode) { return; }
 #ifdef KVIKIO_CUFILE_EXIST
     CUfileDescr_t desc{};  // It is important to set to zero!
     desc.type = CU_FILE_HANDLE_TYPE_OPAQUE_FD;
@@ -182,7 +182,7 @@ class FileHandle {
     : _fd_direct_on{std::exchange(o._fd_direct_on, -1)},
       _fd_direct_off{std::exchange(o._fd_direct_off, -1)},
       _initialized{std::exchange(o._initialized, false)},
-      _combat_mode{std::exchange(o._combat_mode, false)},
+      _compat_mode{std::exchange(o._compat_mode, false)},
       _nbytes{std::exchange(o._nbytes, 0)},
       _handle{std::exchange(o._handle, CUfileHandle_t{})}
   {
@@ -192,7 +192,7 @@ class FileHandle {
     _fd_direct_on  = std::exchange(o._fd_direct_on, -1);
     _fd_direct_off = std::exchange(o._fd_direct_off, -1);
     _initialized   = std::exchange(o._initialized, false);
-    _combat_mode   = std::exchange(o._combat_mode, false);
+    _compat_mode   = std::exchange(o._compat_mode, false);
     _nbytes        = std::exchange(o._nbytes, 0);
     _handle        = std::exchange(o._handle, CUfileHandle_t{});
     return *this;
@@ -208,7 +208,7 @@ class FileHandle {
   {
     if (closed()) { return; }
 #ifdef KVIKIO_CUFILE_EXIST
-    if (!_combat_mode) { cuFileAPI::instance()->HandleDeregister(_handle); }
+    if (!_compat_mode) { cuFileAPI::instance()->HandleDeregister(_handle); }
 #endif
     ::close(_fd_direct_on);
     ::close(_fd_direct_off);
@@ -272,7 +272,7 @@ class FileHandle {
                    std::size_t file_offset,
                    std::size_t devPtr_offset)
   {
-    if (_combat_mode) {
+    if (_compat_mode) {
       return posix_read(_fd_direct_off, devPtr_base, size, file_offset, devPtr_offset);
     }
 #ifdef KVIKIO_CUFILE_EXIST
@@ -322,7 +322,7 @@ class FileHandle {
   {
     _nbytes = 0;  // Invalidate the computed file size
 
-    if (_combat_mode) {
+    if (_compat_mode) {
       return posix_write(_fd_direct_off, devPtr_base, size, file_offset, devPtr_offset);
     }
 #ifdef KVIKIO_CUFILE_EXIST
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -240,6 +240,7 @@ class FileHandle {
    */
   [[nodiscard]] inline std::size_t nbytes() const
   {
+    if (closed()) { return 0; }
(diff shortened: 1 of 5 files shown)


cpp/include/kvikio/file_handle.hpp
```diff
@@ -62,25 +63,43 @@ inline int open_fd_parse_flags(const std::string& flags)
...
  * @param flags Open flags given as a string
+ * @param o_direct Append O_DIRECT to `flags`
  * @param mode Access modes
  * @return File descriptor
  */
-inline int open_fd(const std::string& file_path, const std::string& flags, mode_t mode)
+inline int open_fd(const std::string& file_path,
+                   const std::string& flags,
+                   bool o_direct,
+                   mode_t mode)
```
jakirkham, line 81: Is this a public API or intended for internal usage? Is this a breaking API change? Should we be adding this argument in the middle of existing ones? Should it have a default?

cpp/include/kvikio/file_handle.hpp
```diff
@@ -107,120 +126,121 @@ inline int open_fd(const std::string& file_path, const std::string& flags, mode_
...
+   * @param mode Access modes (see `open(2)`).
+   * @param compat_mode Enable KvikIO's compatibility mode for this file.
    */
-  FileHandle(int fd, bool steal_fd = false) : _fd{fd}, _own_fd{steal_fd}, _closed{false}
+  FileHandle(const std::string& file_path,
+             const std::string& flags = "r",
+             mode_t mode              = m644,
+             bool compat_mode         = defaults::compat_mode())
+    : _fd_direct_on{open_fd(file_path, flags, true, mode)},
+      _fd_direct_off{open_fd(file_path, flags, false, mode)},
```
jakirkham, line 162: Why are we opening with both `O_DIRECT` and without?

cpp/include/kvikio/posix_io.hpp
```diff
@@ -208,6 +208,9 @@ inline std::size_t posix_io(int fd,
 /**
  * @brief Read main memory from disk using POSIX
  *
+ * If `size` or `file_offset` isn't aligned with `page_size` then
+ * `fd` cannot have been opened with the `O_DIRECT` flag.
```
jakirkham, line 212: Does this mean we should be doing some kind of error handling when opening with `O_DIRECT`?

cpp/include/kvikio/file_handle.hpp
```diff
@@ -107,120 +126,121 @@ inline int open_fd(const std::string& file_path, const std::string& flags, mode_
...
   }
 
   /**
-   * @brief Get the file descripter of the open file
+   * @brief Get the file descripter of one of the open files
+   *
    * @return File descripter
    */
-  [[nodiscard]] int fd() const noexcept { return _fd; }
+  [[nodiscard]] int fd() const noexcept { return _fd_direct_off; }
```
jakirkham, line 225: Why use the one with `O_DIRECT` off?

cpp/include/kvikio/file_handle.hpp
```diff
@@ -107,120 +126,121 @@ inline int open_fd(const std::string& file_path, const std::string& flags, mode_
...
    * @brief Get the file size
    *
+   * The value are cached.
+   *
    * @return The number of bytes
    */
   [[nodiscard]] inline std::size_t nbytes() const
   {
-    if (_nbytes == 0 && _fd > 0) { _nbytes = get_file_size(_fd); }
+    if (_nbytes == 0) { _nbytes = get_file_size(_fd_direct_on); }
```
jakirkham, line 243: Why is the file descriptor check dropped?

jakirkham (commented):
Thanks Mads! 😄

Had a few questions below

cpp/include/kvikio/file_handle.hpp
madsbk: No, this is private. In a followup PR I will begin using `namespace kvikio::detail` for internal functions like this one.

cpp/include/kvikio/file_handle.hpp
madsbk: `O_DIRECT` requires page-aligned IO, which we cannot guaranty. cuFIle handles this automatically by opening the file without `O_DIRECT` as well. However, `posix_io` does not support `O_DIRECT`.

cpp/include/kvikio/posix_io.hpp
madsbk: I don't think so, `posix_io` is internal so will move in to `namespace kvikio::detail` as well

cpp/include/kvikio/file_handle.hpp
madsbk: This is kind of an arbitrary decisions, my thinking is that `O_DIRECT`-off is the must compatible file descriptor.

cpp/include/kvikio/file_handle.hpp
madsbk: Fixed

madsbk:
@jakirkham thanks for the review

cpp/include/kvikio/file_handle.hpp
jakirkham: Right, but is there a need to open with both modes? Could we for example try `O_DIRECT` and fallback to not using it (if `O_DIRECT` won't work)?

cpp/include/kvikio/posix_io.hpp
jakirkham: Sorry I mean when `FileHandle` is created, is there a risk an exception is thrown because `O_DIRECT` doesn't work? If so, should we be handling this somehow there?

cpp/include/kvikio/file_handle.hpp
jakirkham: Should we document this then?

jakirkham:
Thanks Mads! Replied above

cpp/include/kvikio/file_handle.hpp
madsbk: Good point. I did this because the posix backend will need both with and without `O_DIRECT` in the future for better performance. Now, we could open the extra fd lazily but that will require that the file path is still valid (not moved or removed) at a later time. 

I think we should stick to opening two file descriptor for now, then later we can look at a more lazy approach.

cpp/include/kvikio/posix_io.hpp
madsbk: > Sorry I mean when `FileHandle` is created, is there a risk an exception is thrown because `O_DIRECT` doesn't work? If so, should we be handling this somehow there?

No this cannot happen. My plan is that `posix_io` will be given two file descriptors, one with and one with the `O_DIRECT` flag. It will then use the `O_DIRECT` file for aligned IO only.

cpp/include/kvikio/file_handle.hpp
madsbk: I like to be a bit vague here as we might change this in the future. 
Change to:
```c++
  /**
   * @brief Get one of the file descriptors
   *
   * Notice, `FileHandle` maintains two file descriptors - one opened with the
   * `O_DIRECT` flag and one without. This function returns one of them but
   * it is unspecified with one.
   *
   * @return File descripter
   */
  [[nodiscard]] int fd() const noexcept { return _fd_direct_off; }

  /**
   * @brief Get the flags of one of the file descriptors (see open(2))
   *
   * Notice, `FileHandle` maintains two file descriptors - one opened with the
   * `O_DIRECT` flag and one without. This function returns the flags of one of
   * them but it is unspecified with one.
   *
   * @return File descripter
   */
  [[nodiscard]] int fd_open_flags() const { return open_flags(_fd_direct_on); }
```

cpp/include/kvikio/file_handle.hpp
madsbk: Actually, changed it so we only opens one fd when in regular mode since cuFile requires `O_DIRECT` always.

madsbk:
@jakirkham thanks for the review, think I addressed all of it.

jakirkham (approved):
Looks good. Thanks for the updates Mads! 😄

jakirkham:
@gpucibot merge

madsbk:
Thanks @jakirkham !

PR #58, opened 2022-04-07
Embed Cython docstrings in generated files.

vyasr:
This PR embeds signatures in Cython functions and classes so that they'll be accessible for normal Python help. While docstrings are currently present as expected, we still have to provide Cython with the appropriate directives to tell it that we want signatures.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -16,6 +16,10 @@ cmake_minimum_required(VERSION 3.20.1 FATAL_ERROR)
 
 set(kvikio_version 22.06.00)
 
+set(CYTHON_FLAGS
+    "--directive binding=True,embedsignature=True,always_allow_keywords=True"
+    CACHE STRING "The directives for Cython compilation.")
+
 project(
   kvikio-python
   VERSION ${kvikio_version}


madsbk (approved):
LGTM, thanks @vyasr

python/CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ cmake_minimum_required(VERSION 3.20.1 FATAL_ERROR)
 
 set(kvikio_version 22.06.00)
 
+set(CYTHON_FLAGS
+    "--directive binding=True,embedsignature=True,always_allow_keywords=True"
```
jakirkham, line 20: Are there other things we should handle this way like opting into Python 3 syntax, using C++, etc.?

vyasr: - Using C++ is generally handled by Cython using file extensions, but with scikit-build CMake will take care of that so that isn't necessary.
- We could opt into various optimizations like turning of bounds checking and wraparounds, but since we don't actually write much Cython code and it's just there for bindings I don't think it's necessary. We can always revisit if we ever start incorporating actual Cython code into performance-critical sections.
- Opting into Python 3 syntax is definitely a possibility. I'd be open to specifying a language level here. Note that whenever we upgrade to Cython 3.0 (I think the main obstacle right now is getting https://github.com/conda-forge/cython-feedstock/pull/75 through) the language level will automatically be set to 3, but no harm in doing it before. I don't know how big of a difference it makes given our limited Cython (Python files default to the language level of the interpreter, so those will already be 3), so let me know what you'd like me to do here.
- We may want to add some CMake options to enable Cython profiling that would modify these flags, but that's probably something that can wait until someone has a use case?

madsbk:
@gpucibot merge

python/CMakeLists.txt
jakirkham: Coming back to this comment, wondering if we should handle this as part of other needed updates to the build tooling here ( https://github.com/rapidsai/kvikio/issues/186 )

vyasr: - I'd still want to see the benefits of turning off bounds checking etc. Also in general IMO we'd be better off using Cython's decorator syntax to do this on a per-function basis rather than globally where it is more likely to unintentionally lead to bugs.
- Same deal for the profiling flags, not much reason to add them unless someone is actually profiling Cython IMO.
- We recently tried opting into `language_level=3` and found some issues, which I suspect may be bugs in the Cython<3 implementation of it but haven't had time to confirm. We'll probably need to revisit, but at this point I would expect that transition to happen as part of a broader RAPIDS switch to using Cython 3.

PR #59, opened 2022-04-20
Keep the rapids-cmake version insync with calver

robertmaynard:
Update to rapids-cmake 22.06 and also update the release scripts so we don't have a calver dsync going forward.


madsbk (approved):
LGTM, thanks @robertmaynard!

madsbk:
@gpucibot merge

PR #68, opened 2022-05-05
Use CMake provided targets for the cuda driver and dl libraries

robertmaynard:


madsbk (approved):
LGTM, thanks @robertmaynard

madsbk:
@gpucibot merge

PR #69, opened 2022-05-05
Mark detail functions as inline

vyasr:
These function needs to be inline so that the header can be safely included in multiple files in the same TU without ODR violations.


madsbk (approved):
Thanks @vyasr for the quick fix!
I have added some more missing inlines :)

hyperbolic2346 (approved):

vyasr:
> Thanks @vyasr for the quick fix! I have added some more missing inlines :)

Thanks, I only caught the ones that are files that cudf is actually including right now!

vyasr:
@gpucibot merge

PR #70, opened 2022-05-06
dlopen `libcuda.so`

madsbk:
Like we do with cuFile, dlopen the CUDA driver library

@shwina

### DIFF
--- a/cpp/include/kvikio/shim/cuda.hpp
+++ b/cpp/include/kvikio/shim/cuda.hpp
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cuda.h>
+
+#include <kvikio/shim/utils.hpp>
+
+namespace kvikio {
+
+/**
+ * @brief Shim layer of the cuda C-API
+ *
+ * This is a singleton class that use `dlopen` on construction to load the C-API of cuda.
+ *
+ * For example, `cudaAPI::instance()->MemHostAlloc()` corresponds to calling `cuMemHostAlloc()`
+ */
+class cudaAPI {
+ public:
+  decltype(cuInit)* Init{nullptr};
+  decltype(cuMemHostAlloc)* MemHostAlloc{nullptr};
+  decltype(cuMemFreeHost)* MemFreeHost{nullptr};
+  decltype(cuMemcpyHtoD)* MemcpyHtoD{nullptr};
+  decltype(cuMemcpyDtoH)* MemcpyDtoH{nullptr};
+  decltype(cuPointerGetAttribute)* PointerGetAttribute{nullptr};
+  decltype(cuCtxPushCurrent)* CtxPushCurrent{nullptr};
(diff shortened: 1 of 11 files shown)


cpp/include/kvikio/shim/cuda.hpp
```diff
@@ -0,0 +1,71 @@
...
+    get_symbol(MemcpyDtoH, lib, KVIKIO_STRINGIFY(cuMemcpyDtoH));
+    get_symbol(PointerGetAttribute, lib, KVIKIO_STRINGIFY(cuPointerGetAttribute));
+    get_symbol(CtxPushCurrent, lib, KVIKIO_STRINGIFY(cuCtxPushCurrent));
+    get_symbol(CtxPopCurrent, lib, KVIKIO_STRINGIFY(cuCtxPopCurrent));
+    get_symbol(MemGetAddressRange, lib, KVIKIO_STRINGIFY(cuMemGetAddressRange));
+    get_symbol(GetErrorName, lib, KVIKIO_STRINGIFY(cuGetErrorName));
+    get_symbol(GetErrorString, lib, KVIKIO_STRINGIFY(cuGetErrorString));
+  }
+
+  static cudaAPI* instance()
```
robertmaynard, line 64: I wonder if we should return a reference instead of a pointer. The construction of the `cudaAPI` can't fail so we know it won't be a nullptr. In addition it provides better intent to the caller that they shouldn't delete the pointer returned.

robertmaynard (changes requested):
Overall looks great!

We also need to drop the following line from the CMakeLists.txt since we don't need to link to `libcuda.so` anymore.

```cmake

target_link_libraries(kvikio INTERFACE CUDA::cuda_driver)
```

madsbk:
> We also need to drop the following line from the CMakeLists.txt since we don't need to link to `libcuda.so` anymore.

@robertmaynard since we still need `cuda.h`, do you know if we can do something similar to:
```cmake
target_link_libraries(kvikio INTERFACE cufile::cuFile_interface)
```
Or do we just assume that `cuda.h` in the include path?

robertmaynard:
> @robertmaynard since we still need `cuda.h`, do you know if we can do something similar to:
> 
> ```cmake
> target_link_libraries(kvikio INTERFACE cufile::cuFile_interface)
> ```

Yes, if you link to the `CUDA::toolkit` target you will get the required include directories with no actual linking to a cuda library

madsbk:
Thanks for the review @robertmaynard, we now return references and linking like:
```cmake
target_link_libraries(kvikio INTERFACE CUDA::toolkit)
```

robertmaynard (approved):

jakirkham:
Should we add some kind of test to ensure we are not linking to `libcuda.so`?

Is there anything we need to do on the Python side? Or for nvCOMP bindings?

madsbk:
> Should we add some kind of test to ensure we are not linking to `libcuda.so`?

I checked this previously https://github.com/rapidsai/kvikio/pull/70/commits/f8372148f080fe9f9d1b2f852c96fc80f4b9d80c but now that we don't link to `libcuda.so`, we get a compile error if using the CUDA API directly:

> Is there anything we need to do on the Python side? Or for nvCOMP bindings?

I don't think so, nvCOMP is using the Runtime CUDA API.

robertmaynard:
Why did we drop the following check, as it also confirmed that we didn't bring in a hard-dependency transitively?

```
# Check that `libcuda.so` is NOT being linked
LDD_BASIC_IO=`ldd ${WORKSPACE}/libkvikio-debug-build/examples/basic_io`
if [[ "$LDD_BASIC_IO" == *"libcuda.so"* ]]; then
  echo "[ERROR] examples/basic_io shouln't link to libcuda.so: ${LDD_BASIC_IO}"
  return 1
fi
```

madsbk:
> Why did we drop the following check, as it also confirmed that we didn't bring in a hard-dependency transitively?

Good point, adding the check again :)

jakirkham:
@robertmaynard what do you think about the changes above? 🙂

robertmaynard (approved):

robertmaynard:
> @robertmaynard what do you think about the changes above? slightly_smiling_face

Approved

jakirkham (approved):

jakirkham:
@gpucibot merge

madsbk:
Thanks @robertmaynard and @jakirkham !

PR #74, opened 2022-05-23
Fix conda recipes

Ethyling:
- Simplify recipes using `{{ target_platform }}` for `sysroot` package and `compiler()` function in `ignore_run_exports_from`
- Add missisng `ignore_run_exports_from`


Ethyling:
@gpucibot merge

PR #82, opened 2022-06-07
Support host memory

### LINKED ISSUE #76: Writing a mix of host & device frames doesn't work
Currently if a user tries to write a mixture of host & device frames, this doesn't work as the following error will show up for any host frames

https://github.com/rapidsai/kvikio/blob/7a43759ae8638c17a02f68f4e8d777e52480fe37/python/kvikio/_lib/libkvikio.pyx#L86

This matters as using KvikIO in Distributed & Dask-CUDA will depend on handling some host frames along the way while [reading]( https://github.com/rapidsai/dask-cuda/blob/8df1507490d5c89ba2d880aece5eb508970b3433/dask_cuda/disk_io.py#L209-L213 ) and [writing]( https://github.com/rapidsai/dask-cuda/blob/8df1507490d5c89ba2d880aece5eb508970b3433/dask_cuda/disk_io.py#L166-L171 ).

Maybe there could be some kind of fallback to handle host memory in these cases?

madsbk:
Implementing support of host memory IO and some clean up of the POSIX code. 

Closes #76.

cc. @jakirkham


madsbk:
rerun tests

madsbk:
rerun tests

jakirkham:
rerun tests

cpp/examples/basic_io.cpp
```diff
@@ -33,6 +33,9 @@ void check(bool condition)
   }
 }
 
+constexpr int NELEM = 1024;
```
jakirkham, line 36: Should we add a comment explaining this value? In particular why 1024?

python/kvikio/_lib/libkvikio.pyx
```diff
@@ -75,15 +75,15 @@ def task_size_reset(nthreads: int) -> None:
     kvikio_cxx_api.task_size_reset(nthreads)
 
 
-cdef pair[uintptr_t, size_t] _parse_buffer(buf, size):
+cdef pair[uintptr_t, size_t] _parse_buffer(buf, size, accept_host_buffer) except *:
```
jakirkham, line 78: Should we try to type these?

```suggestion
cdef pair[uintptr_t, size_t] _parse_buffer(buf, size_t size, bint accept_host_buffer) except *:
```

Note: We may also want to do this elsewhere.

python/kvikio/cufile.py
```diff
@@ -197,13 +196,13 @@ def read(self, buf, size: int = None, file_offset: int = 0, task_size=None) -> i
...
         Notes
         -----
-        cuFile can only make use of GPUDirect Storage for reads that are
-        aligned to a page boundary. The GPU page size used by cuFile is 4kB, so
-        all reads must be at an offset that is a multiple of 4096 bytes. If the
-        desired `file_offset` is not a multiple of 4096 it is likely desirable
-        to round down to the nearest multiple of 4096 and discard any undesired
-        bytes from the resulting data. Similarly, it is optimal for `size` to
-        be a multiple of 4096 bytes.
+        When reading into device memory, cuFile can only make use of GDS for reads
```
jakirkham, line 199: Same thought as above except reading into pinned memory and copying to device memory.

python/kvikio/cufile.py
```diff
@@ -230,13 +229,13 @@ def write(self, buf, size: int = None, file_offset: int = 0, task_size=None) ->
...
         Notes
         -----
-        cuFile can only make use of GPUDirect Storage for writes that are
-        aligned to a page boundary. The GPU page size used by cuFile is 4kB, so
-        all reads must be at an offset that is a multiple of 4096 bytes. If the
-        desired `file_offset` is not a multiple of 4096 it is likely desirable
-        to round down to the nearest multiple of 4096 and discard any undesired
-        bytes from the resulting data. Similarly, it is optimal for `size` to
-        be a multiple of 4096 bytes.
+        When writing from device memory, cuFile can only make use of GDS for writes
```
jakirkham, line 232: Wonder if we can make a fallback (for when GDS is not present) where it copies from device memory to pinned memory and then writes out. The allocation cost could be amortized away by reusing allocations from a pinned memory pool.

python/tests/test_basic_io.py
```diff
@@ -28,14 +29,15 @@ def check_bit_flags(
 @pytest.mark.parametrize("size", [1, 10, 100, 1000, 1024, 4096, 4096 * 10])
 @pytest.mark.parametrize("nthreads", [1, 3, 4, 16])
 @pytest.mark.parametrize("tasksize", [199, 1024])
-def test_read_write(tmp_path, size, nthreads, tasksize):
+@pytest.mark.parametrize("memtype", [cupy, numpy])
```
jakirkham, line 32: Wonder if we should use `xp` to match CuPy and others doing this

```suggestion
@pytest.mark.parametrize("xp", [cupy, numpy])
```

jakirkham (commented):
Thanks Mads! 🙏

This looks good. Had a couple comments and questions below 🙂

cpp/examples/basic_io.cpp
madsbk: This is just an arbitrary value, added a comment

jakirkham: Ok wasn't sure if it was alignment related :)

python/tests/test_basic_io.py
madsbk: Fixed

python/kvikio/cufile.py
madsbk: We actually do this, updated the docs

madsbk:
> Thanks Mads! pray
> 
> This looks good. Had a couple comments and questions below slightly_smiling_face

Thanks @jakirkham, I think I have addressed all of your comments.

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Mads! 🙏

PR #85, opened 2022-06-13
FindcuFile now searches in the current CUDA Toolkit location

robertmaynard:
We shouldn't presume that the CUDA Toolkit is always installed at /usr/local/cuda. Instead we use use `find_package(CUDAToolkit)` to find the correct location.

In addition since `cufile::cuFile_interface` is part of our public interface we also install the `FindcuFile.cmake` file.

### DIFF
--- a/cpp/cmake/Modules/FindcuFile.cmake
+++ b/cpp/cmake/Modules/FindcuFile.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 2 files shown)


jakirkham:
rerun tests

cpp/cmake/Modules/FindcuFile.cmake
```diff
@@ -54,22 +54,24 @@ pkg_check_modules(PKG_cuFile QUIET cuFile)
 set(cuFile_COMPILE_OPTIONS ${PKG_cuFile_CFLAGS_OTHER})
 set(cuFile_VERSION ${PKG_cuFile_VERSION})
 
+#Find the location of the CUDA Toolkit
```
madsbk, line 57:
```suggestion
# Find the location of the CUDA Toolkit
```

madsbk (approved):
LGTM, thanks @robertmaynard

madsbk:
@gpucibot merge

PR #87, opened 2022-06-14
dlopen: use already loaded symbols if available

madsbk:
CI is [failing](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fkvikio%2Fprb%2Fkvikio-gpu-build/detail/kvikio-gpu-build/41/pipeline#log-1560) because `dlopen("libcuda.so")` is finding a stub (`/usr/local/cuda/lib64/stubs/libcuda.so`). 

This PR fixes the issue by trying to use already loaded symbols. However, this could be a problem if the symbols are unloaded later in the execution, unlikely but possible. 

Is this the best way to handle such an issue or is there a better approach?

cc. @jakirkham @robertmaynard

### DIFF
--- a/cpp/include/kvikio/shim/cuda.hpp
+++ b/cpp/include/kvikio/shim/cuda.hpp
@@ -45,7 +45,15 @@ class cudaAPI {
  private:
   cudaAPI()
   {
-    void* lib = load_library("libcuda.so");
+    // First we check if the symbols are already loaded.
+    void* lib = load_library(nullptr);
+    try {
+      get_symbol(Init, lib, KVIKIO_STRINGIFY(cuInit));
+    } catch (const std::runtime_error& e) {
+      // If not, we load them.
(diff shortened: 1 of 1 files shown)


robertmaynard:
I wonder if we can ignore this issue. Currently the RAPIDS container images are symlinking the stub image so it gets loaded when it shouldn't.

madsbk:
> I wonder if we can ignore this issue. Currently the RAPIDS container images are symlinking the stub image so it gets loaded when it shouldn't.

That would be great :)

jakirkham:
Agree with Robert. We should also be able to restart CI on failing PRs at this point

madsbk:
rerun tests

vyasr:
I think this can be closed assuming that we're unblocked on CI now?

PR #89, opened 2022-07-08
minor docstring fixes

grlee77:
Just fixing a few issues that I noticed in the docstrings.

Default task size currently seems to be [4 MB rather than 16 MB](https://github.com/rapidsai/kvikio/blob/bc7a313c68c677189ff7cd78bd4e58e16a587bda/cpp/include/kvikio/defaults.hpp#L106).


madsbk (approved):
Thanks @grlee77

madsbk:
@gpucibot merge

PR #91, opened 2022-07-21
add Notes related to page-aligned read/write methods to the Python docstrings

grlee77:
I initially did not realize the importance of aligning the offsets to 4096-byte boundaries and was experiencing slower than expected performance. It is likely to be helpful to advertise the necessity of this alignment more prominently as proposed here.


madsbk (approved):
Thanks @grlee77 !

madsbk:
@gpucibot merge

PR #101, opened 2022-08-01
Require `python` in `run`

jakirkham:
When `python` is a dependency, it needs to be listed in `host` & `run`. This aligns the version used when installing and makes sure `python` gets installed (and the right `kvikio` package with it) at runtime.

cc @madsbk @charlesbluca


madsbk (approved):
Thanks @jakirkham

madsbk:
@gpucibot merge

Issue #106, opened 2022-08-02
Use `rapids_cpm_nvcomp` in `get_nvcomp.cmake`


vyasr:
Our [`get_nvcomp.cmake`](https://github.com/rapidsai/kvikio/blob/branch-22.08/python/cmake/thirdparty/get_nvcomp.cmake) should be updated to use [`rapids_cpm_nvcomp` from `rapids-cmake`](https://github.com/rapidsai/rapids-cmake/blob/branch-22.08/rapids-cmake/cpm/nvcomp.cmake). That will keep kvikio from getting out of sync with the rest of RAPIDS and better ensure that it stays up to date. I believe that there have been a lot of relevant bugfixes/improvements (for cudf) to nvcomp in the last couple of releases.

_Originally posted by @vyasr in https://github.com/rapidsai/kvikio/issues/96#issuecomment-1203085933_

thomcom:
I used the .cmake file you recommended @vyasr and will update this with a PR later.

vyasr:
nvcomp has been entirely removed from kvikio so this issue is no longer relevant.

PR #117, opened 2022-08-05
Changed .cmake to use rapids_cpm_nvcomp

thomcom:
Resolves https://github.com/rapidsai/kvikio/issues/106. This is a breaking change and completely breaks python building as the nvCOMP API surface has changed completely. That fix is coming up next week in another PR. Most likely I'll just include this fix in that fix, since they are dependent.

### DIFF
--- a/python/cmake/thirdparty/get_nvcomp.cmake
+++ b/python/cmake/thirdparty/get_nvcomp.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2021-2022, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


python/cmake/thirdparty/get_nvcomp.cmake
```diff
@@ -13,28 +13,19 @@
...
-    nvcomp ${VERSION}
-    GLOBAL_TARGETS nvcomp::nvcomp
-    CPM_ARGS GITHUB_REPOSITORY NVIDIA/nvcomp
-    GIT_TAG c435afaf4ba8a8d12f379d688effcb185886cec1
-    OPTIONS "BUILD_STATIC ON" "BUILD_TESTS OFF" "BUILD_BENCHMARKS OFF" "BUILD_EXAMPLES OFF"
+  include(${rapids-cmake-dir}/cpm/nvcomp.cmake)
+  rapids_cpm_nvcomp(
+    BUILD_EXPORT_SET cuspatial-exports
+    INSTALL_EXPORT_SET cuspatial-exports
+    USE_PROPRIETARY_BINARY ${CUDF_USE_PROPRIETARY_NVCOMP}
```
robertmaynard, line 22: This line looks to be a renmant of copy and paste from CUDF. 

Either kvikio should add a cmake option like `KVIKIO_USE_PROPRIETARY_NVCOMP` or you should have a hardcoded value. The reason that cudf provides an option is that nvcomp is closed source and therefore comes with the nvidia closed source license. Therefore users have the choice to not agree to this license and use an older version.

robertmaynard (changes requested):

python/cmake/thirdparty/get_nvcomp.cmake
thomcom: This is something very relevant to the work I'm doing in https://github.com/rapidsai/kvikio/pull/120. I need to support the secret build which is currently 2.3.3, and the public build which is currently 2.2.0. I'm not sure now to disambiguate them or support the difference.

robertmaynard: cudf uses the C++ macro `__has_include` to determine what support the provided nvcomp has.
https://github.com/rapidsai/cudf/blob/branch-22.10/cpp/src/io/comp/nvcomp_adapter.cpp#L24

thomcom: I guess I'll just set `USE_PROPRIETARY_BINARY` to False for now? I don't understand what context I'd want to use the proprietary binary or not, and I don't want to move a licensing agreement onto the kvikio users.

robertmaynard: If you want to use 2.3+ you will need to enable the proprietary binary.

thomcom: Is `cudf` configured to use the proprietary binary indefinitely?

robertmaynard: yes

Issue #96, opened 2022-07-30
nvcomp LZ4 compression Issue


ajs-88:
Hi,

I'm trying to compress a file using the nvcomp LZ4Compressor but ended up gettting file more than the actual size.
Kindly advise what i'm doing wrong.

```python
import os
import cupy as cp
import numpy as np
import kvikio.nvcomp as nvc
import humanize

DTYPE = cp.int32
dtype = cp.dtype(DTYPE)

filename = '/home/arul/Downloads/kjv10.txt'
print('File :', filename, ' Size: ', humanize.naturalsize(os.path.getsize(filename)))
testfile = open(filename).read()
data = np.frombuffer(bytes(testfile, 'utf-8'), dtype=np.int8)

data_gpu = cp.array(data, dtype=DTYPE)
compressor = nvc.LZ4Compressor(data_gpu.dtype)
compressed = compressor.compress(data_gpu)
gpu_comp_file = open('/home/arul/Downloads/kjv10-gpu-compressed.txt.lz4', 'wb')
gpu_comp_file.write(compressed.tobytes())
gpu_comp_file.close()

print('Compressed Size: ', humanize.naturalsize(compressed.size))
del compressor
del compressed
```

Result:
File : /home/arul/Downloads/kjv10.txt  Size:  4.4 MB
Compressed Size:  17.4 MB

jakirkham:
cc @thomcom

ajs-88:
Dear guys,

Any help / advise will be greatly appreciated.

ajs-88:
and also, when i use CascadedCompressor on the same program, the result is different.

`compressor = nvc.CascadedCompressor(data_gpu.dtype)`

File : /home/arul/Downloads/kjv10.txt  Size:  4.4 MB
Compressed Size:  5.8 MB

madsbk:
What is the size of the uncompressed memory?
```python
print('Uncompressed Size: ', humanize.naturalsize(data_gpu.nbytes))
```
Also, is `/home/arul/Downloads/kjv10.txt` available somewhere?

ajs-88:
Hi @madsbk,

I used the same file from the notebook example,
https://github.com/rapidsai/kvikio/blob/branch-22.08/notebooks/nvcomp_vs_zarr_lz4.ipynb

http://textfiles.com/etext/NONFICTION/kjv10.txt

[...]

PR #120, opened 2022-08-10
Update nvComp bindings to 2.3.3

thomcom:
This PR updates the python bindings for nvcomp. It gets rid of all of the previous bindings and replaces them with a layer wrapped around nvcomp's C++ `Manager` hierarchy. Currently supported are the six C++ algorithms that come with nvcomp:

- ANS
- Bitcomp
- Cascaded
- Gdeflate
- LZ4
- Snappy

Tests are also written for every `nvcompManagerBase` function and their implementations for the three algorithms. The decompression manager created through the `nvcompManagerFactory::create_manager` is also implemented.

Usage for the algorithms is very easy:

```py
compressor_instance = LZ4Manager()
compressed = compressor_instance.compress(cupy.arange(1_000_000))
decompressed = compressor_instance.decompress(compressed)
```

The compressors all work with a set of default configuration options, but can also have their options configured by passing them in as validly named arguments for the base type. This is with the exception of the `cudaStream` argument that I had some trouble with casting for. It segfaults if I send anything except a `cudaStream_t{0}` value. I'm not planning on fixing it at this time.

Finally I included a command-line tool `gpu_compressor` that exposes the GPU apis to command line use within a valid Kvikio build environment. It uses Kvikio CuFile for I/O and Nvcomp for compression. It is useful for testing, and I know that @trxcllnt is planning on devising a packaged binary that includes any environment needed so we can distribute it.

Using ANS, results on a 2GB .csv file are extremely fast:
```py
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -c ans taxi2015.csv -v
GPU Compression Initialized
File read time: 0.451 seconds.
Compressing 1985964692 bytes
Compression time: 0.0336 seconds
File write time: 0.879 seconds
Compressed file size 957416516
Created file taxi2015.csv.gpc
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -d taxi2015.csv.gpc -o output -v
GPU Compression Initialized
File read time: 0.242 seconds.
Decompressing 957416516 bytes
Decompression time: 0.00965 seconds
File write time: 1.63 seconds
Decompressed file size 1985964692
Created file output
```

Bitcomp:

```py
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -c bitcomp taxi2015.csv -v
GPU Compression Initialized
File read time: 0.438 seconds.
Compressing 1985964692 bytes
Compression time: 0.0642 seconds
File write time: 1.33 seconds
Compressed file size 1363618088
Created file taxi2015.csv.gpc
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -d taxi2015.csv.gpc -o output -v
GPU Compression Initialized
File read time: 0.333 seconds.
Decompressing 1363618088 bytes
Decompression time: 0.0198 seconds
File write time: 1.85 seconds
Decompressed file size 1985964692
Created file output
```

Cascaded:

```py
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -c cascaded taxi2015.csv -v
GPU Compression Initialized
File read time: 0.438 seconds.
Compressing 1985964692 bytes
Compression time: 0.072 seconds
File write time: 1.77 seconds
Compressed file size 1983588600
Created file taxi2015.csv.gpc
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -d taxi2015.csv.gpc -o output -v
GPU Compression Initialized
File read time: 0.464 seconds.
Decompressing 1983588600 bytes
Decompression time: 0.0133 seconds
File write time: 1.68 seconds
Decompressed file size 1985964692
Created file output
```

Gdeflate:
```py
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -c gdeflate taxi2015.csv -v
GPU Compression Initialized
File read time: 0.44 seconds.
Compressing 1985964692 bytes
Compression time: 1.06 seconds
File write time: 1.01 seconds
Compressed file size 748221792
Created file taxi2015.csv.gpc
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -d taxi2015.csv.gpc -o output -v
GPU Compression Initialized
File read time: 0.196 seconds.
Decompressing 748221792 bytes
Decompression time: 0.0149 seconds
File write time: 1.66 seconds
Decompressed file size 1985964692
Created file output
```

Results on a 2GB .csv file using LZ4 are as follows:
```py
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py taxi2015.csv -v
GPU Compression Initialized
File read time: 0.437 seconds.
Compressing 1985964692 bytes
Compression time: 1.58 seconds
File write time: 1.05 seconds
Compressed file size 1028511262
Created file taxi2015.csv.gpc
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -d taxi2015.csv.gpc -o output -v
GPU Compression Initialized
File read time: 0.253 seconds.
Decompressing 1028511262 bytes
Decompression time: 0.00943 seconds
File write time: 1.85 seconds
Decompressed file size 1985964692
Created file output
```

Results are similar for `Snappy` in time and compression ratio. 
```py
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -c snappy taxi2015.csv -v
GPU Compression Initialized
File read time: 0.433 seconds.
Compressing 1985964692 bytes
Compression time: 0.519 seconds
File write time: 1.04 seconds
Compressed file size 994736305
Created file taxi2015.csv.gpc
(rapids) rapids@compose:~/cuspatial/kvikio/python/cli$ python gpu_compressor.py -d taxi2015.csv.gpc -o output -v
GPU Compression Initialized
File read time: 0.243 seconds.
Decompressing 994736305 bytes
Decompression time: 0.0107 seconds
File write time: 1.88 seconds
Decompressed file size 1985964692
Created file output
```

- ANS achieves 2x compression in only 33 msec.
- Bitcomp gets 1.5x compression in 65 msec.
- Cascaded does not appear to compress the input using the default configuration.
- Gdeflate achives 2.6 compression in 1 second.
- LZ4 is 1.8x compression in 1.6 seconds.
- Snappy is 2x compression in 519 msec.

Ready for review.

Fixes https://github.com/rapidsai/kvikio/issues/96


quasiben:
Will this PR close https://github.com/rapidsai/kvikio/issues/96 ?

thomcom:
Correct, thanks for linking it in.

jakirkham:
Added that to the OP. Hope that is ok :)

thomcom:
Yes everything is going swimmingly over here. :D
Wrote LZ4 already, working on parallel classes to LZ4 now. Don't plan on supporting GDeflate or Bitcomp at this time, and I'm not sure how I'm going to navigate the 2.2.0 vs. 2.3.3 build situation yet, but for 2.2.0 it is looking grand.

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -24,24 +24,42 @@ from enum import Enum
...
+    nvcompBatchedCascadedDefaultOpts,
+    nvcompBatchedLZ4Opts_t,
+    nvcompBatchedLZ4DefaultOpts,
+    nvcompBatchedLZ4CompressGetTempSize,
+    nvcompBatchedLZ4CompressGetMaxOutputChunkSize,
+    nvcompBatchedLZ4CompressAsync,
+    nvcompBatchedLZ4DecompressGetTempSize,
+    nvcompBatchedLZ4DecompressGetTempSizeEx,
+    nvcompBatchedLZ4DecompressAsync,
+    nvcompBatchedLZ4GetDecompressSizeAsync
```
thomcom, line 62: Going to drop all the low level methods tomorrrow.

thomcom:
Ready for review.

python/tests/test_nvcomp.py
```diff
@@ -10,60 +10,358 @@
...
+            {
+                "num_chunks": 1,
+                "decomp_data_size": 10000,
+            },
+        ],
+    ),
+)
+def test_get_decompression_config_with_default_options(compressor_size):
+    compressor = compressor_size[0]
+    expected = compressor_size[1]
```
madsbk, line 245:
```suggestion
@pytest.mark.parametrize(
    "compressor,expected",
    zip(
        [
            libnvcomp.LZ4Manager,
            libnvcomp.CascadedManager,
            libnvcomp.SnappyManager,
        ],
        [
            {
                "num_chunks": 1,
                "decomp_data_size": 10000,
            },
            {
                "num_chunks": 3,
                "decomp_data_size": 10000,
            },
            {
                "num_chunks": 1,
                "decomp_data_size": 10000,
            },
        ],
    ),
)
def test_get_decompression_config_with_default_options(
    compressor, expected
):
```

python/tests/test_nvcomp.py
```diff
@@ -10,60 +10,358 @@
...
+            libnvcomp.LZ4Manager,
+            libnvcomp.CascadedManager,
+            libnvcomp.SnappyManager,
+        ],
+        [252334080, 1641600, 67311200],
+    ),
+)
+def test_get_required_scratch_buffer_size(compressor_size):
+    compressor = compressor_size[0]
+    expected = compressor_size[1]
```
madsbk, line 299:
```suggestion
@pytest.mark.parametrize(
    "compressor,expected",
    zip(
        [
            libnvcomp.LZ4Manager,
            libnvcomp.CascadedManager,
            libnvcomp.SnappyManager,
        ],
        [252334080, 1641600, 67311200],
    ),
)
def test_get_required_scratch_buffer_size(compressor, expected):
```

python/tests/test_nvcomp.py
```diff
@@ -10,60 +10,358 @@
...
+            libnvcomp.LZ4Manager,
+            libnvcomp.CascadedManager,
+            libnvcomp.SnappyManager,
+        ],
+        [401, 624, 3556],
+    ),
+)
+def test_get_compressed_output_size(compressor_size):
+    compressor = compressor_size[0]
+    expected = compressor_size[1]
```
madsbk, line 328:
```suggestion
@pytest.mark.parametrize(
    "compressor,expected",
    zip(
        [
            libnvcomp.LZ4Manager,
            libnvcomp.CascadedManager,
            libnvcomp.SnappyManager,
        ],
        [401, 624, 3556],
    ),
)
def test_get_compressed_output_size(compressor, expected):
```

madsbk (changes requested):
Overall looks good, this work @thomcom.
Only have some minor suggestions.

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +61,150 @@ class pyNvcompType_t(Enum):
...
+cdef class _CascadedManager(_nvcompManager):
+    def __cinit__(
         self,
-        max_chunk_size,
-        max_compressed_size,
-        format_opts
+        _options,
+        user_stream,
+        const int device_id,
     ):
```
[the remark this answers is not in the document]
thomcom: How do you feel about raising a `NotImplementedError` instead of a `ValueError`?

python/kvikio/nvcomp.py
```diff
@@ -79,147 +107,240 @@ def compress(self, data: cp.ndarray) -> cp.ndarray:
...
-            data,
-            data_size,
-            self.compress_temp_buffer,
-            self.compress_temp_size,
-            self.compress_out_buffer,
-            self.gpu_out_size,
-            self.s.ptr,
-        )
-        return self.compress_out_buffer[: self.compress_out_size[0]]
+    def set_scratch_buffer(self, new_scratch_buffer: cp.ndarray) -> None:
```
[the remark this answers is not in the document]
thomcom: Are you asking if the underlying C++ has a danger of freeing the scratch buffer that we're specifying here?

Or are you asking if the underlying library is going to take a piece of ownership of the scratch buffer and prevent it from being freed?

thomcom:
Mads I committed all your suggestions and added some xfail tests that take device_id and stream.

thomcom:
One thing needs to be addressed with this build, that is if it works with both the `KVIKIO_USE_PROPRIETARY_BINARY` set to true or false. I'll let you know when I've determined if it is going to work as expected.

python/kvikio/_lib/libnvcomp.pyx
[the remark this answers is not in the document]
madsbk: > How do you feel about raising a `NotImplementedError` instead of a `ValueError`?

Sounds good :)

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
madsbk: I was just wondering if users should be able to unset the scratch buffer and use it for something else (or free it) without having to delete the `nvCompManager`.

thomcom:
Ok, Cmake is updated to build with the 2.3.3 proprietary bindings. The python bindings work with either 2.2.0 or 2.3.3!

This is ready for final review unless you want me to work on adding the third party compression libraries like ANS, GDeflate, or Bitcomp, or if you want me to drill down on the compatibility issues with user defined streams and device_id. I'll spend another 30-60 minutes on the user_stream and device_id issue today.

thomcom:
I enabled `device_id` but the cudaStream_t argument is super brittle and segfaults if I pass anything other than 0. I'm not going to mess with it until someone asks.

thomcom:
This is ready for final review.

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
thomcom: I can't make promises about what nvcomp does under the hood, but nowhere in the bindings do we take ownership of the scratch buffer. I assume that they don't either, since they're just taking a pointer as always.

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -23,23 +23,27 @@
...
-    __CascadedCompressor,
-    __CascadedDecompressor,
-    __LZ4Compressor,
-    __LZ4Decompressor,
+    ANSManager,
+    CascadedManager,
+    CompressionConfig,
+    DecompressionConfig,
+    LZ4Manager,
+    SnappyManager,
```
ndickson-nvidia, line 41: Just double-checking that gdeflate and bitcomp are intentionally excluded here.  I noticed that the equivalents in nvcomp.py raise NotImplementedError, but ANS does the same there, but ANS is in in this list and gdeflate and bitcomp aren't.

python/kvikio/nvcomp.py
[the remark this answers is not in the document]
eschmidt-nvidia: This is a persistent state that you're providing the NvcompManager. Basically nvcompManager will use this buffer for all future compress/decompress operations until you specify a new pointer. You're free to use it for something else in the same CUDA stream (it doesn't store state between kernels). If you free it, you'll synchronize the stream that nvcompManager's been using. Then you need to provide a new scratch buffer before before the next call. Currently there's no way to "clear" this variable (which would lead to nvcompManager allocating internally if necessary).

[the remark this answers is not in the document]
madsbk: Thanks for clarifying @eschmidt-nvidia

python/cmake/thirdparty/get_nvcomp.cmake
```diff
@@ -12,29 +12,22 @@
 # the License.
 # =============================================================================
 
+set(KVIKIO_USE_PROPRIETARY_BINARY ON)
```
robertmaynard, line 15: You might want to make this an option so a user can opt-out.

python/kvikio/_lib/libnvcomp.pyx
thomcom: Thanks @ndickson-nvidia . ANS was accidentally left in the import list after I tested to see if I "already had it". I did not. I could actually include the bindings for ANS, GDeflate, and Bitcomp in the python and it would throw library import errors if the user tried to use them without them being installed. I can't test them without having gone through all those processes yet, hence not having done so. I accidentally left ANSManager in after testing it.

eschmidt-nvidia: Could we provide an error such as "for 2.2.0, you need to install the closed source compressors or upgrade to nvCOMP 2.3"? (now all the compressors are included by default)

thomcom: All the compressors are in 2.3.3? Perhaps I made a mistake, I will circle around and recheck it because I didn't get the version properly updated from 2.2.0 to 2.3.3 until yesterday. I'd love to include all the compressors.

eschmidt-nvidia: Yes. When we moved to closed source we removed the need to install separate packages for some of the compressors.

python/cmake/thirdparty/get_nvcomp.cmake
thomcom: Presently I've decided to only support 2.3.3 as I'd need to rewrite a bunch of tests and determine what is running in order to swap out their variations. If you don't mind I'd like to commit as is and consider that down the road.

thomcom:
This is fully baked I think. Please ship it.

madsbk (approved):
LGTM, thanks @thomcom

thomcom:
@gpucibot merge

quasiben:
I suspect gpucibot can't merge because OPS is required to merge when ci things change like: `ci/checks/style.sh` .This file is empty fwif

thomcom:
That's really curious, here on my local branch the file is intact. I will look into it.

thomcom:
Ah, I had modified the permissions locally and accidentally committed that change.

jakirkham:
I have some review comments that I'd still like to share

thomcom:
I'll wait for your review!

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
+                <uint8_t*><uintptr_t>comp_buffer.data.ptr
+            )
+        )
+        self._decompression_config = make_shared[DecompressionConfig](
+            (move(partial.get()[0]))
+        )
+        return {
+            "decomp_data_size": self._decompression_config.get()[0].decomp_data_size,
+            "num_chunks": self._decompression_config.get()[0].num_chunks
+        }
```
jakirkham, line 116:
```suggestion
        cdef const DecompressionConfig* decompression_config_ptr = self._decompression_config.get()
        return {
            "decomp_data_size": decompression_config_ptr.decomp_data_size,
            "num_chunks": decompression_config_ptr.num_chunks
        }
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
         self,
-        in_arr,
-        in_bytes,
-        temp_arr,
-        temp_bytes,
-        out_arr,
-        out_bytes,
-        uintptr_t stream=0
+        decomp_buffer,
+        comp_buffer,
```
jakirkham, line 121:
```suggestion
        Array decomp_buffer,
        Array comp_buffer,
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
+        )
+        return {
+            "uncompressed_buffer_size": self._compression_config.get()[
+                0
+            ].uncompressed_buffer_size,
+            "max_compressed_buffer_size": self._compression_config.get()[
+                0
+            ].max_compressed_buffer_size,
+            "num_chunks": self._compression_config.get()[0].num_chunks
+        }
```
jakirkham, line 88: Could simplify by storing the pointer. Also the indexing by `0` can be skipped as Cython knows can do attribute access on pointers.

```suggestion
        cdef const CompressionConfig* compression_config_ptr = self._compression_config.get()
        return {
            "uncompressed_buffer_size": compression_config_ptr.uncompressed_buffer_size,
            "max_compressed_buffer_size": compression_config_ptr.max_compressed_buffer_size,
            "num_chunks": compression_config_ptr.num_chunks
        }
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
-            <size_t>in_bytes,
-            <void*>temp_ptr,
-            <size_t>temp_bytes,
-            <void*>out_ptr,
-            <size_t>out_bytes,
-            <cudaStream_t>stream)
-
+        self._impl.decompress(
+            <uint8_t*><uintptr_t>decomp_buffer.data.ptr,
+            <const uint8_t*><uintptr_t>comp_buffer.data.ptr,
```
jakirkham, line 125:
```suggestion
            <uint8_t*>decomp_buffer.ptr,
            <const uint8_t*>comp_buffer.ptr,
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
 
-# _Cascaded Compressor / Decompressor
-cdef class _CascadedCompressor:
-    cdef __CascadedCompressor* c
+cdef class _nvcompManager:
+    # Temporary storage for factory allocated manager to prevent cleanup
+    cdef shared_ptr[nvcompManagerBase] _mgr
+    cdef nvcompManagerBase* _impl
+    cdef shared_ptr[CompressionConfig] _compression_config
+    cdef shared_ptr[DecompressionConfig] _decompression_config
```
jakirkham, line 71:
```suggestion
    cdef shared_ptr[DecompressionConfig] _decompression_config
    # Other internal bits to keep objects alive
    cdef Array _scratch_buffer
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
+            <DecompressionConfig&>self._decompression_config.get()[0]
+        )
 
-# LZ4 Compressor / Decompressor
-cdef class _LZ4Compressor:
-    cdef __LZ4Compressor* c
+    def set_scratch_buffer(self, new_scratch_buffer):
+        return self._impl.set_scratch_buffer(
+            <uint8_t*><uintptr_t>new_scratch_buffer.data.ptr
+        )
```
jakirkham, line 132: Would suggest hanging onto this to ensure it stays alive.

Also do we need to do anything in the manager to release the old buffer?

```suggestion
    def set_scratch_buffer(self, Array new_scratch_buffer):
        self._scratch_buffer = new_scratch_buffer
        return self._impl.set_scratch_buffer(
            <uint8_t*>self._scratch_buffer.ptr
        )
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
-        self.c = new __LZ4Compressor()
+    def get_required_scratch_buffer_size(self):
+        return self._impl.get_required_scratch_buffer_size()
 
-    def __dealloc__(self):
-        del self.c
+    def get_compressed_output_size(self, comp_buffer):
+        return self._impl.get_compressed_output_size(
+            <uint8_t*><uintptr_t>comp_buffer.data.ptr
+        )
```
jakirkham, line 140:
```suggestion
    def get_compressed_output_size(self, Array comp_buffer):
        return self._impl.get_compressed_output_size(
            <uint8_t*>comp_buffer.ptr
        )
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
         self,
-        in_arr,
-        in_bytes,
-        temp_arr,
-        temp_bytes,
-        out_arr,
-        out_bytes,
-        uintptr_t stream=0
+        size_t uncomp_chunk_size,
+        user_stream,
```
jakirkham, line 147: Also should this default to `0` (as before)?

```suggestion
        uintptr_t user_stream,
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
+        const int device_id,
```
jakirkham, line 148: Why does it need to be `const`? Could it just be...?

```suggestion
        int device_id,
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
-    def __cinit__(self):
-        self.d = new __LZ4Decompressor()
-
-    def __dealloc__(self):
-        del self.d
+        self._impl = <nvcompManagerBase*>new ANSManager(
+            uncomp_chunk_size,
+            <cudaStream_t><void*>0,  # TODO
+            device_id
+        )
```
jakirkham, line 154: Would suggest raising for unsupported `user_stream` values as opposed to ignoring them. This way when they are implemented the error can be dropped and the rest of the code kept as-is

Also this can be applied in other compressors below

```suggestion
        if user_stream != 0:
            raise NotImplementedError(
                f"Only `user_stream` of `0` is supported currently. Got: {user_stream}"
            )
        self._impl = <nvcompManagerBase*>new ANSManager(
            uncomp_chunk_size,
            <cudaStream_t>user_stream,
            device_id
        )
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
-        in_arr,
-        in_bytes,
-        temp_arr,
-        temp_bytes,
-        out_arr,
-        out_bytes,
-        uintptr_t stream=0
+        nvcompType_t data_type,
+        int bitcomp_algo,
+        user_stream,
```
jakirkham, line 162:
```suggestion
        uintptr_t user_stream,
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
+        const int device_id
```
jakirkham, line 163:
```suggestion
        int device_id
```

python/kvikio/nvcomp.py
```diff
@@ -79,147 +121,315 @@ def compress(self, data: cp.ndarray) -> cp.ndarray:
...
-            self.compress_temp_size,
-            self.compress_out_buffer,
-            self.compress_out_size,
-            self.s.ptr,
+        self.config = self._manager.configure_compression(data_size)
+        self.compress_out_buffer = cp.zeros(
+            self.config["max_compressed_buffer_size"], dtype="uint8"
         )
-        return self.compress_out_buffer[: self.compress_out_size[0]]
+        size = self._manager.compress(data, self.compress_out_buffer)
```
jakirkham, line 128: Would coerce objects going into to `Array`. Also would need to add `from kvikio._lib.arr import asarray`

```suggestion
        size = self._manager.compress(asarray(data), asarray(self.compress_out_buffer))
```

python/kvikio/nvcomp.py
```diff
@@ -79,147 +121,315 @@ def compress(self, data: cp.ndarray) -> cp.ndarray:
...
-        self.decompress_out_buffer = cp.zeros(self.decompress_out_size, dtype=np.uint8)
-        self.decompressor.decompress_async(
-            data,
-            data_size,
-            self.decompress_temp_buffer,
-            self.decompress_temp_size,
-            self.decompress_out_buffer,
-            self.decompress_out_size,
-            self.s.ptr,
+        decomp_buffer = cp.zeros(
```
jakirkham, line 147:
```suggestion
        decomp_buffer = cp.empty(
```

python/kvikio/nvcomp.py
```diff
@@ -79,147 +121,315 @@ def compress(self, data: cp.ndarray) -> cp.ndarray:
...
+            "decomp_data_size": The size of each decompression chunk.
+            "num_chunks": The number of chunks that the decompressed data is returned
+            in.
+        }
         """
-        self.dtype = cp.dtype(dtype)
-        self.compressor = _lib._LZ4Compressor()
-        self.decompressor = _lib._LZ4Decompressor()
-        self.s = cp.cuda.Stream()
+        return self._manager.configure_decompression_with_compressed_buffer(data)
```
jakirkham, line 190:
```suggestion
        return self._manager.configure_decompression_with_compressed_buffer(asarray(data))
```

python/kvikio/nvcomp.py
```diff
@@ -79,147 +121,315 @@ def compress(self, data: cp.ndarray) -> cp.ndarray:
...
-        self.decompress_temp_size = np.zeros((1,), dtype=np.int64)
-        self.decompress_out_size = np.zeros((1,), dtype=np.int64)
-
-        self.decompressor.configure(
-            data,
-            data_size,
-            self.decompress_temp_size,
-            self.decompress_out_size,
-            self.s.ptr,
+        return self._manager.set_scratch_buffer(new_scratch_buffer)
```
jakirkham, line 219:
```suggestion
        return self._manager.set_scratch_buffer(asarray(new_scratch_buffer))
```

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
-            <void*>in_ptr,
-            <size_t>in_bytes,
-            <void*>temp_ptr,
-            <size_t>temp_bytes,
-            <void*>out_ptr,
-            <size_t>out_bytes,
-            <cudaStream_t>stream)
+        self._impl = <nvcompManagerBase*>new BitcompManager(
+            <nvcompType_t>data_type,
+            <int>bitcomp_algo,
```
jakirkham, line 167: As these already have these types, can skip the cast. This can be applied in a few other places

```suggestion
            data_type,
            bitcomp_algo,
```

jakirkham (commented):
Thanks Thomson! 🙏

Had a few suggestions to make it easier to extract metadata about arrays (pointers, sizes, etc.) by coercing objects to `Array` in the Python interface so that Cython can benefit from the faster C-style attribute access in the interface code.

There are some other smaller suggestions (like typing a few things, etc.).

Tried to include code suggestions where possible. Though may have missed instances where similar patterns could be applied.

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +62,146 @@ class pyNvcompType_t(Enum):
...
-
-    def __cinit__(self, nvcompType_t t, int num_RLEs, int num_deltas, bool use_bp):
-        self.c = new __CascadedCompressor(
-            t,
-            num_RLEs,
-            num_deltas,
-            use_bp
-        )
+cdef class _nvcompManager:
+    cdef nvcompManagerBase* _impl
```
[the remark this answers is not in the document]
thomcom: I would rather not. I appreciate your concern here, but the abstraction layer here is pretty sensitive. The complexity here is right on the edge of my comprehension of Cython's capabilities and weaknesses. In particular Cython doesn't support fully abstract virtual classes - anything instantiated in Cython has to have a real vtable. Secondly, Cython doesn't support classes that lack a default constructor. nvcomp C++ is based on the PIMPL pattern, so `nvcompManagerBase` is a purely virtual class, and it also lacks a default constructor.

Attempting to create a stack version of the `nvcompManagerBase` puts Cython into a binary failure mode that is hard to resolve: Either it cannot create a stack instance of `nvcompManagerBase` or it cannot create a concrete instance due to the lack of vtables. I can resolve the first error by dropping `except +` from the child class constructor method, which is not super desirable, but then Cython will fail because it wants to create a concrete instance.

I'm able to use `shared_ptr` with `CompressionConfig` and `DecompressionConfig` because they are not based on the PIMPL pattern. I'm also able to use `shared_ptr` with the last class, `ManagedManager`, because it is actually created purely in C++ inside of the `create_manager` factory function.

I don't doubt that there is some encantation that will make `shared_ptr` usable with `self._impl`, and if you can show me the pattern I'm happy to use it. Otherwise, fiddling with Cython builds could eat up days and not give me any fruit to show, or at best will produce the same result as what I already have.

I'm going to spend a little more time on it today, however, because you pointed out an issue that I created when I worked out the pattern that is used for `ManagedManager` - there used to be a `dealloc: del self._impl` but I dropped it when I created the `ManagedManager` because of a double free. That's not acceptable, so I will work out that fix at least.

My inclination is just to check if `self._mgr` is none, and if so, do the free, otherwise let `shared_ptr` handle it.

python/kvikio/nvcomp.py
```diff
@@ -79,147 +121,315 @@ def compress(self, data: cp.ndarray) -> cp.ndarray:
...
+        Parameters
+        ----------
+        comp_buffer: cp.ndarray
+            A GPU buffer that has been previously compressed.
+
+        Returns
+        -------
+        int
+        """
+        return self._manager.get_compressed_output_size(comp_buffer)
```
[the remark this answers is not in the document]
thomcom: This is already typed in mypy as `cp.ndarray`, do I need the `asarray` call?

python/kvikio/_lib/libnvcomp.pyx
```diff
@@ -57,305 +63,184 @@ class pyNvcompType_t(Enum):
...
+    def compress(self, decomp_buffer, comp_buffer):
+        self._impl.compress(
+            <const uint8_t*><uintptr_t>decomp_buffer.data.ptr,
+            <uint8_t*><uintptr_t>comp_buffer.data.ptr,
+            <CompressionConfig&>self._compression_config.get()[0]
+        )
+        size = self._impl.get_compressed_output_size(
+            <uint8_t*><uintptr_t>comp_buffer.data.ptr
         )
+        return size
```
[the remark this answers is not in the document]
thomcom: Because the buffer has already been allocated, and prior to completing compression it has an unknown size. A possible refactor would be for the library to create the compression storage buffer and then return it, sliced to the correct size, when compression is finished.

python/cli/gpu_compressor.py
```diff
@@ -0,0 +1,127 @@
...
+import time
+
+import cupy
+
+import kvikio
+import kvikio.nvcomp as nvcomp
+
+if __name__ == "__main__":
+
+    class NvcompParser(argparse.ArgumentParser):
```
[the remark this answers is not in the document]
thomcom: Maybe, this class is dedicated just to parsing the args of the `__main__` function here, I'm not sure how I would like to refactor `gpu_compressor.py` yet, if at all. Probably move all of the compressor options into their own class.

python/cli/gpu_compressor.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        Handle special case and show help on invalid argument
+        """
+
+        def error(self, message):
+            sys.stderr.write("\nERROR: {}\n\n".format(message))
+            self.print_help()
+            sys.exit(2)
+
+    parser = NvcompParser()
```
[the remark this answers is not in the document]
thomcom: I'm still not sure how to simplify this. `gpu_compressor.py` is a command line tool. The main thing it is missing is the ability to compress directories, etc, which I think makes it much more complicated. With regards to making it callable by another python module, it does not seem necessary to me. The usage for `kvikio.nvcomp` from within python is very simple:

```
import kvikio.nvcomp as nvcomp
compressor = nvcomp.LZ4Compressor()
result = compressor.compress(my_gpu_data)
```

python/kvikio/_lib/libnvcomp.pyx
thomcom: The scratch buffer is owned by the caller, `nvcomp.py`, I don't think it needs to be held here?

python/kvikio/_lib/libnvcomp.pyx
thomcom: @madsbk has this same conversation with @eschmidt-nvidia earlier. The C++ library has no ownership over the scratch buffer and the caller is free to do with it as they please. As above it is owned by the caller of `nvcomp.py`.

python/kvikio/_lib/libnvcomp.pyx
thomcom: It is const only because it is a copy of the C++ signature. C++ practice is, I thought, const all the things!

thomcom:
Ready to merge on your approval @jakirkham

python/tests/test_nvcomp.py
```diff
@@ -10,60 +10,458 @@
...
+    assert len(final) == 3548
+
+
+@pytest.mark.parametrize(
+    "compressor_size",
+    zip(
+        [
+            libnvcomp.LZ4Manager,
+            libnvcomp.CascadedManager,
+            libnvcomp.SnappyManager,
```
jakirkham, line 280: Did we want to support the full set of managers here?

jakirkham (commented):
Thanks Thomson! 🙏

Looks good. Had one question about one of the tests.

python/tests/test_nvcomp.py
thomcom: Thanks for reminding me, hitting that, then merging.

thomcom:
Weird @jakirkham the `asarray()` errors you expected are happening, but only on Ubuntu 20.04, and not on my local machine Ubuntu 20.04.

thomcom:
@gpucibot merge

thomcom:
Please merge when ready @madsbk

thomcom:
@gpucibot merge

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Thomson! 🙏

Issue #125, opened 2022-08-29
c++ `basic_io` example doesn't work out of the box for me


alsam:
Hello, the steps for reproducing:
```sh
gh repo clone rapidsai/kvikio
cd kvikio/cpp
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Debug
make
./basic_io
KvikIO defaults: 
  Compatibility mode: disabled
DriverProperties: 
  Version: 0.0
  Allow compatibility mode: true
  Pool mode - enabled: false, threshold: 4 kb
  Max pinned memory: 4294967292 kb
terminate called after throwing an instance of 'std::system_error'
```
debugging:
```sh
cgdb ./basic_io
...
bt
...
#7  0x00007ffff78a5fcd in __cxxabiv1::__cxa_throw (obj=<optimized out>, tinfo=0x555555597860 <typeinfo for std::system_error@GLIBCXX_3.4.11>, dest=0x7ffff78d5fd0 <std::system_error::~system_error()>)
    at /usr/src/debug/gcc/libstdc++-v3/libsupc++/eh_throw.cc:98
#8  0x0000555555564dd4 in kvikio::detail::open_fd (file_path="/tmp/test-file", flags="w", o_direct=true, mode=436) at /home/alsam/work/github/kvikio/cpp/examples/../include/kvikio/file_handle.hpp:85
#9  0x0000555555564f48 in kvikio::FileHandle::FileHandle (this=0x7fffffffe6c0, file_path="/tmp/test-file", flags="w", mode=436, compat_mode=false)
```
failed to open the file `/tmp/test-file` for writing with `o_direct=true, mode=436`

```sh
$ touch /tmp/test-file
$ echo $?
0
```
Thanks

madsbk:
Thanks for reporting the issue @alsam, what environment are you running in? This works fine on my workstation. 

KvikIO tries to open `/tmp/test-file` with the `O_DIRECT` flag, maybe that isn't supported? 
Can you try running `KVIKIO_COMPAT_MODE=ON  ./examples/basic_io` and try to change the `/tmp/test-file` file paths in [`basic_io.cpp`](https://github.com/rapidsai/kvikio/blob/branch-22.10/cpp/examples/basic_io.cpp).

alsam:
Thanks @madsbk for answering, you are quite right, with this patch:
```diff
diff --git a/cpp/examples/basic_io.cpp b/cpp/examples/basic_io.cpp
[...]

PR #126, opened 2022-09-01
Fall back to compat mode if we cannot open the file with `O_DIRECT`

madsbk:
Fixes https://github.com/rapidsai/kvikio/issues/125

Some filesystems such as `tmpfs` doesn't support the `O_DIRECT` flag. We now fall back to compat mode if we cannot open the file with `O_DIRECT`.

### DIFF
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -158,11 +158,16 @@ class FileHandle {
              const std::string& flags = "r",
              mode_t mode              = m644,
              bool compat_mode         = defaults::compat_mode())
-    : _fd_direct_on{detail::open_fd(file_path, flags, true, mode)},
-      _fd_direct_off{detail::open_fd(file_path, flags, false, mode)},
+    : _fd_direct_off{detail::open_fd(file_path, flags, false, mode)},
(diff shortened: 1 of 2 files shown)


madsbk:
@alsam can I ask you to run `basic_io` with this PR? 
It should now work without your patch and without `KVIKIO_COMPAT_MODE=ON`

alsam:
Thanks @madsbk it works fine, from scratch:
```sh
gh repo clone rapidsai/kvikio
gh pr checkout 126
cd cpp/
mkdir build
cd build/
cmake ..
make
examples/basic_io 
KvikIO defaults: 
  Compatibility mode: disabled
DriverProperties: 
  Version: 0.0
  Allow compatibility mode: true
  Pool mode - enabled: false, threshold: 4 kb
  Max pinned memory: 4294967292 kb
Write: 4000
Read:  4000
Parallel write (16 threads): 4000
Parallel write (16 threads): 4000
Parallel POSIX write (16 threads): 4000
Parallel POSIX read (16 threads): 4000
```

madsbk:
Great, thanks for testing!

quasiben (approved):

quasiben:
@gpucibot merge

PR #127, opened 2022-09-01
Set version when dlopen() cuda and cufile

madsbk:
In order to avoid ABI incompatibilities and loading of stub libraries, we now specify the major version when loading `libcuda.so.1` and `libcufile.so.1`.


quasiben (approved):

quasiben:
@gpucibot merge

Issue #119, opened 2022-08-09
Filter out Zarr's consolidated metadata (`.zmetadata`)


jakirkham:
Currently there are a few metadata keys that get handled specially in the `GDSStore` (see below).

https://github.com/rapidsai/kvikio/blob/0251229fb4bd60a1f3e686756a97c5e9f4b5b711/python/kvikio/zarr.py#L35-L39

Though one that isn't handled currently is [the consolidated metadata key (`.zmetadata`)]( https://github.com/zarr-developers/zarr-python/blob/8e7f4cc01764f50b54b4d90a70495bba16766c7d/zarr/storage.py#L2850 ), which can show up for some Zarr stores.

Would be good to filter this one out too

jakirkham:
cc @dcherian (who pointed this out to us :)

PR #131, opened 2022-09-13
Use the Zarr's new `getitems()` API

madsbk:
By using the new API in https://github.com/zarr-developers/zarr-python/pull/1131, we do not have to guess whether to read into host or device memory. That is, no more filtering of specify keys like:
```python
 if os.path.basename(fn) in [ 
     zarr.storage.array_meta_key, 
     zarr.storage.group_meta_key, 
     zarr.storage.attrs_key, 
 ]: 
```

Notice, this PR is on hold until Zarr v2.15 is released 

Closes https://github.com/rapidsai/kvikio/issues/119

UPDATE: Zarr v2.15 has been released


python/tests/test_zarr.py
```diff
@@ -10,58 +10,129 @@
 zarr = pytest.importorskip("zarr")
```
wence-, line 10: The top-level `cupy` importorskip will mean that if cupy is not available we will never run numpy-based tests. Is that what you want?

python/tests/test_zarr.py
```diff
@@ -10,58 +10,129 @@
...
-    module = pytest.importorskip(array_type)
-    a = module.arange(5, dtype="u1")
+    a = xp.arange(5, dtype="u1")
     store["a"] = a
     b = store["a"]
 
-    # Notice, GDSStore always returns a cupy array
-    assert type(b) is cupy.ndarray
-    cupy.testing.assert_array_equal(a, b)
+    # Notice, when not using getitems(), GDSStore returns bytes always
```
wence-, line 32:
```suggestion
    # Notice, unless using getitems(), GDSStore always returns bytes
```

python/tests/test_zarr.py
```diff
@@ -10,58 +10,129 @@
...
+    assert isinstance(b, bytes)
+    a.data == b
```
wence-, line 34: This equality test doesn't do anything (since there is no assertion).
```suggestion
    assert (xp.frombuffer(b, dtype="u1") == a).all()
```
?

python/tests/test_zarr.py
```diff
@@ -10,58 +10,129 @@
...
-    g = zarr.open_group(store, meta_array=cupy.empty(()))
+    g = zarr.open_group(store, meta_array=xp.empty(()))
     g.ones("data", shape=(10, 11), dtype=int, compressor=None)
     a = g["data"]
     assert a.shape == (10, 11)
     assert a.dtype == int
     assert isinstance(a, zarr.Array)
-    assert isinstance(a[:], cupy.ndarray)
+    assert isinstance(a.meta_array, type(xp.empty(())))
+    assert isinstance(a[:], xp.ndarray)
```
wence-, line 86: Why are both things not `isinstance(foo, xp.ndarray)`?

python/tests/test_zarr.py
```diff
@@ -10,58 +10,129 @@
...
+        shape=a.shape,
+        dtype=a.dtype,
+        chunks=(10,),
+        compressor=None,
+        meta_array=xp.empty(()),
+    )
+    z[:] = a
+    assert a.shape == z.shape
+    assert a.dtype == z.dtype
+    assert isinstance(a, type(z[:]))
```
wence-, line 105: Why do you need to slice `z` here? Ah, it's a zarr array so you must slice to get the backing store?

python/kvikio/zarr.py
```diff
@@ -15,49 +17,86 @@
...
+
+            for future, nbytes in io_results:
+                nbytes_read = future.get()
+                if nbytes_read != nbytes:
+                    raise RuntimeError(
+                        f"Incomplete read ({nbytes_read}) expected {nbytes}"
+                    )
+        finally:
+            for f in files:
+                f.close()
```
wence-, line 98: The idiomatic way to do this is...

```
import contextlib

...
with contextlib.ExitStack() as stack:
    for key in keys:
        f = stack.enter_context(kvikio.CuFile(filepath, "r"))
        ret[key] = ..
        io_results.append((f.pread(ret[key]), nbytes)
   for fut, nbytes in io_results:
        nbytes_read = fut.get()
        ...
```

wence-: I'll note that both the new code (and my proposed approach) leave dangling futures if the block produces an exception anywhere. I guess that is fine because that propagates and the program is hosed at that point anyway?

wence- (changes requested):
Thanks @madsbk, I think we can do a little bit of cleanup, to avoid repeating ourselves with the minimum supported version.

python/tests/test_zarr.py
madsbk: They are :)
fixed

python/tests/test_zarr.py
madsbk: yes

python/kvikio/zarr.py
```diff
@@ -15,49 +17,86 @@
 
 import kvikio
 import kvikio.nvcomp
-from kvikio._lib.arr import asarray
+import kvikio.zarr
```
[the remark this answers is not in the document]
madsbk: Fixed :)

python/kvikio/zarr.py
```diff
@@ -15,49 +17,86 @@
...
+        with kvikio.CuFile(fn, "w") as f:
+            written = f.write(a)
+            assert written == a.nbytes
+
+    def getitems(
+        self,
+        keys: Sequence[str],
+        *,
+        contexts: Mapping[str, Mapping] = {},
+    ) -> Mapping[str, Any]:
```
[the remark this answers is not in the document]
madsbk: Added docstring

python/kvikio/zarr.py
```diff
@@ -15,49 +17,86 @@
...
+        default_meta_array = numpy.empty(())
```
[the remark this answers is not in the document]
madsbk: done

python/tests/test_zarr.py
```diff
@@ -10,58 +10,129 @@
...
-    pytest.skip("requires Zarr v2.13+", allow_module_level=True)
+
+if not kvikio_zarr.supported():
+    pytest.skip("requires Zarr v2.15+", allow_module_level=True)
 
 
 @pytest.fixture
 def store(tmp_path):
     """Fixture that creates a GDS Store"""
+    cupy.arange(1)  # Making sure that CUDA has been initialized
```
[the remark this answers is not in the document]
madsbk: They should, removed

wence- (approved):
Thanks Mads.

wence- (approved):

wence-:
I think we need to wait for the other codeowners to approve too.

jjacobelli (approved):

jakirkham:
Thanks all! 🙏

Great work 🥳

PR #133, opened 2022-10-03
`is_host_memory()`: returns `true` when `CUDA_ERROR_NOT_INITIALIZED`

madsbk:
Instead of raising an exception when CUDA hasn't been initialized, `is_host_memory()` now assumes that the pointer is host memory.


quasiben (approved):

quasiben:
@gpucibot merge

PR #134, opened 2022-10-03
Update nvcomp's expected sizes when testing

madsbk:


madsbk:
@thomcom is it possible to avoid hard coding the expected sizes? Can we get the sizes from somewhere? Or maybe use a range?

Anyways, I think we should merge this to get CI passing

thomcom:
We could move them to a fixture in `conftest.py` but it doesn't improve the problem much. We're tracking very low level implementation details with these sizes: we could just drop them, but I think that the values are really nice-to-know.

I think most of these values have a single point of truth, so there's no extra API to validate them against. It wouldn't be unreasonable to move most or all to some kind of "extended tests" section that isn't run by CI?

thomcom:
@gpucibot merge

madsbk:
@gpucibot merge

madsbk:
Thanks @thomcom

quasiben (approved):

jakirkham:
Could we please raise this with the nvCOMP team? Maybe they can offer a better way of handling this so we don't need to hardcode things here

PR #146, opened 2022-11-17
Don't use CMake 3.25.0 as it has a FindCUDAToolkit show stopping bug

robertmaynard:
Also bring kvikio minimum required version inline with the rest of RAPIDS

### DIFF
--- a/conda/environments/kvikio_dev_cuda11.5.yml
+++ b/conda/environments/kvikio_dev_cuda11.5.yml
@@ -12,7 +12,7 @@ dependencies:
   - clang=11.1.0
   - clang-tools=11.1.0
   - cupy>=9.5.0,<12.0.0a0
-  - cmake>=3.20.1,!=3.23.0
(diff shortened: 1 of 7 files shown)


vyasr (changes requested):
This also needs to update the recipe meta.yaml files and the pyproject.toml file.

robertmaynard:
> This also needs to update the recipe meta.yaml files and the pyproject.toml file.

Corrected.

galipremsagar (commented):
Should we also update cmake minimum version in all Cmakelists.txt files? Similar to #147?

robertmaynard:
> Should we also update cmake minimum version in all Cmakelists.txt files? Similar to #147?

Good point. Done.

galipremsagar (approved):

ajschmidt8 (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks all! 🙏

PR #153, opened 2023-01-16
Support RDMA

madsbk:
By default, we now try to load `libcufile_rdma.so`. To control this use:
```bash
# The default, first try to load `libcufile_rdma.so` then `libcufile.so`
unset KVIKIO_RDMA or KVIKIO_RDMA="auto". 

# Load `libcufile_rdma.so` only
KVIKIO_RDMA=on. 

# Load `libcufile.so` only
KVIKIO_RDMA=off. 
```

### DIFF
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -23,54 +23,11 @@
 #include <string>
 #include <utility>
 
+#include <kvikio/env_var.hpp>
 #include <kvikio/shim/cufile.hpp>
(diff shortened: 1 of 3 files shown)


madsbk:
If userspace verbs access is available, cuFile and KvikIO should use it automatically. 
Run the GDS script `gdscheck.py -p` to get the current status of userspace RDMA.

madsbk:
In turns out, cuFile will dlopen `libcufile_rdma.so` by itself so this PR isn't needed.

PR #154, opened 2023-01-20
Use pre-commit for style checks.

bdice:
This PR updates kvikio to use `pre-commit` for style checks. This will align it with the rest of RAPIDS and help with the transition to GitHub Actions for CI. I modeled this repo's set of hooks on the set of hooks used for cudf, with a few removed because they aren't relevant for this repo.

Reviewers: Now that I have a working version with all pre-commit checks enabled, I would be happy to split this PR up and enable a few pre-commit hooks at a time. If you'd like to review each hook's changes separately, let me know and I'll break it into 2-3 PRs with fewer changes.

PR #156, opened 2023-01-23
nvcomp tests: update hardcoded lengths

madsbk:
@thomcom can we avoid hard coding the expected compression lengths?

### DIFF
--- a/python/tests/test_nvcomp.py
+++ b/python/tests/test_nvcomp.py
@@ -5,7 +5,6 @@ import pytest
 
 np = pytest.importorskip("numpy")
 cupy = pytest.importorskip("cupy")
-cudf = pytest.importorskip("cudf")
 kvikio = pytest.importorskip("kvikio")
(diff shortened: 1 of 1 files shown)


thomcom:
Seems like we could expect a range, or a maximum, or just not test them. 

1. Testing a range is like having hardcoded values with a lower frequency of changes required
2. Testing a maximum allows the results to improve in any way but makes us do leg work if the upstream team produces something with lower quality, potentially allowing us to flag it for them
3. Not testing the results leaves it in the hands of the NVCOMP team.

What do you suggest?

madsbk:
Thanks @thomcom, we are now using `np.testing.assert_allclose(actual, desired, rtol=0.1)` when comparing sizes. I think that is sufficient for now.

bdice:
Can this PR move forward? It is blocking #154 which in turn blocks GitHub Actions migrations. cc: @ajschmidt8

quasiben (approved):

Issue #159, opened 2023-01-26
Crash/segfault on exit when running libcudf tests with kvikIO and CUDA 11.8


vuule:
Getting inconsistent errors at the end of process lifetime when "KVIKIO" policy is used in cuIO.
No issues in the actual tests.
The are multiple possible outcomes: crash, segfault, success. Sample output:
```
30: double free or corruption (!prev)
1/3 Test #30: JSON_TEST ........................Subprocess aborted***Exception:   2.68 sec
```

Repro persists when all cuFile calls outside of kvikIO are removed.

No repro with other versions of CUDA toolkit.
No repro when using libcudf's cuFile path instead of kvikIO.

PR #160, opened 2023-01-26
Manually open and close the cuFile driver

madsbk:
cuFile is supposed to open and close the driver automatically but because of a bug in CUDA 11.8, it sometimes segfault.
Closes https://github.com/rapidsai/kvikio/issues/159

cc. @vuule

### DIFF
--- a/cpp/include/kvikio/shim/cufile.hpp
+++ b/cpp/include/kvikio/shim/cufile.hpp
@@ -17,6 +17,7 @@
 
 #include <stdexcept>
 
+#include <iostream>
 #include <kvikio/shim/cufile_h_wrapper.hpp>
 #include <kvikio/shim/utils.hpp>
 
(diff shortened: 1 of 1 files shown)


cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -62,6 +63,23 @@ class cuFileAPI {
...
+                               KVIKIO_STRINGIFY(__LINE__) + ": " +
+                               cufileop_status_error(error.err));
+    }
+  }
+  ~cuFileAPI()
+  {
+    CUfileError_t const error = DriverClose();
+    if (error.err != CU_FILE_SUCCESS) {
+      std::cerr << "Unable to close GDS file driver: " << cufileop_status_error(error.err)
+                << std::endl;
```
vuule, line 81: is this a temporary change?

vuule (approved):
looks good except for the suspicious `cerr` output

madsbk:
Notice, the CI error is unrelated and fixed in https://github.com/rapidsai/kvikio/pull/156

cpp/include/kvikio/shim/cufile.hpp
madsbk: No, to avoid throwing an exception in the destructor, we print a warning instead

vuule: Right, no exceptions in the dtor.
Posted the comment because libcudf does not do any std output, thought the same might hold for kvikio

quasiben (approved):

quasiben:
rerun tests

PR #183, opened 2023-03-14
Legate Support

madsbk:
Support of [Legate stores](https://github.com/nv-legate/legate.core) and [cuNumeric arrays](https://github.com/nv-legate/cunumeric).

## Install 
Make sure cuNumeric is installed and then build like:
```
cd legate
pip install .
```
## Usage
The API is very similar to regular KvikIO:
```python
import cunumeric as num
from legate_kvikio import CuFile
from legate.core import get_legate_runtime

a = num.arange(100)
with CuFile("/tmp/my-file", "w") as f:
    f.write(a)

    # In order to make sure the file has been written before the following
    # read execute, we insert a fence between the write and read.
    # Notice, this call isn't blocking.
    get_legate_runtime().issue_execution_fence(block=False)
    
    b = num.empty_like(a)
    f.read(b)

# In order to make sure the file has been written before re-opening
# it for reading, we block the execution.
get_legate_runtime().issue_execution_fence(block=True)

c = num.empty_like(a)
with kvikio.CuFile("/tmp/my-file", "r") as f:
    f.read(c)
print("sum: ", c.sum())
```
Run using Legate launcher (using 10GB device and host memory) like:
```bash
legate --sysmem 10000 --fbmem 10000 --cpus 2 --gpus 2 my_io_script.py
```


quasiben (approved):
preemptively approving to allow folks to move quickly here

legate/legate_kvikio/core.py
```diff
@@ -0,0 +1,102 @@
...
+
+def read(path: str, obj: Any) -> None:
+    """Read data from disk using KvikIO
+
+    Args:
+        obj (Any): A Legate store or any object implementing
+                     the Legate data interface.
+    """
+    output = _get_legate_store(obj)
+    task = user_context.create_auto_task(TaskOpCode.READ)
```
manopapad, line 98: I think this task would be more appropriate as an unbounded-output task, similar to how `unique` works in cuNumeric https://github.com/nv-legate/cunumeric/blob/branch-23.05/cunumeric/deferred.py#L3426.

This would also change the interface, such that the output store is created inside the function call, and returned as the result (rather than having the user pre-allocate a store of appropriate size).

Alternatively, if possible you could inspect the file's metadata, and preallocate an appropriate store inside this function, before doing the actual read. You wouldn't need to use unbounded stores in that case.

I believe for safety you also want to make it clear to legate that this must be launched as a singleton task:

```
    task = user_context.create_auto_task(TaskOpCode.READ)
    task.add_scalar_arg(path, types.string)
    task.add_output(output)
    # the whole output Store must be accessible by all point tasks
    # Stores accessed with write permissions cannot be shared
    # therefore the launch can only contain one point task.
    task.add_broadcast(output)  
    task.execute()
```

madsbk: > I think this task would be more appropriate as an unbounded-output task, similar to how `unique` works in cuNumeric https://github.com/nv-legate/cunumeric/blob/branch-23.05/cunumeric/deferred.py#L3426.
> 
> This would also change the interface, such that the output store is created inside the function call, and returned as the result (rather than having the user pre-allocate a store of appropriate size).

In this case, I would still have to know the total size of the file beforehand in order for each task to determine their file offset, right? 

> Alternatively, if possible you could inspect the file's metadata, and preallocate an appropriate store inside this function, before doing the actual read. You wouldn't need to use unbounded stores in that case.

Agree, the user API could support a default that infer the buffer type. However, the plan is to incorporate this into the existing KvikIO API so that something like the following will just work, no matter if `buf` is a NumPy, CuPy, or cuNumeric array:
```python
f = kvikio.CuFile("test-file", "r")
f.read(buf)
```  

> I believe for safety you also want to make it clear to legate that this must be launched as a singleton task:

I don't think this should be a singleton task, I want Legate to run multiple tasks in parallel. I am hoping that by calculating the offset based on the input store, each task will read its non-overlapping part of the file?
```c++
    auto shape = store.shape<1>();
    auto acc = store.read_accessor<char, 1>();
    size_t strides[1];
    const char* data = acc.ptr(shape, strides);
    size_t itemsize  = sizeof_legate_type_code(store.code());
    assert(strides[0] == itemsize);  // Must be contiguous
    size_t nbytes = shape.volume() * itemsize;  
    size_t offset = shape.lo.x * itemsize;  // Offset in bytes
```

legate/cpp/legate_kvikio.cpp
```diff
@@ -0,0 +1,120 @@
...
+class WriteTask : public Task<WriteTask, TaskOpCode::OP_WRITE> {
+ public:
+  static void cpu_variant(legate::TaskContext& context)
+  {
+    std::string path     = context.scalars()[0].value<std::string>();
+    legate::Store& store = context.inputs()[0];
+    auto shape           = store.shape<1>();
+    auto acc             = store.read_accessor<char, 1>();
+    size_t strides[1];
+    const char* data = acc.ptr(shape, strides);
```
madsbk, line 54: @manopapad, can I assume that `data` is contiguous? 
If not, can I create two task variants - one to handle contiguous stores and one for non-contiguous stores?

legate/cpp/legate_kvikio.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <kvikio/file_handle.hpp>
+
+namespace legate_kvikio {
+
+struct elem_size_fn {
+  template <legate::LegateTypeCode DTYPE>
+  size_t operator()()
+  {
+    using dtype = typename legate::LegateTypeOf<DTYPE>::type;
+    return sizeof(dtype);
```
wence-, line 36:
```suggestion
    return sizeof(legate::legate_type_of<DTYPE>);
```
?

legate/legate_kvikio/_version.py
wence-, line 1: FWIW, the rest of rapids is in the process of/has moved away from using versioneer to provide versions.

madsbk: We should do that here as well but let's do it once for the whole repos.

legate/legate_kvikio/core.py
manopapad: > In this case, I would still have to know the total size of the file beforehand in order for each task to determine their file offset, right?

Not necessarily. Each point task could query the size of the file independently and start reading at `size * point_task_id / num_point_tasks`. Each task would need to return to the caller how many elements it actually read, so the returned buffers can be virtually "stitched together". See the code in `unique` for how that's done (the term "weights" refers to how many elements each point task read).

> Agree, the user API could support a default that infer the buffer type. However, the plan is to incorporate this into the existing KvikIO API

Then I assume there's calls so that the user code can query the size of the file, so it can pre-allocate an array of the right size?

> I want Legate to run multiple tasks in parallel. I am hoping that by calculating the offset based on the input store, each task will read its non-overlapping part of the file?

Yes, that should work.

legate/cpp/legate_kvikio.cpp
manopapad: > can I assume that data is contiguous?

You will certainly want to check this, since the API allows the user to pass in an arbitrary array (if the API were to create its own output array, then you could guarantee it internally). You could run into trouble if the user did something like:

```
x = np.zeros((4,5))
read_into(x[:,3])
```

where the user passes in a column, which is technically 1d, but that slice is not contiguous. Even in this case, you can force Legion to give you a "compacted" view of the slice, by declaring that the task needs an "exact" instance, similar to how it's done here https://github.com/nv-legate/cunumeric/blob/branch-23.05/src/cunumeric/mapper.cc#L111.

> If not, can I create two task variants - one to handle contiguous stores and one for non-contiguous stores?

Currently Legate's variant registration system is specialized for making 3 specific variants (CPU, OpenMP, GPU), so we'd  need some work to extend it beyond that. The simpler alternative, if you want to support non-contiguous stores, would be to take cases inside the task body as written.

legate/legate_kvikio/core.py
madsbk: Thanks for the clarification @manopapad

PR #184, opened 2023-03-17
posix_io: fix error message and allow `nbytes == 0` on write

madsbk:
The posix-write interpreted a return value of `0` as a `EOF` error, which is isn't.


jakirkham (approved):

madsbk:
thanks @jakirkham

PR #185, opened 2023-03-27
Implement `build.sh`

madsbk:
Implement the `build.sh` that other RAPIDS projects use.  

### Need Help
Running `./build.sh libkvikio kvikio legate` now works and cleaning the environment also works  `./build.sh clean`.

However, after a clean, installing `legate` doesn't work.
```
$ ./build.sh clean legate
CMake Error at /home/mkristensen/apps/miniconda3/envs/legate-0227/lib/cmake/legate_core/legate_core-config.cmake:332 (add_library):
  add_library cannot create ALIAS target "legate::legate_kvikio_python"
  because another target with the same name already exists.
Call Stack (most recent call first):
  CMakeLists.txt:36 (legate_default_python_install)
```
If I delete the installed cmake files from the conda environment, it works again:
```
$ rm -R /home/mkristensen/apps/miniconda3/envs/legate-0227/lib/cmake/legate_kvikio
$ ./build.sh clean legate
```
Anyone knows what is going on?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 *.*~
 build
 python/_skbuild
+legate/_skbuild
+legate/legate_kvikio/install_info.py
+legate/record.txt
+python/record.txt
 *.so
 dask-worker-space
 __pycache__
--- a/build.sh
+++ b/build.sh
@@ -18,11 +18,12 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libkvikio kvikio -v -g -n -s --ptds -h"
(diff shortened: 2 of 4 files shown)


robertmaynard:
You are adding the `legate::legate_kvikio_python` target on line 32 and the `legate_add_cpp_subdirectory` is also trying to add the target which fails. 

This occurs since the `legate_add_cpp_subdirectory` only checks for an existing `legate::legate_kvikio_python` target when invoked from scikit-build. To work around this you will need to do something like:

```cmake

find_package(legate_kvikio QUIET)
find_package(legate_core REQUIRED)

if(NOT TARGET legate::legate_kvikio_python)
  legate_add_cpp_subdirectory(cpp TARGET legate_kvikio EXPORT legate_kvikio-export)
  legate_default_python_install(legate_kvikio EXPORT legate_kvikio-export)
endif()
```

madsbk:
Thanks @robertmaynard, it works now. Can I ask you to review this PR?

robertmaynard (approved):

PR #188, opened 2023-03-28
Remove underscore in build string.

bdice:
This PR removes the underscore after "cuda" to align with other RAPIDS packages.


madsbk (approved):
Thanks @bdice

PR #189, opened 2023-03-29
Fallback to use the CUDA primary context 

madsbk:
Fixes https://github.com/rapidsai/cudf/issues/13019


cpp/include/kvikio/utils.hpp
```diff
@@ -51,37 +51,62 @@ inline constexpr std::size_t page_size = 4096;
...
+  CUdeviceptr dev_ptr = convert_void2deviceptr(devPtr);
+
+  // First we try to get the current context and making sure it can access the device pointer.
+  {
+    CUcontext ctx;
+    CUDA_DRIVER_TRY(cudaAPI::instance().CtxGetCurrent(&ctx));
+    if (ctx != nullptr) {
+      CUdeviceptr current_ctx_dev_ptr{};
+      CUDA_DRIVER_TRY(cudaAPI::instance().PointerGetAttribute(
+        &current_ctx_dev_ptr, CU_POINTER_ATTRIBUTE_DEVICE_POINTER, dev_ptr));
```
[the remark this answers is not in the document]
madsbk: Good point, fixed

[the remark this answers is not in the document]
wence-: Ah so we're getting a RMM allocation from a pool that's been created on the main thread on the worker, but then the read_json call might be happening on another thread which hasn't had a chance to call cuCtxCreate?

What is the consequence if the pointer is associated with a context that doesn't belong to the current thread? Or is the context enough to ensure that the driver knows what to do?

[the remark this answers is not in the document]
madsbk: > Ah so we're getting a RMM allocation from a pool that's been created on the main thread on the worker, but then the read_json call might be happening on another thread which hasn't had a chance to call cuCtxCreate?

Yes, exactly.

> Or is the context enough to ensure that the driver knows what to do?

Yes, the context is enough

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,84 @@ inline constexpr std::size_t page_size = 4096;
...
-  return ret;
+/**
+ * @brief Return a CUDA that can be used with the given device pointer
+ *
+ * @param devPtr Device pointer to query
+ * @return Usable CUDA context
+ */
+[[nodiscard]] inline CUcontext get_context_from_pointer(const void* devPtr)
+{
+  static std::map<int, CudaPrimaryContext> _primary_contexts;
```
wence-, line 105: So this keeps any primary contexts we create alive until the program exits.

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,84 @@ inline constexpr std::size_t page_size = 4096;
...
+  // See <https://docs.nvidia.com/cuda/cuda-c-programming-guide/#pointer-attributes>
+  CUcontext ctx;
+  const CUresult err =
+    cudaAPI::instance().PointerGetAttribute(&ctx, CU_POINTER_ATTRIBUTE_CONTEXT, dev_ptr);
+  if (err != CUDA_ERROR_INVALID_VALUE) {
+    // If this isn't the case, we grab the current context
+    CUDA_DRIVER_TRY(cudaAPI::instance().CtxGetCurrent(&ctx));
+  } else {
+    CUDA_DRIVER_TRY(err);  // Check for errors from previous asynchronous launches
+  }
```
wence-, line 119: I think this logic is wrong? Let's try and work through it:

The options for the allocation are:
1. stream ordered: `CU_POINTER_ATTRIBUTE_CONTEXT` will return `CUDA_SUCCESS` and put `ctx = nullptr`
2. not stream ordered: `CU_POINTER_ATTRIBUTE_CONTEXT` will return `CUDA_SUCCESS` and put `ctx = some_context`
3. not a UVM allocation: `CU_POINTER_ATTRIBUTE_CONTEXT` will return `CUDA_ERROR_INVALID_VALUE` 
4. Some other thing: some other cuda error

So if we hit case 3 or 4, we want to error, if we hit case 2, we want to return the context that the query returned, if we hit case 1 we need to do some more work: I think we can just use `CU_POINTER_ATTRIBUTE_DEVICE_POINTER` query and then do something based on that (either return a non-null current context or else build a primary context and return that).

So perhaps:
```c++
CUcontext ctx;
// either the pointer is valid (CUDA_SUCCESS), or it isn't (some other error)
CUDA_DRIVER_TRY(cudaAPI::instance().PointerGetAttribute(&ctx, CU_POINTER_ATTRIBUTE_CONTEXT, dev_ptr));
if (ctx != nullptr) {
    // not a stream-ordered allocation, this context is fine.
    return ctx;
}
// stream-ordered allocation, let's see if the current device context exists and we can use it
CUDA_DRIVER_TRY(cudaAPI::instance().CtxGetCurrent(&ctx));
if (ctx != nullptr) {
    CUdeviceptr target{};
    auto const err = cudaAPI::instance().PointerGetAttribute(&target, CU_POINTER_ATTRIBUTE_DEVICE_POINTER, dev_ptr);
    if (err == CUDA_SUCCESS && target == dev_ptr) return ctx;
}
// Either no context, or the current context can't access the device pointer?
int ord = get_device_ordinal_from_pointer(devPtr);
_primary_contexts.try_emplate(ord, ord);
return _primary_contexts.at(ord).ctx;
```

WDYT?

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,84 @@ inline constexpr std::size_t page_size = 4096;
...
+
+  // If we found a context and it can access `devPtr`, we return it.
+  if (ctx != nullptr) {
+    CUdeviceptr current_ctx_dev_ptr{};
+    CUDA_DRIVER_TRY(cudaAPI::instance().PointerGetAttribute(
+      &current_ctx_dev_ptr, CU_POINTER_ATTRIBUTE_DEVICE_POINTER, dev_ptr));
+    if (current_ctx_dev_ptr != dev_ptr) { return ctx; }
```
wence-, line 126: Should check for equality?

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,84 @@ inline constexpr std::size_t page_size = 4096;
...
- * @param check_owning_devPtr If not NULL, a device memory pointer that must have
- * been allocated by, mapped by, or registered with the current context. If this
- * isn't the case, a CUfileException is thrown.
- *
- * @return The current cuda context
+ * @param ptr Device pointer to query
+ * @return The device ordinal
  */
-[[nodiscard]] inline CUcontext get_current_context(const void* check_owning_devPtr = nullptr)
+[[nodiscard]] inline int get_device_ordinal_from_pointer(const void* devPtr)
```
wence-, line 61: Does it make sense to accept a `CUdeviceptr` instead of a `void * const`? The only _current_ caller already has a `CUdeviceptr` hanging around. I guess it's a wash.

wence- (commented):
Can we work through the context-getting logic, I am not sure it is right?

cpp/include/kvikio/utils.hpp
madsbk: Fixed

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+  CUdeviceptr dev_ptr = convert_void2deviceptr(devPtr);
+
+  // First we check if a context has been associated with `devPtr`.
+  // Notice, this is not the case for stream ordered device memory allocations.
+  // See <https://docs.nvidia.com/cuda/cuda-c-programming-guide/#pointer-attributes>
+  {
+    CUcontext ctx;
+    const CUresult err =
+      cudaAPI::instance().PointerGetAttribute(&ctx, CU_POINTER_ATTRIBUTE_CONTEXT, dev_ptr);
+    if (err != CUDA_ERROR_INVALID_VALUE) {
```
wence-, line 114: Why do we fall through if the result is `CUDA_ERROR_INVALID_VALUE`? Ah, I guess if we don't have UVM, this might still be a device pointer?

wence- (commented):
Still one quibble about the case where we have a context and have queried the device pointer attribute.

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+  // If this isn't the case, we check the current context. If it exist and can access `devPtr`,
+  // we return it.
+  {
+    CUcontext ctx;
+    CUDA_DRIVER_TRY(cudaAPI::instance().CtxGetCurrent(&ctx));
+    if (ctx != nullptr) {
+      CUdeviceptr current_ctx_dev_ptr{};
+      CUDA_DRIVER_TRY(cudaAPI::instance().PointerGetAttribute(
+        &current_ctx_dev_ptr, CU_POINTER_ATTRIBUTE_DEVICE_POINTER, dev_ptr));
+      if (current_ctx_dev_ptr != dev_ptr) { return ctx; }
```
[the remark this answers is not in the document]
madsbk: If `dev_ptr != current_ctx_dev_ptr`, we cannot access `dev_ptr` using the current context right?

cpp/include/kvikio/utils.hpp
madsbk: We don't want to crash when getting `CUDA_ERROR_INVALID_VALUE` so we cannot use `CUDA_DRIVER_TRY()`. Instead, we want to get the current context (if any).

madsbk:
Marked the PR as a draft again. I will write some tests tomorrow before we merge

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+  // First we check if a context has been associated with `devPtr`.
+  // Notice, this is not the case for stream ordered device memory allocations.
+  // See <https://docs.nvidia.com/cuda/cuda-c-programming-guide/#pointer-attributes>
+  {
+    CUcontext ctx;
+    const CUresult err =
+      cudaAPI::instance().PointerGetAttribute(&ctx, CU_POINTER_ATTRIBUTE_CONTEXT, dev_ptr);
+    if (err != CUDA_ERROR_INVALID_VALUE) {
+      CUDA_DRIVER_TRY(err);  // Check for other errors
+      return ctx;
```
wence-, line 116: Need to check that `ctx != nullptr`

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+  // First we check if a context has been associated with `devPtr`.
+  // Notice, this is not the case for stream ordered device memory allocations.
+  // See <https://docs.nvidia.com/cuda/cuda-c-programming-guide/#pointer-attributes>
+  {
+    CUcontext ctx;
+    const CUresult err =
+      cudaAPI::instance().PointerGetAttribute(&ctx, CU_POINTER_ATTRIBUTE_CONTEXT, dev_ptr);
+    if (err != CUDA_ERROR_INVALID_VALUE && ctx != nullptr) {
+      CUDA_DRIVER_TRY(err);  // Check for other errors
+      return ctx;
```
wence-, line 116:
```suggestion
    if (err != CUDA_ERROR_INVALID_VALUE) {
      CUDA_DRIVER_TRY(err);  // Check for other errors
      if (ctx != nullptr) {
          return ctx;
      }
```

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+[[nodiscard]] inline CUcontext get_context_from_pointer(const void* devPtr)
+{
+  static std::map<int, CudaPrimaryContext> _primary_contexts;
+  CUdeviceptr dev_ptr = convert_void2deviceptr(devPtr);
+
+  // First we check if a context has been associated with `devPtr`.
+  // Notice, this is not the case for stream ordered device memory allocations.
+  // See <https://docs.nvidia.com/cuda/cuda-c-programming-guide/#pointer-attributes>
+  {
+    CUcontext ctx;
```
wence-, line 111:
```suggestion
    CUcontext ctx = nullptr;
```

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+      cudaAPI::instance().PointerGetAttribute(&ctx, CU_POINTER_ATTRIBUTE_CONTEXT, dev_ptr);
+    if (err != CUDA_ERROR_INVALID_VALUE && ctx != nullptr) {
+      CUDA_DRIVER_TRY(err);  // Check for other errors
+      return ctx;
+    }
+  }
+  // If this isn't the case, we check the current context. If it exists and can access `devPtr`,
+  // we return it.
+  {
+    CUcontext ctx;
```
wence-, line 125:
```suggestion
    CUcontext ctx = nullptr;
```

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+    CUDA_DRIVER_TRY(cudaAPI::instance().CtxGetCurrent(&ctx));
+    if (ctx != nullptr) {
+      CUdeviceptr current_ctx_dev_ptr{};
+      CUDA_DRIVER_TRY(cudaAPI::instance().PointerGetAttribute(
```
wence-, line 129: We could have `CUDA_ERROR_INVALID_VALUE` here (the current context is not set up for allowing us to access this pointer). Is that a recoverable error?

wence-: Should certainly handle other errors as failure.

cpp/include/kvikio/utils.hpp
```diff
@@ -52,36 +53,87 @@ inline constexpr std::size_t page_size = 4096;
...
+        &current_ctx_dev_ptr, CU_POINTER_ATTRIBUTE_DEVICE_POINTER, dev_ptr));
+      if (current_ctx_dev_ptr != dev_ptr) { return ctx; }
+    }
+  }
+  // Finally, if we didn't find any usable context, we return the primary context of the
+  // device that owns `devPtr`. Notice, we use `_primary_contexts` to cache the primary
+  // context of each device.
+  int ordinal = get_device_ordinal_from_pointer(dev_ptr);
```
wence-, line 137: If `dev_ptr` was allocated with `cuMemHostAlloc` with `CU_MEMHOSTALLOC_WRITECOMBINED`, and is the "host address". It looks like a UVM address, but it is unmapped on the device, so cannot be used in a kernel, even with this primary context that we just made. In that case, the pointer that is usable on the device is the one obtained from querying `CUDA_POINTER_ATTRIBUTE_DEVICE_POINTER`.

wence-: After setting the primary context, I think if we go round and query `CUDA_POINTER_ATTRIBUTE_DEVICE_POINTER` _again_ and find that the result is different from the input, then we get an error.

wence-: Maybe we should try and clarify some more with the cuda team.

cpp/include/kvikio/utils.hpp
```diff
@@ -112,6 +81,136 @@ inline bool is_host_memory(const void* ptr)
...
+    } catch (const CUfileException& e) {
+      std::cerr << e.what() << std::endl;
+    }
+  }
+};
+
+/**
+ * @brief Given a device ordinal, return the primary context of the device.
+ *
+ * This function cache the primary contexts retrieved until program exit
```
wence-, line 128:
```suggestion
 * This function caches the primary contexts retrieved until program exit
```

cpp/include/kvikio/utils.hpp
```diff
@@ -112,6 +81,136 @@ inline bool is_host_memory(const void* ptr)
...
+  return {};
+}
+
+/**
+ * @brief Check if the current CUDA context can access the given device pointer
+ *
+ * @param dev_ptr Device pointer to query
+ * @return The boolean answer
+ */
+[[nodiscard]] inline bool can_current_context_access_pointer(CUdeviceptr dev_ptr)
```
wence-, line 162:
```suggestion
[[nodiscard]] inline bool current_context_can_access_pointer(CUdeviceptr dev_ptr)
```

cpp/include/kvikio/utils.hpp
```diff
@@ -112,6 +81,136 @@ inline bool is_host_memory(const void* ptr)
...
+  if (err != CUDA_ERROR_INVALID_VALUE) { CUDA_DRIVER_TRY(err); }
+  return false;
+}
+
+/**
+ * @brief Return a CUDA context that can be used with the given device pointer
+ *
+ * For robustness, we look for an usabale context in the following order:
+ *   1) If a context has been associated with `devPtr`, it is returned
+ *   2) If the current context exist and can access `devPtr`, it is returned.
```
wence-, line 177:
```suggestion
 *   2) If the current context exists and can access `devPtr`, it is returned.
```

cpp/include/kvikio/utils.hpp
```diff
@@ -112,6 +81,136 @@ inline bool is_host_memory(const void* ptr)
...
+    }
+  }
+
+  // If this isn't the case, we check the current context. If it exist and can access `devPtr`, we
+  // return the current context.
+  {
+    CUcontext ctx = nullptr;
+    CUDA_DRIVER_TRY(cudaAPI::instance().CtxGetCurrent(&ctx));
+    if (ctx != nullptr && can_current_context_access_pointer(dev_ptr)) {
+      std::cout << "get_context_from_pointer() - can_current_context_access_pointer" << std::endl;
```
wence-, line 202:
```suggestion
```

cpp/include/kvikio/utils.hpp
```diff
@@ -112,6 +81,136 @@ inline bool is_host_memory(const void* ptr)
...
+ */
+[[nodiscard]] inline CUcontext get_context_from_pointer(const void* devPtr)
+{
+  CUdeviceptr dev_ptr = convert_void2deviceptr(devPtr);
+
+  // First we check if a context has been associated with `devPtr`.
+  {
+    auto ctx = get_context_associated_pointer(dev_ptr);
+    if (ctx.has_value()) {
+      std::cout << "get_context_from_pointer() - context_associated" << std::endl;
```
wence-, line 191:
```suggestion
 ```

Debugging leftovers I guess?

File: cpp/include/kvikio/utils.hpp [hunk 19]
```diff
@@ -112,6 +81,136 @@ inline bool is_host_memory(const void* ptr)
...
+    CUcontext ctx = nullptr;
+    CUDA_DRIVER_TRY(cudaAPI::instance().CtxGetCurrent(&ctx));
+    if (ctx != nullptr && can_current_context_access_pointer(dev_ptr)) {
+      std::cout << "get_context_from_pointer() - can_current_context_access_pointer" << std::endl;
+      return ctx;
+    }
+  }
+
+  std::cout << "get_context_from_pointer() - get_primary_cuda_context" << std::endl;
+
```
wence-, line 208:
```suggestion
```

python/tests/test_basic_io.py
```diff
@@ -133,6 +134,69 @@ def test_read_write_slices(tmp_path, xp, nthreads, tasksize, start, end):
...
+
+    with cupy.cuda.using_allocator(cupy.cuda.malloc_async):
+        ary = cupy.arange(100)
+
+    with with_no_cuda_context():
+        with kvikio.CuFile(filename, "w") as f:
+            f.write(ary)
+
+    with kvikio.CuFile(filename, "w") as f:
+        f.write(ary)
```
wence-, line 197: Can we check we got what we were expecting in these files?

cpp/include/kvikio/utils.hpp
```diff
@@ -112,6 +81,136 @@ inline bool is_host_memory(const void* ptr)
...
+  {
+    auto ctx = get_context_associated_pointer(dev_ptr);
+    if (ctx.has_value()) {
+      std::cout << "get_context_from_pointer() - context_associated" << std::endl;
+      return ctx.value();
+    }
+  }
+
+  // If this isn't the case, we check the current context. If it exist and can access `devPtr`, we
+  // return the current context.
```
wence-, line 197: Maybe explain what you mean by "can access"?

wence- (commented):
Thanks, just some minor debugging leftovers to clean up I think

madsbk:
Thanks for the help @wence- !

wence- (approved):
Thanks @madsbk! How shall we test that this works for the initial reported cudf bug?

madsbk:
> How shall we test that this works for the initial reported cudf bug?

In `test_no_current_cuda_context()`, we test the underlying problem of the reported cudf+dask bug. 
But it might be too expensive in terms of CI resources to test the exact cudf code.
Since KvikIO is header-only, CI would have to build cudf.

wence-:
> > How shall we test that this works for the initial reported cudf bug?
> 
> In `test_no_current_cuda_context()`, we test the underlying problem of the reported cudf+dask bug. But it might be too expensive in terms of CI resources to test the exact cudf code. Since KvikIO is header-only, CI would have to build cudf every time.

I was going to just build locally and check?

madsbk:
> I was going to just build locally and check?

Ahh sure, I have done that continually :)

To built cudf using this branch, add the KvikIO source in the `build.sh` script like:
```diff
diff --git a/build.sh b/build.sh
index bee66d819b..b6bcf5ce75 100755
--- a/build.sh
+++ b/build.sh
@@ -288,6 +288,7 @@ if buildAll || hasArg libcudf; then
           -DDISABLE_DEPRECATION_WARNINGS=${BUILD_DISABLE_DEPRECATION_WARNINGS} \
           -DCUDF_USE_PER_THREAD_DEFAULT_STREAM=${BUILD_PER_THREAD_DEFAULT_STREAM} \
           -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
+          -DCPM_KvikIO_SOURCE=/home/mkristensen/repos/kvikio \
           ${EXTRA_CMAKE_ARGS}
 
     cd ${LIB_BUILD_DIR}
```

wence-:
Can confirm this fixes the reported cudf issue

PR #194, opened 2023-04-11
`"Programming Language :: Python :: 3.10"`

madsbk:


jakirkham:
Should we add `python_requires = ">=3.8"` as well ([ref]( https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement ), under `setup.py` tab)?

bdice:
@madsbk All project information should migrate from `setup.py` and `setup.cfg` to `pyproject.toml`. That is the direction that the Python community is headed, and it's what we're striving for in RAPIDS. @vyasr can probably note the small number of exceptions where we cannot yet use `pyproject.toml`.

bdice:
Reference PRs:
- https://github.com/rapidsai/rmm/pull/1151
- https://github.com/rapidsai/cudf/pull/12850
- https://github.com/rapidsai/raft/pull/1324
- https://github.com/rapidsai/cuml/pull/5251
- https://github.com/rapidsai/cugraph/pull/3317

vyasr:
The current state of the art is captured fairly well in the pyproject.toml files of any of the core RAPIDS repos e.g. cudf or raft. [This part of the scikit-build documentation lists which parameters need to stay in setup.py](https://scikit-build.readthedocs.io/en/latest/usage.html#setuptools-options). Everything else can go into pyproject.toml. The only addendum is, don't use MANIFEST.in files for now (see https://github.com/scikit-build/scikit-build/issues/886).

PR #197, opened 2023-04-17
Add API to get compatibility mode status in a FileHandle object

vuule:
Adding `is_compat_mode_on()` to FileHandle to make the run-time behavior more transparent for the users.


madsbk (approved):
Looks good, thanks @vuule

PR #203, opened 2023-04-25
Only look for libcufile.so.0

### LINKED ISSUE #200: KvikIO: Need to canonicalize dlopen'd library names
Documenting an offline discussion with @jakirkham in preparation of CUDA 12 bring-up on conda-forge.

Currently, there're some places where `libXXXXX.so` without any SONAME being dlopen'd, for example,
https://github.com/rapidsai/kvikio/blob/db0a3e71dccb17bbe7443a295c84aeeadb1359bc/cpp/include/kvikio/shim/cufile.hpp#L53
There will become problematic because the `libXXXXX.so` symlink is supposed to exist only in the `libXXXXX-dev` package by the stand practice of Linux distros, not in `libXXXXX` which only contains `libXXXXX.so.{major}`. The dlopen'd names need to be canonicalized.

Note: Text borrowed from Leo's issue ( https://github.com/rapidsai/cudf/issues/12708 ) and tweaked

wence-:
Rather than searching for a list of possible library names for libcufile, only look for the single ABI-compatible version of libcufile that we support. This avoids both ABI incompatibilities and loading stub libraries. It morally does what #127 attempted, but with the correct versioning suffix.

Closes #200.

### DIFF
--- a/cpp/include/kvikio/shim/cufile.hpp
+++ b/cpp/include/kvikio/shim/cufile.hpp
@@ -50,7 +50,17 @@ class cuFileAPI {
  private:
   cuFileAPI()
   {
-    void* lib = load_library("libcufile.so.0");
+    // CUDA versions before CUDA 11.7.1 did not ship libcufile.so.0, so this is
+    // a workaround that adds support for all prior versions of libcufile.
+    void* lib = load_library({"libcufile.so.0",
+                              "libcufile.so.1.3.0" /* 11.7.0 */,
+                              "libcufile.so.1.2.1" /* 11.6.2, 11.6.1 */,
+                              "libcufile.so.1.2.0" /* 11.6.0 */,
+                              "libcufile.so.1.1.1" /* 11.5.1 */,
+                              "libcufile.so.1.1.0" /* 11.5.0 */,
+                              "libcufile.so.1.0.2" /* 11.4.4, 11.4.3, 11.4.2 */,
+                              "libcufile.so.1.0.1" /* 11.4.1 */,
+                              "libcufile.so.1.0.0" /* 11.4.0 */});
     get_symbol(HandleRegister, lib, KVIKIO_STRINGIFY(cuFileHandleRegister));
     get_symbol(HandleDeregister, lib, KVIKIO_STRINGIFY(cuFileHandleDeregister));
     get_symbol(Read, lib, KVIKIO_STRINGIFY(cuFileRead));
--- a/cpp/include/kvikio/shim/cufile.hpp
+++ b/cpp/include/kvikio/shim/cufile.hpp
@@ -50,7 +50,7 @@ class cuFileAPI {
  private:
   cuFileAPI()
   {
-    void* lib = load_library({"libcufile.so.1", "libcufile.so.0", "libcufile.so"});
+    void* lib = load_library("libcufile.so.0");
     get_symbol(HandleRegister, lib, KVIKIO_STRINGIFY(cuFileHandleRegister));
     get_symbol(HandleDeregister, lib, KVIKIO_STRINGIFY(cuFileHandleDeregister));
     get_symbol(Read, lib, KVIKIO_STRINGIFY(cuFileRead));


bdice (approved):
This looks correct to me.

madsbk:
Is `libcufile.so.0` always there?
I seem to remember installations where `libcufile.so.1` or `libcufile.so` was the only one installed?
E.g. looking for `libcufile.so.1` exclusively did work when https://github.com/rapidsai/kvikio/pull/127 was merged.

madsbk:
CUDA v11.5 has:
```
mkristensen@mkristensen-workstation:~/apps/cuda-11.5.2/lib64$ ls -l libcufile*
lrwxrwxrwx 1 mkristensen mkristensen       23 Aug  2  2022 libcufile_rdma.so -> libcufile_rdma.so.1.1.1
-rw-r--r-- 1 mkristensen mkristensen    39128 Nov  4  2021 libcufile_rdma.so.1.1.1
-rw-r--r-- 1 mkristensen mkristensen    61778 Nov  4  2021 libcufile_rdma_static.a
lrwxrwxrwx 1 mkristensen mkristensen       18 Aug  2  2022 libcufile.so -> libcufile.so.1.1.1
-rw-r--r-- 1 mkristensen mkristensen  1293192 Nov  4  2021 libcufile.so.1.1.1
-rw-r--r-- 1 mkristensen mkristensen 62822740 Nov  4  2021 libcufile_static.a
```

wence-:
> CUDA v11.5 has:
> 
> ```
> mkristensen@mkristensen-workstation:~/apps/cuda-11.5.2/lib64$ ls -l libcufile*
> lrwxrwxrwx 1 mkristensen mkristensen       23 Aug  2  2022 libcufile_rdma.so -> libcufile_rdma.so.1.1.1
> -rw-r--r-- 1 mkristensen mkristensen    39128 Nov  4  2021 libcufile_rdma.so.1.1.1
> -rw-r--r-- 1 mkristensen mkristensen    61778 Nov  4  2021 libcufile_rdma_static.a
> lrwxrwxrwx 1 mkristensen mkristensen       18 Aug  2  2022 libcufile.so -> libcufile.so.1.1.1
> -rw-r--r-- 1 mkristensen mkristensen  1293192 Nov  4  2021 libcufile.so.1.1.1
> -rw-r--r-- 1 mkristensen mkristensen 62822740 Nov  4  2021 libcufile_static.a
> ```

Oh, ugh. Although, where do you see that? In the `nvidia/cuda:11.5.1-runtime-ubuntu20.04` docker image I see:
```
root@1e1fde448191:/# ls -l /usr/local/cuda/lib64/libcufile*
lrwxrwxrwx 1 root root      18 Dec 14 22:20 /usr/local/cuda/lib64/libcufile.so.0 -> libcufile.so.1.1.1
-rw-r--r-- 1 root root 1293192 Nov  3  2021 /usr/local/cuda/lib64/libcufile.so.1.1.1
lrwxrwxrwx 1 root root      23 Dec 14 22:20 /usr/local/cuda/lib64/libcufile_rdma.so.1 -> libcufile_rdma.so.1.1.1
-rw-r--r-- 1 root root   39128 Nov  3  2021 /usr/local/cuda/lib64/libcufile_rdma.so.1.1.1
```

madsbk:
I think it was installed using the Ubuntu debian package: https://developer.download.nvidia.com/compute/cuda/11.5.2/local_installers/cuda-repo-ubuntu2004-11-5-local_11.5.2-495.29.05-1_amd64.deb

bdice:
I can confirm the Ubuntu 20.04 Debian package linked above only contains `libcufile.so.1.1.1` and `libcufile.so` (a symlink to `libcufile.so.1.1.1`).

wence-:
I think then, given that _actually_ the ABI has never changed we could just use `libcufile.so.0` and then encode the particular exact library versions as well that we know work.

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -50,7 +50,7 @@ class cuFileAPI {
  private:
   cuFileAPI()
   {
-    void* lib = load_library({"libcufile.so.1", "libcufile.so.0", "libcufile.so"});
+    void* lib = load_library("libcufile.so.0");
```
jakirkham, line 53: Suggestion from offline discussion is to do something like this, but ideally only under CUDA 11. Seems this shouldn't be an issue any more in CUDA 12

```suggestion
    void* lib = load_library({"libcufile.so.0", "libcufile.so"});
```

wence-: Hard no.

bdice: Discussed via Slack. This is an alternative that works for CUDA versions before 11.7.1 when this was fixed. (Also this passes clang-format.)

```suggestion
    // CUDA versions before CUDA 11.7.1 did not ship libcufile.so.0, so this is
    // a workaround that adds support for all prior versions of libcufile.
    void* lib = load_library({"libcufile.so.0",
                              "libcufile.so.1.3.0" /* 11.7.0 */,
                              "libcufile.so.1.2.1" /* 11.6.2, 11.6.1 */,
                              "libcufile.so.1.2.0" /* 11.6.0 */,
                              "libcufile.so.1.1.1" /* 11.5.1 */,
                              "libcufile.so.1.1.0" /* 11.5.0 */,
                              "libcufile.so.1.0.2" /* 11.4.4, 11.4.3, 11.4.2 */,
                              "libcufile.so.1.0.1" /* 11.4.1 */,
                              "libcufile.so.1.0.0" /* 11.4.0 */});
```

wence-: Thanks @bdice!

wence-:
OK, I think this is good for another look @madsbk.

madsbk (approved):
LGTM, thanks all!

jakirkham:
Thanks all! 🙏

PR #210, opened 2023-05-01
Enable sccache hits for local builds

AyodeAwe:
This change passes through the value of `SCCACHE_S3_NO_CREDENTIALS` to our `conda` builds, enabling devs to utilize the `sccache` cache that's populated by CI when they are reproducing build issues locally as per [these](https://docs.rapids.ai/resources/reproducing-ci/) instructions.


madsbk (approved):
Thanks @AyodeAwe

PR #213, opened 2023-05-03
Update cupy dependency

vyasr:
Switch to requiring cupy 12


jakirkham:
Do we want to update the Conda recipe as well?

https://github.com/rapidsai/kvikio/blob/2e74e8ff91027136fde452c4cdcde15fb284a905/conda/recipes/kvikio/meta.yaml#L46

PR #222, opened 2023-05-23
Legate HDF5 using kerchunk

madsbk:
Support reading of HDF5 files using kerchunk.

#### Limitations

- Only read is supported. 
- Can only read whole arrays -- no partial access.
- No compression -- only  HDF5 files written without compression is supported.
- Groups not supported -- only access to individual Zarr arrays are implemented.

#### Example

```python

import h5py
import numpy as np

import legate_kvikio.kerchunk
import legate_kvikio.zarr


def hdf5_io(filename):
    a = np.arange(10000).reshape((100, 100))

    # Write array using h5py
    with h5py.File(filename, "w") as f:
        f.create_dataset("mydataset", chunks=(10, 10), data=a)

    # Read hdf5 file using legate+kerchunk
    b = legate_kvikio.kerchunk.hdf5_read(filename, dataset_name="mydataset")

    # They should be equal
    assert (a == b).all()


if __name__ == "__main__":
    hdf5_io("/tmp/legate-kvikio-io.hdf5")
```


wence- (commented):
Some minor suggestions and a few queries.

legate/benchmarks/hdf5_read.py
```diff
@@ -0,0 +1,188 @@
...
+
+    yield f
+
+
+API = {
+    "dask-h5py": dask_h5py,
+    "legate": run_legate,
+}
+
+OP = {"add": lambda xp, a, b: a + b, "matmul": lambda xp, a, b: xp.matmul(a, b)}
```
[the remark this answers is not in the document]
madsbk: Fixed

legate/legate_kvikio/tile.py
```diff
@@ -102,3 +102,39 @@ def read_tiles(
...
+    task = context.create_manual_task(
+        TaskOpCode.TILE_READ_BY_OFFSETS,
+        launch_domain=Rect(launch_shape),
+    )
+
+    task.add_output(store_partition)
+    task.add_scalar_arg(filepath, types.string)
+    task.add_scalar_arg(offsets, (types.uint64,))
+    task.add_scalar_arg(sizes, (types.uint64,))
+    task.add_scalar_arg(tile_shape, (types.uint64,))
```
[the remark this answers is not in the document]
madsbk: Yes, sorry forgot to add a doc string for the function. Fixed.

legate/cpp/tile_io.cpp
```diff
@@ -105,6 +106,71 @@ struct tile_read_write_fn {
...
+    std::array<size_t, DIM> strides{};
+    if (nbytes != sizes[flatten_task_index]) {
+      throw std::runtime_error("sizes doesn't match the size of the output store");
+    }
+
+    // We know that the accessor is contiguous because we set `policy.exact = true`
+    // in `Mapper::store_mappings()`.
+    kvikio::FileHandle f(path, "r");
+    auto* data = store.write_accessor<DTYPE, DIM>().ptr(shape, strides.data());
+    f.pread(data, nbytes, offsets[flatten_task_index]).get();
```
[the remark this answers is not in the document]
madsbk: Yes

legate/legate_kvikio/kerchunk.py
```diff
@@ -0,0 +1,80 @@
...
+    refs = annotations["refs"]
+    offsets = []
+    sizes = []
+    for chunk_coord in itertools.product(
+        *(range(math.ceil(s / c)) for s, c in zip(zarr_ary.shape, zarr_ary.chunks))
+    ):
+        key = zarr_ary._chunk_key(chunk_coord)
+        _, offset, nbytes = refs[key]
+        offsets.append(offset)
+        sizes.append(nbytes)
```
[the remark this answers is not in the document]
madsbk: Each chunk in the hdf5 file is contiguous.  We use `SingleHdf5ToZarr()` for the conversion, which should raise an exception if it cannot find any chunks.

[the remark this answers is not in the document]
madsbk: Notice, if `SingleHdf5ToZarr` isn't able to convert the hdf5 file for any reason, we accept the failure.

legate/cpp/tile_io.cpp
```diff
@@ -105,6 +106,71 @@ struct tile_read_write_fn {
...
+                 const legate::DomainPoint& point_dp)
+{
+  legate::Point<DIM> lo      = lo_dp;
+  legate::Point<DIM> hi      = hi_dp;
+  legate::Point<DIM> point   = point_dp;
+  legate::Point<DIM> extents = hi - lo + legate::Point<DIM>::ONES();
+  size_t idx                 = 0;
+  for (int32_t dim = 0; dim < DIM; ++dim) {
+    idx = idx * extents[dim] + point[dim] - lo[dim];
+  }
```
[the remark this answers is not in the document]
madsbk: Done. 
`point_dp` must be within `lo` and `hi`. Since they come from `context.get_task_index()` and `context.get_launch_domain()` directly , I don't we need a check here.

madsbk:
Thanks @wence- , I think I have addressed all of your comments.

legate/legate_kvikio/tile.py
```diff
@@ -102,3 +102,66 @@ def read_tiles(
...
     return _tile_read_write(TaskOpCode.TILE_READ, ary, dirpath, tile_shape, tile_start)
+
+
+def read_tiles_by_offsets(
+    ary: cunumeric.ndarray,
+    filepath: Iterable[pathlib.Path | str],
+    offsets: Tuple[int],
+    sizes: Tuple[int],
+    tile_shape: Tuple[int],
+) -> None:
```
wence-, line 113: FWIW, even though this modifies in place, I think for API uniformity one could return an `ndarray` (just returning `ary` at the end of the function). WDYT?

legate/legate_kvikio/tile.py
```diff
@@ -102,3 +102,66 @@ def read_tiles(
...
+        Filepath to the file.
+    offsets
+        The offset of each tile in the file (in bytes).
+    sizes
+        The size of each tile in the file (in bytes).
+    tile_shape
+        The shape of each tile.
+    """
+
+    if any(d % c != 0 for d, c in zip(ary.shape, tile_shape)):
```
wence-, line 134: Does the number of modes of the `tile_shape` have to match the number of modes of the array? If so maybe add

```
if len(ary.shape) != len(tile_shape):
    raise ValueError("Tile shape and array shape must have  same number of axes")
```

legate/cpp/tile_io.cpp
```diff
@@ -105,6 +106,71 @@ struct tile_read_write_fn {
...
+    size_t nbytes = shape_volume * sizeof(DTYPE);
+    std::array<size_t, DIM> strides{};
+    if (nbytes != sizes[flatten_task_index]) {
+      throw std::runtime_error("sizes doesn't match the size of the output store");
+    }
+
+    // We know that the accessor is contiguous because we set `policy.exact = true`
+    // in `Mapper::store_mappings()`.
+    kvikio::FileHandle f(path, "r");
+    auto* data = store.write_accessor<DTYPE, DIM>().ptr(shape, strides.data());
```
wence-, line 169: `strides` is default initialised to all zeros on line 161 and then never accessed after this line. Do you need it at all? That is, would
```suggestion
    auto* data = store.write_accessor<DTYPE, DIM>().ptr(shape);
```
work here?

legate/cpp/tile_io.cpp
```diff
@@ -105,6 +106,71 @@ struct tile_read_write_fn {
...
+    const std::string path                  = context.scalars().at(0).value<std::string>();
+    legate::Span<const uint64_t> offsets    = context.scalars().at(1).values<uint64_t>();
+    legate::Span<const uint64_t> sizes      = context.scalars().at(2).values<uint64_t>();
+    legate::Span<const uint64_t> tile_shape = context.scalars().at(3).values<uint64_t>();
+
+    // Flatten task index
+    uint32_t flatten_task_index = 0;
+    if (!context.is_single_task()) {
+      flatten_task_index = linearize<DIM>(launch_domain.lo(), launch_domain.hi(), task_index);
+    }
```
wence-, line 155: Why do you need to special-case the single task case here? Or is this a micro-optimisation?

wence- (commented):
Thanks @madsbk , one required correctness-checking change and I've now had a chance to go through the C++ task description and have one more query there.

legate/legate_kvikio/tile.py
wence-: OK, thanks.

legate/legate_kvikio/tile.py
```diff
@@ -102,3 +102,66 @@ def read_tiles(
...
+    ary
+       The cuNumeric array to read into.
+    filepath
+        Filepath to the file.
+    offsets
+        The offset of each tile in the file (in bytes).
+    sizes
+        The size of each tile in the file (in bytes).
+    tile_shape
+        The shape of each tile.
```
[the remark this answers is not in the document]
wence-: Ah OK, so if your hdf5 on-disk data is whatever-compressed then you need to know the size independent of the tile-shape to read the right bit. I can buy that for leaving things in.

legate/cpp/tile_io.cpp
madsbk: Good catch, removed.

legate/cpp/tile_io.cpp
madsbk: Right, it isn't needed anymore. Removed

legate/legate_kvikio/tile.py
madsbk: Done

madsbk:
Thanks @wence-!

legate/legate_kvikio/tile.py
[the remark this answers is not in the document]
madsbk: Exactly, but in order to support decompression, we would need more changes to the task signature anyways so leaving `sizes` out for now.

legate/legate_kvikio/kerchunk.py
```diff
@@ -0,0 +1,83 @@
...
+    """Read an HDF5 array from disk using Kerchunk and KvikIO
+
+    We use Kerchunk's `SingleHdf5ToZarr` to find the data chunks embedded
+    in the hdf5 file. If it fails for any reason, this function fails as well.
+
+    Notes
+    -----
+    The returned array is padded to make its shape divisible by the shape of
+    the Zarr chunks on disk (if not already). This means that the returned
+    Legate store can be larger than the returned cuNumeric array.
```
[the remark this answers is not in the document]
wence-: Sounds good!

legate/cpp/tile_io.cpp
wence-: Did you push that change?

wence- (approved):
I think there was one "fixed" that you didn't push, but looks good, thanks

legate/cpp/tile_io.cpp
madsbk: Now, I have :)

madsbk:
Thanks @wence- !

Issue #175, opened 2023-02-14
KvikIO: CUDA 12 Conda Packages


jakirkham:
Pushing for CUDA 12 Conda package builds of libraries for RAPIDS 23.04. This issue is for tracking that work.

Requirements (please recheck):

```[tasklist]
### Developer tasks
- [x] I've confirmed the list below is correct (or updated if not)
```

```[tasklist]
### Toolkit dependencies from Conda
- [ ] https://github.com/conda-forge/staged-recipes/pull/21723
- [ ] https://github.com/conda-forge/staged-recipes/pull/21350
- [ ] https://github.com/conda-forge/staged-recipes/pull/21908
```

```[tasklist]
### Other optional dependencies from Conda using CUDA
- [ ] https://github.com/conda-forge/staged-recipes/pull/23034
- [ ] https://github.com/conda-forge/pynvml-feedstock/pull/11
```

Note: `nvcomp` is only need for compression (not IO). Also pynvml is only needed in the benchmarks (not for library usage).

madsbk:
Thanks @jakirkham. For building, both `libcufile` and `nvcomp` are required. At runtime, KvikIO will dlopen `libcufile` and if it isn't found, it will use its fallback compatibility mode. If `nvcomp` isn't available at runtime, the IO still works.

PR #224, opened 2023-05-25
KvikIO: Build CUDA 12 packages

bdice:
## Description
This PR builds `libkvikio` and `kvikio` conda packages using CUDA 12. Resolves https://github.com/rapidsai/kvikio/issues/175

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

<hr>

Fixes https://github.com/rapidsai/kvikio/issues/175


bdice:
I'm [seeing failures here](https://github.com/rapidsai/kvikio/actions/runs/5084252440/jobs/9136338681?pr=224#step:7:501) that appear to be unrelated to my changes. I can't compile `branch-23.08` locally [using the RAPIDS CI images](https://docs.rapids.ai/resources/reproducing-ci/). Testing locally with `branch-23.06`.

<details><summary>Details</summary>
<p>

```
CMake Error at /opt/conda/conda-bld/_build_env/share/cmake-3.26/Modules/CMakeTestCXXCompiler.cmake:60 (message):
  The C++ compiler

    "/opt/conda/conda-bld/_build_env/bin/aarch64-conda-linux-gnu-c++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /opt/conda/conda-bld/work/cpp/build/CMakeFiles/CMakeScratch/TryCompile-HnL0zR
    
    Run Build Command(s):/opt/conda/conda-bld/_build_env/bin/ninja -v cmTC_969ad && [1/2] sccache /opt/conda/conda-bld/_build_env/bin/aarch64-conda-linux-gnu-c++   -fvisibility-inlines-hidden -fmessage-length=0 -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O3 -pipe -isystem /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/include -fdebug-prefix-map=/opt/conda/conda-bld/work=/usr/local/src/conda/libkvikio-split- -fdebug-prefix-map=/opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho=/usr/local/src/conda-prefix -isystem /usr/local/cuda/include -o CMakeFiles/cmTC_969ad.dir/testCXXCompiler.cxx.o -c /opt/conda/conda-bld/work/cpp/build/CMakeFiles/CMakeScratch/TryCompile-HnL0zR/testCXXCompiler.cxx
    FAILED: CMakeFiles/cmTC_969ad.dir/testCXXCompiler.cxx.o 
    sccache /opt/conda/conda-bld/_build_env/bin/aarch64-conda-linux-gnu-c++   -fvisibility-inlines-hidden -fmessage-length=0 -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O3 -pipe -isystem /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/include -fdebug-prefix-map=/opt/conda/conda-bld/work=/usr/local/src/conda/libkvikio-split- -fdebug-prefix-map=/opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho=/usr/local/src/conda-prefix -isystem /usr/local/cuda/include -o CMakeFiles/cmTC_969ad.dir/testCXXCompiler.cxx.o -c /opt/conda/conda-bld/work/cpp/build/CMakeFiles/CMakeScratch/TryCompile-HnL0zR/testCXXCompiler.cxx
    sccache: error: Timed out waiting for server startup. Maybe the remote service is unreachable?
    Run with SCCACHE_LOG=debug SCCACHE_NO_DAEMON=1 to get more information
    ninja: build stopped: subcommand failed.
    
    

  

  CMake will not be able to correctly generate this project.
```

</p>
</details>

jakirkham:
Thanks Bradley! 🙏

Adding a trivial change PR ( https://github.com/rapidsai/kvikio/pull/225 ) to test CI

bdice:
CI issues were fixed by #226.

bdice:
@madsbk Can you add a "DO NOT MERGE" label to prevent merger?

bdice:
CI is expected to fail for `kvikio` because `cupy` packages for CUDA 12 are not yet available. In the meantime, we can use the CI packages from `libkvikio` to unblock https://github.com/rapidsai/cudf/pull/12922.

jakirkham:
> Can you add a "DO NOT MERGE" label to prevent merger?

Created one and added it here. Label checker seems to pick it up ok

jakirkham:
Looks like PR ( https://github.com/rapidsai/kvikio/pull/241 ) introduced some conflicts here

conda/recipes/kvikio/meta.yaml
```diff
@@ -44,23 +45,33 @@ requirements:
...
     - setuptools
     - pip
     - cython >=0.29,<0.30
-    - cudatoolkit {{ cuda_version }}.*
+    {% if cuda_major == "11" %}
+    - cudatoolkit ={{ cuda_version }}
+    {% endif %}
     - scikit-build >=0.13.1
-    - libkvikio {{ version }}
+    - libkvikio {{ version }}*=cuda{{ cuda_major }}*
```
jakirkham, line 64: Are we able to drop the build string here?

```suggestion
    - libkvikio {{ version }}
```

vyasr: Yup using cuda-version now.

raydouglass (approved):

conda/recipes/libkvikio/meta.yaml
```diff
@@ -75,13 +94,21 @@ outputs:
...
       number: {{ GIT_DESCRIBE_NUMBER }}
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+      ignore_run_exports_from:
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
+        {% endif %}
     requirements:
       build:
         - cmake {{ cmake_version }}
       run:
```
bdice, line 104: This is missing DSOs, like we had for librmm-tests.
```
WARNING (libkvikio-tests,bin/tests/libkvikio/BASIC_IO_TEST): $RPATH/libcudart.so.12 not found in packages, sysroot(s) nor the missing_dso_whitelist.
.. is this binary repackaging?
```
For that package, we solved it with:
```suggestion
      host:
        - cuda-version ={{ cuda_version }}
        {% if cuda_major == "11" %}
        - cudatoolkit ={{ cuda_version }}
        {% else %}
        - cuda-cudart-dev
        {% endif %}
      run:
```
Then use `pin_compatible` in `run:`.

But if we're taking the approach of putting `cuda-cudart-dev` in `host` in order to get its `run_export` of `cuda-cudart` into `run`, should we do the same with `libcufile` and list its dev package in host instead of run? 😕 Figuring out how to be consistent here is important to set an example for the rest of RAPIDS.

bdice (commented):
I can't request changes on my own PR -- but I have several questions for us to consider.

conda/recipes/libkvikio/meta.yaml
jakirkham: Let's make the change proposed here

Would file [an issue on the `libcufile` feedstock]( https://github.com/conda-forge/libcufile-feedstock/issues/new ) for any follow up work there

bdice: There's no work needed on `libcufile` here, since [the `-dev` package already has a run-export](https://github.com/conda-forge/libcufile-feedstock/blob/5a47af81743d9edfc60b1acf29b79c76fd184b5c/recipe/meta.yaml#L75-L76). My question is about how we should specify RAPIDS dependencies.

We saw that librmm-tests was missing a DSO for `libcudart.so.12`. That is provided by `cuda-cudart`. [John proposed](https://github.com/rapidsai/rmm/pull/1223#pullrequestreview-1489881267) to 
have `host: cuda-cudart-dev` to fix this missing DSO _rather than_ `run: cuda-cudart` as I expected. We merged that change.

Do we think it's better to list CUDA libraries (including `cuda-cudart` and `libcufile`) as "dev" packages in the `host`, rather than just listing the actual runtime library in `run`? I don't necessarily see why this is preferred. I wrote a longer comment here: https://github.com/rapidsai/rmm/pull/1223#issuecomment-1599985240 which proposed two other solutions (putting the "dev" package in `run` for the header-only parent library, or just listing the non-dev package in `run` for anything that is overlinking). Let's discuss.

jakirkham: Yes if we are building against / linking to these libraries at some point (as is the case with `libkvikio-tests`), then we should have them in `host` for that package (`libkvikio-tests` here)

bdice: @jakirkham The issue here is that we have a split recipe. We have `-dev` packages in the top-level host. Should we be re-listing these in `host` for each of the package outputs? I think it would be sufficient to list `-dev` packages in the top-level `host` and then only list the runtime libraries in the `run` of the package output. Does this not work? Currently we have no `host:` section for most package outputs, only the top level.

bdice: Conclusion: conda-build won't correctly detect linked DSOs from the `run:` section of a split package output. We decided to use `host:` here.

See also: https://github.com/conda/conda-build/issues/3478.

In an ideal world, we'd be able to get the `run_exports` from the top-level into the `run:` dependencies of the subpackage outputs, and the DSO link check would be able to see the top-level host/run environment contains the desired DSOs.

bdice: Resolved in 4d8fd6e / fbcb1b0.

vyasr (approved):

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #237, opened 2023-06-07
Fix libcufile dependency.

bdice:
This PR fixes the libcufile dependency specification so that it's only specified for x86_64. It also fixes the redundant cudatoolkit re-specification for aarch64.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -167,6 +167,9 @@ dependencies:
               # so 11.2 uses 11.4 packages (the oldest available).
               - *libcufile_114
               - *libcufile_dev114
+          # Fallback matrix for aarch64, which doesn't support libcufile.
+          - matrix:
+            packages:
   docs:
     common:
       - output_types: [conda, requirements]
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -144,6 +144,7 @@ dependencies:
               cuda: "11.8"
               arch: x86_64
             packages:
+              # libcufile package version reference: https://anaconda.org/nvidia/libcufile/files
               - libcufile=1.4.0.31
               - libcufile-dev=1.4.0.31
           - matrix:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -124,32 +124,44 @@ dependencies:
         matrices:
           - matrix:
               cuda: "11.8"
-              arch: x86_64
-            packages:
-              - cudatoolkit=11.8
-              - libcufile=1.4.0.31
-              - libcufile-dev=1.4.0.31
-          - matrix:
-              cuda: "11.8"
-              arch: aarch64
             packages:
               - cudatoolkit=11.8
           - matrix:
               cuda: "11.5"
             packages:
               - cudatoolkit=11.5
-              - libcufile>=1.1.0.37,<=1.1.1.25
-              - libcufile-dev>=1.1.0.37,<=1.1.1.25
           - matrix:
               cuda: "11.4"
             packages:
               - cudatoolkit=11.4
-              - &libcufile_114 libcufile>=1.0.0.82,<=1.0.2.10
-              - &libcufile_dev114 libcufile-dev>=1.0.0.82,<=1.0.2.10
           - matrix:
               cuda: "11.2"
             packages:
               - cudatoolkit=11.2
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "11.8"
+              arch: x86_64
+            packages:
+              - libcufile=1.4.0.31
+              - libcufile-dev=1.4.0.31
+          - matrix:
+              cuda: "11.5"
+              arch: x86_64
+            packages:
+              - libcufile>=1.1.0.37,<=1.1.1.25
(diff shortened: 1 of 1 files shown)


bdice:
See also: https://github.com/rapidsai/cudf/pull/13523

jakirkham:
cc @vyasr

dependencies.yaml
```diff
@@ -124,32 +124,44 @@ dependencies:
...
+              - cudatoolkit=11.5
+          - matrix:
+              cuda: "11.4"
+            packages:
+              - cudatoolkit=11.4
+          - matrix:
+              cuda: "11.2"
+            packages:
+              - cudatoolkit=11.2
+      - output_types: conda
```
wence-, line 141: I don't understand the schema, but it looks like the `output_types: conda` key is repeated here?

dependencies.yaml
```diff
@@ -124,32 +124,44 @@ dependencies:
         matrices:
           - matrix:
               cuda: "11.8"
-              arch: x86_64
```
wence-, line 127: So we now don't specialise here for x86 (and so this works for aarch and x86?)

dependencies.yaml
```diff
@@ -124,32 +124,44 @@ dependencies:
...
+      - output_types: conda
+        matrices:
           - matrix:
               cuda: "11.8"
-              arch: aarch64
+              arch: x86_64
             packages:
-              - cudatoolkit=11.8
+              - libcufile=1.4.0.31
+              - libcufile-dev=1.4.0.31
```
wence-, line 148: But libcufile is only available on x86? And so we specialise here?

dependencies.yaml
```diff
@@ -124,32 +124,44 @@ dependencies:
...
+              - cudatoolkit=11.2
+      - output_types: conda
+        matrices:
           - matrix:
               cuda: "11.8"
-              arch: aarch64
+              arch: x86_64
             packages:
-              - cudatoolkit=11.8
+              - libcufile=1.4.0.31
```
wence-, line 147: Can you add a comment linking to the page that enumerates these version numbers?

wence- (commented):
A minor query or two.

dependencies.yaml
bdice: This is currently scraped from https://anaconda.org/nvidia/libcufile/files but we'll be using https://anaconda.org/conda-forge/libcufile for CUDA 12 versions. I can add a comment for now but it'll need to be updated later.

bdice:
```suggestion
              # libcufile package version reference: https://anaconda.org/nvidia/libcufile/files
              - libcufile=1.4.0.31
```

dependencies.yaml
bdice: The reason this structure is better is that the cudatoolkit dependency isn't arch-specific. We were just using arch information to ensure that libcufile was only installed on x86_64. However, this will change at some point when libcufile adds aarch64 support. Separating these now ensures that we can cleanly define the dependencies with smaller changes in the future.

bdice: The other important piece to know is that the first matching "matrix" combination is selected, meaning that we cannot just write it in a single list (simplified to 11.8/11.5 for demonstration):

```yaml
  # This is not enough because only the first matching matrix element can be selected
  cudatoolkit:
    specific:
      - output_types: conda
        matrices:
          - matrix:
              cuda: "11.8"
            packages:
              - cudatoolkit=11.8
          - matrix:
              cuda: "11.5"
            packages:
              - cudatoolkit=11.5
          - matrix:
              cuda: "11.8"
              arch: x86_64
            packages:
              # libcufile package version reference: https://anaconda.org/nvidia/libcufile/files
              - libcufile=1.4.0.31
              - libcufile-dev=1.4.0.31
          - matrix:
              cuda: "11.5"
              arch: x86_64
            packages:
              - libcufile>=1.1.0.37,<=1.1.1.25
              - libcufile-dev>=1.1.0.37,<=1.1.1.25
```

```yaml
  # This is a viable alternative based on the "first matching" criterion but it repeats
  # cudatoolkit dependencies in two package lists, which is not ideal
  cudatoolkit:
    specific:
      - output_types: conda
        matrices:
          - matrix:
              cuda: "11.8"
              arch: x86_64
            packages:
              - cudatoolkit=11.8
              # libcufile package version reference: https://anaconda.org/nvidia/libcufile/files
              - libcufile=1.4.0.31
              - libcufile-dev=1.4.0.31
          - matrix:
              cuda: "11.8"
            packages:
              - cudatoolkit=11.8
          - matrix:
              cuda: "11.5"
              arch: x86_64
            packages:
              - cudatoolkit=11.5
              - libcufile>=1.1.0.37,<=1.1.1.25
              - libcufile-dev>=1.1.0.37,<=1.1.1.25
          - matrix:
              cuda: "11.5"
            packages:
              - cudatoolkit=11.5
```

wence-: Thanks for the explanation!

wence- (approved):

PR #241, opened 2023-06-13
Clean up dependency lists

vyasr:
cudf is not actually used anywhere in tests, only in a single notebook. Similarly, dask is only used by legate tests


dependencies.yaml
```diff
@@ -13,6 +13,7 @@ files:
       - py_version
       - run
       - test_python
+      - notebooks
```
bdice, line 16: Alphabetize.
```suggestion
      - notebooks
      - py_version
      - run
      - test_python
```

dependencies.yaml
```diff
@@ -219,8 +220,12 @@ dependencies:
...
-          - cudf==23.8.*
           - dask>=2022.05.2
           - distributed>=2022.05.2
           - pytest
           - pytest-cov
+  notebooks:
+    common:
+      - output_types: conda
+        packages:
+          - cudf==23.8.*
```
bdice, line 231: Add the fix for `update-version.sh` from #224.

bdice (changes requested):

python/pyproject.toml
```diff
@@ -40,7 +40,6 @@ classifiers = [
 [project.optional-dependencies]
 test = [
     "cuda-python>=11.7.1,<12.0",
-    "cudf==23.8.*",
     "dask>=2022.05.2",
     "distributed>=2022.05.2",
```
bdice, line 44: It seems like `distributed` isn't used in `kvikio` -- only the `legate_kvikio` benchmarks.

dependencies.yaml
vyasr: There isn't actually an update-version.sh fix in #224 since cudf was added back as a dependency, but it looks like I can at least remove some of it so I'll do that now.

bdice: My mistake. I misread a0a7b596a99a99c34b85ac69acbaed6999b5ea94, you were adding it _back_ after removing it.

dependencies.yaml
```diff
@@ -219,8 +222,16 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - cuda-python>=11.7.1,<12.0
```
bdice, line 224: We need `dask` in `test_python`, just not `distributed`.

See imports of `dask` in `python/benchmarks/single-node-io.py` and `python/tests/test_benchmarks.py`.
```suggestion
          - cuda-python>=11.7.1,<12.0
          - &dask dask >=2022.05.2
```

dependencies.yaml
```diff
@@ -219,8 +222,16 @@ dependencies:
...
-          - cudf==23.8.*
-          - dask>=2022.05.2
-          - distributed>=2022.05.2
           - pytest
           - pytest-cov
+  test_python_legate:
+    common:
+      - output_types: [conda, requirements, pyproject]
+        packages:
+          - dask>=2022.05.2
```
bdice, line 231:
```suggestion
          - *dask
```

bdice (commented):
A couple small suggestions.

bdice (approved):

bdice:
LGTM. I’ll merge to unblock #224.

bdice:
Just kidding. I see we need several reviews first.

jakirkham (approved):

wence- (approved):
lgtm

raydouglass (approved):

vyasr:
We need @jakirkham to re-approve (or @madsbk to approve) on behalf of legate codeowners before this will merge, now.

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #243, opened 2023-06-14
Unify `KVIKIO_CUFILE_FOUND`

madsbk:
Fixing a mismatch where CMake cannot find cuFile but later when compiling, GCC finds `cufile.h` using `__has_include()`.  

To fix this, CMake now defines `KVIKIO_CUFILE_FOUND` and we stop using `__has_include()`.


wence- (approved):
Thanks, looks right.

madsbk:
Thanks @wence-

PR #245, opened 2023-06-14
Use nvcomp conda package.

### LINKED ISSUE #242: Use `nvcomp` package
Currently nvCOMP is fetched by CMake

https://github.com/rapidsai/kvikio/blob/d9eee8b0392dd581af66cf91838f964b4e415c0e/python/cmake/CMakeLists.txt#L15

Now that there is [an `nvcomp` conda package]( https://anaconda.org/conda-forge/nvcomp ), it would be good to switch over to using that instead

bdice:
## Description
This PR uses conda-forge packages of `nvcomp` rather than fetching a tarball. This means that the nvcomp binary should not be shipped in the kvikio conda package, but is instead listed as a dependency. This will reduce kvikio's conda package size.

Resolves #242.


bdice:
This is probably blocked on https://github.com/rapidsai/cudf/pull/13566#issuecomment-1591193680, I will revisit this PR later but wanted to get the draft open for now.

jakirkham:
Thanks Bradley! 🙏

Do we need to do anything with [`get_nvcomp.cmake`]( https://github.com/rapidsai/kvikio/blob/4a4d28953e235f3c7c4148b9b96cd179b55118d2/python/cmake/thirdparty/get_nvcomp.cmake )?

bdice:
> Thanks Bradley! 🙏
> 
> Do we need to do anything with [`get_nvcomp.cmake`](https://github.com/rapidsai/kvikio/blob/4a4d28953e235f3c7c4148b9b96cd179b55118d2/python/cmake/thirdparty/get_nvcomp.cmake)?

Nope. No changes should be needed there. `rapids_cpm_nvcomp` does exactly what we want.

jakirkham (approved):

raydouglass (approved):

bdice:
I verified that the Python build logs look correct. The conda output shows lines like `Needed DSO lib/libnvcomp.so found in conda-forge::nvcomp-2.6.1-h0800d71_2` which indicates that packaging works as intended.

jakirkham:
Thanks Bradley! 🙏

PR #253, opened 2023-07-19
Enable roundtrip for nvCOMP batch codecs.

Alexey-Kamenev:
This PR enables a roundtrip between codecs other than LZ4. For LZ4 codec, there is already such support.

The "roundtrip" here means ability to compress/store data using default numcodecs codec (e.g. `zstd`) and decompress using nvCOMP batch codecs. Reverse would also be true: ability to compress data using nvCOMP and decompress using CPU numcodecs codec.

The original implementation of the nvCOMP codec was based on the assumption, taken from numcodecs LZ4 codec, that in order for the codecs to be compatible, each codec must write a small header (4 bytes) in the compressed chunk, which contains the original, uncompressed size of the data. However, after subsequent analysis of other numcodecs codecs, it turned out LZ4 was actually an exception! No other codec writes a header or any additional data to the compressed chunk.

This PR:
* moves the header creation down to a specific algorithm, LZ4
* switches to nvCOMP `nvcompBatched*GetDecompressSizeAsync` family of functions to get the original size of data during decompression.


python/kvikio/nvcomp_codec.py
```diff
@@ -95,28 +95,16 @@ def encode_batch(self, bufs: List[Any]) -> List[Any]:
...
-        # Includes header with the original buffer size,
-        # same as in numcodecs codec. This enables data compatibility between
-        # numcodecs default codecs and this nvCOMP batch codec.
-        # TODO(akamenev): probably should use contiguous buffer which stores all chunks?
-        comp_chunks_header = [
-            cp.empty(self.HEADER_SIZE_BYTES + comp_chunk_size, dtype=cp.uint8)
-            for _ in range(num_chunks)
-        ]
-        # comp_chunks is used as a container that stores pointers to actual chunks.
+        comp_chunks = cp.empty((num_chunks, comp_chunk_size), dtype=cp.uint8)
```
wence-, line 103: OK, so you allocate a single contiguous piece of memory for the compressed chunks and then pass offset pointers to nvcomp...

python/kvikio/nvcomp_codec.py
```diff
@@ -196,37 +176,37 @@ def decode_batch(
...
             bufs = [cp.asarray(ensure_contiguous_ndarray_like(b)) for b in bufs]
 
-        # Get uncompressed chunk sizes from the header.
-        uncomp_chunk_sizes = [
-            int(b[: self.HEADER_SIZE_BYTES].view(dtype=cp.uint32)[0]) for b in bufs
-        ]
-        max_chunk_size = max(uncomp_chunk_sizes)
+        # Prepare compressed chunks buffers.
+        comp_chunks = cp.array([b.data.ptr for b in bufs], dtype=cp.uint64)
+        comp_chunk_sizes = cp.array([b.size for b in bufs], dtype=cp.uint64)
```
wence-, line 181: Technically these should probably be [`cp.uintp`](https://numpy.org/devdocs/reference/arrays.scalars.html#numpy.uintp), et seq. throughout:
```suggestion
        comp_chunks = cp.array([b.data.ptr for b in bufs], dtype=cp.uintp)
        comp_chunk_sizes = cp.array([b.size for b in bufs], dtype=cp.uintp)
```

wence- (changes requested):
Thanks, overall I think this looks good. One thing I wonder is whether we should just do all of the pointer manipulation in host buffers and then do a single round of H2D memcopies before calling into `nvcomp`. Since I imagine the number of batches is not _that_ large this is probably faster than lots of tiny (explicit and implicit) memcopies back and forth.

WDYT?

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Agree about pointers, updated the code accordingly. Would still prefer using `uint64` type for sizes, as those are actual sizes, not pointers.

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Yes, I decided to allocate one chunk of memory, reducing the number of allocations. The potential downside is that if there is a memory pressure and fragmentation, the allocator might not find large enough chunk while with smaller, non-contiguous chunks it might succeed.

python/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -346,9 +425,33 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
         comp_chunk_sizes,
         stream
     ):
+        if self.has_header:
+            # If there is a header, we need to:
+            # 1. Copy the uncompressed chunk size to the compressed chunk header.
+            # 2. Update target pointers in comp_chunks to skip the header portion,
+            # which is not compressed.
+            # TODO(akamenev): there are some D2H/H2D copies which
+            # potentially can be removed or replaced with D2D memcpyAsync.
```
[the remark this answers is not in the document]
Alexey-Kamenev: Thank you for the review! Yes, I absolutely agree - my initial version had too many redundant copies. I've re-factored the code to address this issue and replaced device buffers with host buffers in places where it allowed to eliminate H2D/D2H copies. Also reduced redundant synchronization.

python/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -376,6 +526,92 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
...
+        for i in range(batch_size):
+            # Get pointer to the header and copy the data.
+            memcpyAsync(
+                psize,
+                <uintptr_t>comp_chunks[i],
+                sizeof(uint32_t),
+                cudaMemcpyKind.cudaMemcpyDeviceToDevice,
+                stream.ptr
+            )
+            psize += sizeof(uint32_t)
```
wence-, line 558: Perhaps more esoteric, but I wonder if is worthwhile using `cudaMemcpy2DAsync` instead here.

Which would be, I think

```
rows, columns = comp_chunks.shape
cudaMemcpy2DAsync(psize, sizeof(uint32_t), <uintptr_t>comp_chunks[0], 
                  columns, sizeof(uint32_t), rows, 
                  cudaMemcpyKind.cudaMemcpyDeviceToDevice, stream.ptr)
```
Then there's no need for the loop over batches.

WDYT?

python/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -333,8 +430,61 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
...
+            psize = to_ptr(uncomp_chunk_sizes)
+            for i in range(batch_size):
+                # Copy the original data size to the header.
+                memcpyAsync(
+                    <uintptr_t>comp_chunks[i],
+                    psize,
+                    sizeof(uint32_t),
+                    cudaMemcpyKind.cudaMemcpyDeviceToDevice,
+                    stream.ptr
+                )
```
wence-, line 466: Could perhaps also use cudamemcpy2dasync here too (see below).

wence- (commented):
Thanks! Two minor typographic fixes, plus (apologies I didn't spot this first time round), I wonder if we want to use `memcpy2DAsync` to avoid the batch loop and kick everything off in one go. I don't think that's a necessary change though, so leave it up to you.

python/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -333,8 +430,61 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
...
+                psize += sizeof(uint64_t)
+                # Update chunk pointer to skip the header.
+                comp_chunks[i] += self.HEADER_SIZE_BYTES
```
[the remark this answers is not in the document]
Alexey-Kamenev: Yes, that's correct. I've made the suggested modification. I should also mention that the current header structure, introduced in numcodecs, is rigid and essentially prohibits any modification or extension as it lacks any versioning mechanism. So we are now stuck with uint32_t header with fixed semantics, unless we are ready to break compatibility, of course.

python/kvikio/_lib/libnvcomp_ll.pyx
Alexey-Kamenev: This is a good idea, but in this particular case it won't work: the chunks that are passed to `get_decompress_size` will be of different sizes (unless the source data was identical set of bytes or something like that), so it's not really a 2D array.

python/kvikio/_lib/libnvcomp_ll.pyx
Alexey-Kamenev: In this case, `comp_chunks` is currently a host array which keeps pointers to every chunk. Since those chunks were allocated originally as a 2D array, we potentially could pass that one here instead. But then we still need access to each chunk pointer separately to update it before passing to nvCOMP compress and then resetting it back. 
Possibly we could do that by carefully crafting and passing host/device arrays, but I'd like to keep it as is for now. I suggest we create a separate issue/task for that.

python/kvikio/_lib/libnvcomp_ll.pyx
wence-: Ah ok, scratch that then.

Alexey-Kamenev:
Thank you for the review! I've updated the PR's target branch to `branch-23.10`.

madsbk (approved):
LGTM, thanks @Alexey-Kamenev

wence- (approved):
Thanks for the updates, looks good to me too.

Alexey-Kamenev:
Thank you for the review, @wence- and @madsbk ! 
Let me know if there is anything else I need to do to get it merged.

madsbk:
Thanks @Alexey-Kamenev

PR #257, opened 2023-07-28
Modernize Python build

### LINKED ISSUE #186: Use `rapids-cython` & other more recent best practices
Would it make sense to start using `rapids-cython` here as in cuDF?

https://github.com/rapidsai/cudf/blob/766af666575b5db6f265e1fbf6466ccfd46eae30/python/cudf/CMakeLists.txt#L72

Any other best practices we should adopt here to modernize our scikit-build/CMake files?

Maybe some of these items are worth addressing ( https://github.com/rapidsai/kvikio/pull/58#discussion_r845608125 )

vyasr:
Switch to using rapids-cmake's rapids-cython modules for the Python build. It also uses rapids-cuda to support NATIVE or RAPIDS as the list of CUDA architectures.

Resolves #186


python/CMakeLists.txt
bdice, line 1: Do we need changes for `legate/` as well?

bdice (approved):
Looks good, just one question about the legate package.

python/CMakeLists.txt
vyasr: AFAICT that package uses a completely separate build system for compatibility with legate. So the answer is I don't think so, but if it does need updates I would probably defer to Mads.

PR #258, opened 2023-07-28
Update to Cython 3.0.0

vyasr:
This PR bumps the pinning and fixes the new warnings that appeared

### DIFF
--- a/python/kvikio/_lib/libnvcomp_ll.pyx
+++ b/python/kvikio/_lib/libnvcomp_ll.pyx
@@ -76,7 +76,7 @@ class nvCompBatchAlgorithm(ABC):
         self,
         size_t batch_size,
         size_t max_uncompressed_chunk_bytes,
-    ) -> (nvcompStatus_t, size_t):
+    ) -> tuple[nvcompStatus_t, size_t]:
         """Algorithm-specific implementation."""
         ...
 
@@ -312,7 +312,7 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
         self,
         size_t batch_size,
         size_t max_uncompressed_chunk_bytes,
-    ) -> (nvcompStatus_t, size_t):
+    ) -> tuple[nvcompStatus_t, size_t]:
         cdef size_t temp_bytes = 0
 
         err = nvcompBatchedLZ4CompressGetTempSize(
@@ -434,7 +434,7 @@ class nvCompBatchAlgorithmGdeflate(nvCompBatchAlgorithm):
         self,
         size_t batch_size,
         size_t max_uncompressed_chunk_bytes,
-    ) -> (nvcompStatus_t, size_t):
+    ) -> tuple[nvcompStatus_t, size_t]:
         cdef size_t temp_bytes = 0
 
         err = nvcompBatchedGdeflateCompressGetTempSize(
@@ -552,7 +552,7 @@ class nvCompBatchAlgorithmZstd(nvCompBatchAlgorithm):
         self,
         size_t batch_size,
         size_t max_uncompressed_chunk_bytes,
-    ) -> (nvcompStatus_t, size_t):
+    ) -> tuple[nvcompStatus_t, size_t]:
         cdef size_t temp_bytes = 0
 
         err = nvcompBatchedZstdCompressGetTempSize(
@@ -670,7 +670,7 @@ class nvCompBatchAlgorithmSnappy(nvCompBatchAlgorithm):
         self,
         size_t batch_size,
         size_t max_uncompressed_chunk_bytes,
-    ) -> (nvcompStatus_t, size_t):
+    ) -> tuple[nvcompStatus_t, size_t]:
         cdef size_t temp_bytes = 0
 
         err = nvcompBatchedSnappyCompressGetTempSize(
--- a/python/kvikio/_lib/libnvcomp_ll.pyx
+++ b/python/kvikio/_lib/libnvcomp_ll.pyx
@@ -1,6 +1,8 @@
 # Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 from enum import IntEnum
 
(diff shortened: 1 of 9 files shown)


python/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -76,7 +76,7 @@ class nvCompBatchAlgorithm(ABC):
         self,
         size_t batch_size,
         size_t max_uncompressed_chunk_bytes,
-    ) -> (nvcompStatus_t, size_t):
+    ) -> tuple[nvcompStatus_t, size_t]:
```
Alexey-Kamenev, line 79: So I guess this is where the dependency on Python 3.9+ is introduced (PEP-585), right? Up until now, `kvikio` did not use any Python feature which would specifically require 3.9+.

vyasr: That's an interesting point I didn't consider. I believe in this particular context this change is actually backwards compatible because using tuple directly has been supporting since Python 3.7, it [just wasn't fully featured](https://peps.python.org/pep-0585/#implementation):

> Usefulness of this syntax before [PEP 585](https://peps.python.org/pep-0585) is limited as external tooling like Mypy does not recognize standard collections as generic. Moreover, certain features of typing like type aliases or casting require putting types outside of annotations, in runtime context.

I would just need to add a `from __future__ import annotations` to this file. I don't know if Cython 3.0.0 relies on Python 3.9+ functionality to recognize this type, but it wouldn't matter much anyway here since this function isn't a `cdef` function. If you prefer I could switch to using `typing.tuple` for now instead though. That may eventually be deprecated, but not for a couple of years at least, and it would be fully backwards compatible.

raydouglass (approved):

python/kvikio/_lib/nvcomp_cxx_api.pxd
```diff
@@ -73,7 +73,6 @@ cdef extern from "nvcomp/nvcompManager.hpp" namespace 'nvcomp':
             size_t uncompressed_buffer_size) except +
         nvcompStatus_t* get_status() const
         CompressionConfig(CompressionConfig& other)
-        CompressionConfig& operator=(CompressionConfig&& other) except +
```
vuule, line 76: why are these removed?

vyasr: Cython doesn't actually support rvalue reference parameters. It never did, but previously it wouldn't warn about it. Now it does.

Note that despite this, you can still get the desired behavior by writing Cython code with an explicit move `x ==move(CompressionConfig())`. Cython will generate the C++ code assuming it'll use one of the other operators for an lvalue, but then the C++ compiler will use the rvalue version instead.

vuule: The only issue might be that there's zero visibility of the rvalue versions on the Cython side. Should we leave a comment that a move is applicable when calling operator=?

vyasr: We could do that. We may need a more comprehensive strategy for this though since it's a problem that affects all of RAPIDS. There are multiple cases where we have functions that accept rvalues that we may want to leverage in Cython even though Cython doesn't think it knows how. @shwina do you have thoughts on how best to handle this?

shwina: I don't have any fantastic suggestions. Cython doesn't really know about rvalues generally and getting it to generate the right code involves using explicit `move()` calls.

Agree with @vuule - perhaps we just leave a comment?

jakirkham: Proposing such a change below. Maybe we can iterate on that change?

python/kvikio/_lib/nvcomp_cxx_api.pxd
```diff
@@ -73,7 +73,6 @@ cdef extern from "nvcomp/nvcompManager.hpp" namespace 'nvcomp':
...
         CompressionConfig& operator=(const CompressionConfig& other) except +
```
jakirkham, line 76:
```suggestion
        CompressionConfig& operator=(const CompressionConfig& other) except +
        # Commented as Cython doesn't support rvalues, but
        # a user can call `move` with the existing operator and generate correct C++ code
        # xref: https://github.com/cython/cython/issues/1445
        # CompressionConfig& operator=(CompressionConfig&& other) except +
```

python/kvikio/_lib/nvcomp_cxx_api.pxd
```diff
@@ -82,7 +81,6 @@ cdef extern from "nvcomp/nvcompManager.hpp" namespace 'nvcomp':
         DecompressionConfig(PinnedPtrPool[nvcompStatus_t]& pool) except +
         nvcompStatus_t* get_status() const
         DecompressionConfig(DecompressionConfig& other)
-        DecompressionConfig& operator=(DecompressionConfig&& other) except +
         DecompressionConfig& operator=(const DecompressionConfig& other) except +
```
jakirkham, line 84:
```suggestion
        CompressionConfig& operator=(const CompressionConfig& other) except +
        # Commented as Cython doesn't support rvalues, but
        # a user can call `move` with the existing operator and generate correct C++ code
        # xref: https://github.com/cython/cython/issues/1445
        # DecompressionConfig& operator=(DecompressionConfig&& other) except +
```

jakirkham (commented):
Thanks Vyas! 🙏

Proposed a comment that may capture what we are after with rvalues

python/kvikio/_lib/libnvcomp_ll.pyx
Alexey-Kamenev: Apologies for not responding sooner.  I think adding `from __future__ import annotations` is a good idea. The package already requires 3.9+ anyway, so anyone willing to use it with 3.8 will be aware of potential compatibility issues.

python/kvikio/_lib/nvcomp_cxx_api.pxd
vyasr: Done.

jakirkham:
@Alexey-Kamenev @vuule , could you please review the latest changes?

Alexey-Kamenev (approved):
Looks good to me!

vuule (approved):

jakirkham (approved):

PR #259, opened 2023-07-31
Initial changes to support cufile stream I/O.

tell-rebanta:


cpp/CMakeLists.txt
```diff
@@ -71,6 +71,14 @@ else()
     set(cuFile_BATCH_API_FOUND TRUE)
   endif()
   message(STATUS "Found cuFile's Batch API: ${cuFile_BATCH_API_FOUND}")
+  file(READ "${cuFile_INCLUDE_DIRS}/cufile.h" CUFILE_H_STR)
```
madsbk, line 74: I guess we don't need this line?
```suggestion
```

cpp/examples/basic_io.cpp
```diff
@@ -192,4 +195,24 @@ int main()
     check(statuses.empty());
     cout << "Batch canceling of all 4 operations" << endl;
   }
+
+  if (kvikio::is_stream_available()) {
+     off_t f_off = 0, d_off = 0;
+
+     kvikio::StreamHandle s_handle_wr("/tmp/test-file", "w", kvikio::StreamHandle::m644, true);
+     check(cudaMemcpy(a_dev, a, SIZE, cudaMemcpyHostToDevice) == cudaSuccess);
+     kvikio::buffer_register(a_dev, SIZE);
```
madsbk, line 204: Do we need to register the a buffer in order to use it with stream IO?

cpp/include/kvikio/stream.hpp
```diff
@@ -0,0 +1,310 @@
...
+#include <kvikio/shim/cuda.hpp>
+
+namespace kvikio {
+
+/**
+ * @brief Handle of an open file registered with cufile.
+ *
+ * In order to utilize cufile and GDS, a file must be registered with cufile.
+ */
+class StreamHandle {
```
madsbk, line 46: Would it make sense to have `StreamHandle` inherent from `FileHandle` ?

madsbk (commented):
Thanks @tell-rebanta. Overall, I think it looks good, but I am wondering if it makes sense to de-couple the CUDA stream from `StreamHandle`?  

As it is now, you are forced to create a stream per file, but I think a common use case is to read/write to multiple files using the same stream. 

What about implementing `read_async()` and `write_async()` in the existing class `FileHandle` instead? `read_async()` and `write_async()` could then take a `CUstream` as argument. 

##### Side note: 
To check and fix the `check-style` run:  
```bash
pre-commit run --all-file
```

```

**Review by madsbk — CHANGES_REQUESTED** [review]
Thanks @tell-rebanta, I like that the stream is now an explicit argument for `read_async` and `write_async`!
We need to guard the stream API using `KVIKIO_CUFILE_STREAM_API_FOUND` and I suggest that we exclude `StreamHandle` from this PR. 
Let's discuss how to design the stream registration in a follow-up PR (I think we need to de-couple it from the file handle).

File: cpp/include/kvikio/file_handle.hpp [hunk 4]
```diff
@@ -496,6 +496,79 @@ class FileHandle {
     return parallel_io(op, devPtr_base, size, file_offset, task_size, devPtr_offset);
   }
 
+  /**
+   * @brief Reads specified bytes from the file into the device memory.
+   *
+   * This API reads asynchronously the data from the GPU memory to the file at a specified offset
+   * and size bytes by using GDS functionality. The API works correctly for unaligned
```
wence-, line 503: This sentence is in contradiction with the rest of the docstring? I think you mean:

> This API reads size bytes asynchronously from the file into device memory writing to a specified offset using GDS functionality.

?

cpp/include/kvikio/file_handle.hpp
```diff
@@ -496,6 +496,79 @@ class FileHandle {
...
+   * that metadata back to permanent storage. The data is not guaranteed to be present
+   * after a system crash unless the application uses an explicit `fsync(2)` call. If the
+   * file is opened with an `O_SYNC` flag, the metadata will be written to the disk before
+   * the call is complete.
+   * Refer to the note in read for more information about `devPtr_offset`.
+   *
+   * @param devPtr_base Base address of buffer in device memory. For registered buffers,
+   * `devPtr_base` must remain set to the base address used in the `buffer_register` call.
+   * @param size Size in bytes to write.
+   * @param file_offset Offset in the file to write from.
```
wence-, line 554:
```suggestion
   * @param file_offset Offset in the file to write at.
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -496,6 +496,79 @@ class FileHandle {
...
+   * @param devPtr_offset Offset relative to the `devPtr_base` pointer to write into.
```
wence-, line 555:
```suggestion
   * @param devPtr_offset Offset relative to the `devPtr_base` pointer to write from.
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -496,6 +496,79 @@ class FileHandle {
...
+   * the relative offset should be specified in the `devPtr_offset`, and the
+   * `devPtr_base` must remain set to the base address that was used in the
+   * `buffer_register` call.
+   *
+   * @param devPtr_base Base address of buffer in device memory. For registered buffers,
+   * `devPtr_base` must remain set to the base address used in the `buffer_register` call.
+   * @param size Size in bytes to read.
+   * @param file_offset Offset in the file to read from.
+   * @param devPtr_offset Offset relative to the `devPtr_base` pointer to read into.
+   * This parameter should be used only with registered buffers.
```
wence-, line 519: Is there a way to distinguish registered buffers in the type system at the moment? That would be preferable over this `void *` C-like interface.

madsbk: No, not ATM but it would be a cool addition: https://github.com/rapidsai/kvikio/issues/266

cpp/include/kvikio/file_handle.hpp
```diff
@@ -496,6 +496,79 @@ class FileHandle {
...
+   */
+  inline void read_async(void* devPtr_base,
+                         std::size_t* size,
+                         off_t* file_offset,
+                         off_t* devPtr_offset,
+                         ssize_t* bytes_read,
+                         CUstream stream)
+  {
+    CUFILE_TRY(cuFileAPI::instance().ReadAsync(
+      _handle, devPtr_base, size, file_offset, devPtr_offset, bytes_read, stream));
```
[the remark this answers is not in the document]
tell-rebanta: Added.

cpp/include/kvikio/file_handle.hpp
```diff
@@ -496,6 +496,79 @@ class FileHandle {
...
+   */
+  inline void write_async(void* devPtr_base,
+                          std::size_t* size,
+                          off_t* file_offset,
+                          off_t* devPtr_offset,
+                          ssize_t* bytes_written,
+                          CUstream stream)
+  {
+    CUFILE_TRY(cuFileAPI::instance().WriteAsync(
+      _handle, devPtr_base, size, file_offset, devPtr_offset, bytes_written, stream));
```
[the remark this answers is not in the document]
tell-rebanta: Added.

cpp/include/kvikio/shim/cuda.hpp
```diff
@@ -70,6 +76,12 @@ class cudaAPI {
     get_symbol(DeviceGet, lib, KVIKIO_STRINGIFY(cuDeviceGet));
     get_symbol(DevicePrimaryCtxRetain, lib, KVIKIO_STRINGIFY(cuDevicePrimaryCtxRetain));
     get_symbol(DevicePrimaryCtxRelease, lib, KVIKIO_STRINGIFY(cuDevicePrimaryCtxRelease));
+    get_symbol(StreamCreate, lib, KVIKIO_STRINGIFY(cuStreamCreate));
+    get_symbol(StreamCreateWithPriority, lib, KVIKIO_STRINGIFY(cuStreamCreateWithPriority));
+    get_symbol(MemHostRegister, lib, KVIKIO_STRINGIFY(cuMemHostRegister));
+    get_symbol(MemHostUnregister, lib, KVIKIO_STRINGIFY(cuMemHostUnregister));
+    get_symbol(StreamSynchronize, lib, KVIKIO_STRINGIFY(cuStreamSynchronize));
+    get_symbol(StreamDestroy, lib, KVIKIO_STRINGIFY(cuStreamDestroy));
```
madsbk, line 84:
```suggestion
```

cpp/include/kvikio/shim/cuda.hpp
```diff
@@ -46,6 +46,12 @@ class cudaAPI {
   decltype(cuDeviceGet)* DeviceGet{nullptr};
   decltype(cuDevicePrimaryCtxRetain)* DevicePrimaryCtxRetain{nullptr};
   decltype(cuDevicePrimaryCtxRelease)* DevicePrimaryCtxRelease{nullptr};
+  decltype(cuStreamCreate)* StreamCreate{nullptr};
+  decltype(cuStreamCreateWithPriority)* StreamCreateWithPriority{nullptr};
+  decltype(cuMemHostRegister)* MemHostRegister{nullptr};
+  decltype(cuMemHostUnregister)* MemHostUnregister{nullptr};
+  decltype(cuStreamSynchronize)* StreamSynchronize{nullptr};
+  decltype(cuStreamDestroy)* StreamDestroy{nullptr};
```
madsbk, line 54:
```suggestion
```

madsbk (changes requested):
I think we can remove the CUDA stream API

cpp/include/kvikio/shim/cuda.hpp
tell-rebanta: Done.

cpp/include/kvikio/shim/cuda.hpp
tell-rebanta: Done.

madsbk:
Thanks @wence-, added you suggestions.

madsbk (approved):

PR #260, opened 2023-08-01
Remove sphinx pinning

vyasr:
There's no need for this upper bound pinning, the docs build with the latest Sphinx.


jakirkham:
@AjayThorve it looks like this version constraint on Sphinx comes from PR ( https://github.com/rapidsai/kvikio/pull/164 ). Do you recall if there was something that motivated that constraint (some bug to workarounds or something)?

AjayThorve (approved):

vyasr:
Awesome thanks for confirming everyone.

PR #267, opened 2023-08-23
Zarr+CuPy+GDS+nvCOMP made easy

madsbk:
Introducing `open_cupy_array()`, which is a CUDA friendly version of `zarr.open_array` that reads and writes to CuPy arrays.

### DIFF
--- a/python/kvikio/zarr.py
+++ b/python/kvikio/zarr.py
@@ -336,7 +336,7 @@ def open_cupy_array(
         exist); 'w' means create (overwrite if exists); 'w-' means create
         (fail if exists).
     compressor
-        The compressor use when create a Zarr file or None if no compressor
+        The compressor used when creating a Zarr file or None if no compressor
         is to be used. If a `CompatCompressor` is given, `CompatCompressor.gpu`
         is used for compression and decompression; and `CompatCompressor.cpu`
         is written as the compressor in the Zarr file metadata on disk.
--- a/python/kvikio/zarr.py
+++ b/python/kvikio/zarr.py
@@ -37,22 +37,39 @@ class GDSStore(zarr.storage.DirectoryStore):
     It uses KvikIO for reads and writes, which in turn will use GDS
     when applicable.
 
+    Parameters
+    ----------
+    path : string
+        Location of directory to use as the root of the storage hierarchy.
+    normalize_keys : bool, optional
+        If True, all store keys will be normalized to use lower case characters
+        (e.g. 'foo' and 'FOO' will be treated as equivalent). This can be
+        useful to avoid potential discrepancies between case-sensitive and
+        case-insensitive file system. Default value is False.
+    dimension_separator : {'.', '/'}, optional
+        Separator placed between the dimensions of a chunk.
+
     Notes
     -----
-    GDSStore doesn't implement `_fromfile()` thus non-array data such as
-    meta data is always read into host memory.
-    This is because only zarr.Array use getitems() to retrieve data.
+    Atomic writes are used, which means that data are first written to a
+    temporary file, then moved into place when the write is successfully
+    completed. Files are only held open while they are being read or written and are
+    closed immediately afterwards, so there is no need to manually close any files.
+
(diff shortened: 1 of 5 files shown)


quasiben (approved):

python/kvikio/zarr.py
```diff
@@ -237,3 +297,107 @@ def get_nvcomp_manager(self):
...
+    mode: Literal["r", "r+", "a", "w", "w-"] = "a",
+    compressor: Codec | CompatCompressor = Snappy(device_ordinal=0),
+    meta_array=cupy.empty(()),
+    **kwargs,
+) -> zarr.Array:
+    """Open an Zarr array as a CuPy-like array using file-mode-like semantics.
+
+    This function is a CUDA friendly version of `zarr.open_array` that reads
+    and writes to CuPy arrays. Beside the arguments listed below, the arguments
+    have the same semantic as in `zarr.open_array`.
```
quasiben, line 326:
```suggestion
    have the same semantics as in `zarr.open_array`.
```

python/kvikio/zarr.py
```diff
@@ -237,3 +297,107 @@ def get_nvcomp_manager(self):
...
+        )
+        # If we are reading a LZ4-CPU compressed file, we overwrite the metadata
+        # on-the-fly to make Zarr use LZ4-GPU for both compression and decompression.
+        compat_lz4 = CompatCompressor.lz4()
+        if ret.compressor == compat_lz4.cpu:
+            ret = zarr.open_array(
+                store=kvikio.zarr.GDSStore(
+                    path=store,
+                    compressor_config_overwrite=compat_lz4.cpu.get_config(),
+                    decompressor_config_overwrite=compat_lz4.gpu.get_config(),
```
wence-, line 377: And this says, "we're compressed with lz4, but we have a gpu-accelerated version, so actually do any compression/decompression with the nvcomp version of lz4".

This seems like a bad metadata schema (but perhaps you cannot change it), why should the metadata specify the instance of the compressor to use, rather than the _type_ of compression (and any parameters).

madsbk: Correct, this is a limitation of Zarr-Python :/

python/kvikio/zarr.py
```diff
@@ -237,3 +297,107 @@ def get_nvcomp_manager(self):
...
+            **kwargs,
+        )
+        # If we are reading a LZ4-CPU compressed file, we overwrite the metadata
+        # on-the-fly to make Zarr use LZ4-GPU for both compression and decompression.
+        compat_lz4 = CompatCompressor.lz4()
+        if ret.compressor == compat_lz4.cpu:
+            ret = zarr.open_array(
+                store=kvikio.zarr.GDSStore(
+                    path=store,
+                    compressor_config_overwrite=compat_lz4.cpu.get_config(),
```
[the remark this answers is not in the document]
madsbk: yes

python/kvikio/zarr.py
```diff
@@ -237,3 +297,107 @@ def get_nvcomp_manager(self):
...
+    if mode in ("r", "r+"):
+        ret = zarr.open_array(
+            store=kvikio.zarr.GDSStore(path=store),
+            mode=mode,
+            meta_array=meta_array,
+            **kwargs,
+        )
+        # If we are reading a LZ4-CPU compressed file, we overwrite the metadata
+        # on-the-fly to make Zarr use LZ4-GPU for both compression and decompression.
+        compat_lz4 = CompatCompressor.lz4()
```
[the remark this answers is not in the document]
madsbk: Initially, I had the logic in `GDSStore` but I didn't like the added complexity to `GDSStore`. It is a hack, it shouldn't really be up to `GDSStore` to overwrite the value of `.zarray`.

I think moving _what-to-overwrite-in-.zarray_ to `open_cupy_array` and then only having `compressor_config_overwrite` and `decompressor_config_overwrite` arguments in `GDSStore` is a good compromise.

wence- (approved):
Thanks Mads.

PR #271, opened 2023-08-28
Unify batch and stream API check

madsbk:
We now use the `cuFileReadAsync` symbol to check for availability of both the stream and batch API of cuFile. 

We used to look for the mangled symbol `_ZTS13CUfileOpError` to determine the availability of the batch API. However, since mangling is compiler specific, we prefer using `cuFileReadAsync` even though it means that the batch API isn't available until CUDA version 12.2.

Additionally, this PR also implements `CUFILE_CHECK_STREAM_IO()`, which is used to check async IO errors.
cc. @tell-rebanta


robertmaynard (changes requested):

cpp/CMakeLists.txt
```diff
@@ -55,11 +55,7 @@ rapids_find_package(
   INSTALL_EXPORT_SET kvikio-exports
 )
 
-rapids_find_package(
-  cuFile
-  BUILD_EXPORT_SET kvikio-exports
-  INSTALL_EXPORT_SET kvikio-exports
-)
+# rapids_find_package( cuFile BUILD_EXPORT_SET kvikio-exports INSTALL_EXPORT_SET kvikio-exports )
```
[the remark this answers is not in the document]
madsbk: Good catch, this was for debugging only.

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -193,39 +177,25 @@ inline bool is_cufile_available()
...
-  }
-}
-#else
-constexpr bool is_batch_available() { return false; }
-#endif
-
-/**
- * @brief Check if cuFile's stream API is available
+ * Technically, the batch API is available in CUDA 12.1 but since there is no good
+ * way to check CUDA version using the driver API, we check for the existing of the
```
wence-, line 183: What's wrong with [`cuDriverGetVersion`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VERSION.html#group__CUDA__VERSION_1g8b7a10395392e049006e61bcdc8ebe71)? Or do you mean check the cufile version?

cpp/examples/basic_io.cpp
```diff
@@ -195,39 +197,33 @@ int main()
...
-
-      kvikio::buffer_register(a_dev, SIZE);
+      check(cudaMemcpyAsync(a_dev, a, SIZE, cudaMemcpyHostToDevice, stream) == cudaSuccess);
       f_handle.write_async(a_dev, &io_size, &f_off, &d_off, &bytes_done, stream);
+
+      // After synchronizing `stream`, we can read the number of bytes written
       check(cudaStreamSynchronize(stream) == cudaSuccess);
+      // Note, `bytes_done` might be negative, which indicate an IO error thus we
+      // use `CUFILE_CHECK_STREAM_IO` to check for errors.
+      CUFILE_CHECK_STREAM_IO(bytes_done);
```
wence-, line 216: Wow, this is a garbage error reporting api.

cpp/examples/basic_io.cpp
```diff
@@ -195,39 +197,33 @@ int main()
...
-  if (kvikio::is_stream_available()) {
     {
-      cout << "Performing stream I/O using file handle" << endl;
-      off_t f_off = 0, d_off = 0;
-      ssize_t bytes_done;
-      CUstream stream;
+      cout << "Performing async I/O using file handle" << endl;
+      off_t f_off{0};
+      off_t d_off{0};
+      ssize_t bytes_done{0};
```
wence-, line 205: I think this is incorrect, per the documentation for [`cuFileWriteAsync`](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufilewriteasync):

> The bytes_written pointer should be allocated with `cuMemHostAlloc` or registered with `cuMemHostRegister`, and the pointer to access that memory from the device can be obtained by using `cuMemHostGetDevicePointer`. 

Which we are not doing here.

cpp/CMakeLists.txt
```diff
@@ -71,8 +71,8 @@ else()
     set(cuFile_BATCH_API_FOUND TRUE)
   endif()
   message(STATUS "Found cuFile's Batch API: ${cuFile_BATCH_API_FOUND}")
-  string(FIND "${CUFILE_H_STR}" "cuFileGetVersion" cuFileGetVersion_location)
```
wence-, line 74: This function has never existed, right?

cpp/include/kvikio/error.hpp
```diff
@@ -75,8 +75,33 @@ struct CUfileException : public std::runtime_error {
...
+  (__VA_ARGS__)
+#define GET_CUFILE_CHECK_STREAM_IO_MACRO(_1, _2, NAME, ...) NAME
+#ifdef KVIKIO_CUFILE_FOUND
+#define CUFILE_CHECK_STREAM_IO_2(_nbytes_done, _exception_type)                                    \
+  do {                                                                                             \
+    int const _nbytes = (_nbytes_done);                                                            \
+    if (_nbytes < 0) {                                                                             \
+      throw(_exception_type){std::string{"cuFile error at: "} + __FILE__ + ":" +                   \
+                             KVIKIO_STRINGIFY(__LINE__) + ": " +                                   \
+                             cufileop_status_error((CUfileOpError)(CUFILEOP_BASE_ERR - _nbytes))}; \
```
wence-, line 95: So the documentation says that this returns:

> - The number of bytes successfully written. or;
> - -1 on IO errors. or;
> - All other errors will return a negative integer value of the `CUfileOpError` enum value. 

But this code suggests that it returns `-(some_cufile_op_error - CUFILEOP_BASE_ERR)`. Which can't be right given the above because it value-puns `CU_FILE_DRIVER_NOT_INITIALIZED` with an IO error.

cpp/include/kvikio/error.hpp
```diff
@@ -75,8 +75,33 @@ struct CUfileException : public std::runtime_error {
                              cufileop_status_error(error.err)};                  \
     }                                                                            \
   } while (0)
-#define CUFILE_TRY_1(_call) CUFILE_TRY_2(_call, CUfileException)
+#define CUFILE_TRY_1(_call) CUFILE_TRY_2(_call, kvikio::CUfileException)
 #endif
 #endif
 
+#ifndef CUFILE_CHECK_STREAM_IO
+#define CUFILE_CHECK_STREAM_IO(...)                                  \
```
wence-, line 83: Add a docstring/comment? While we're here. Does this need to be a macro or would a `static inline` function be fine? Then you wouldn't have worries about variable capture/etc...:

```
static inline void check_stream_io(ssize_t nbytes_written) {
   ...
}
```
Or do you want to be able to report line numbers easily?

wence- (changes requested):
I think the error handling for negative bytes written in the async write is incorrect. And, if I read the docs correctly, the pointer we're passing is also bad :(.

cpp/include/kvikio/shim/cufile.hpp
madsbk: [cuDriverGetVersion](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VERSION.html#group__CUDA__VERSION_1g8b7a10395392e049006e61bcdc8ebe71) returns the latest CUDA version supported by driver and **not** the current CUDA toolkit version :/

robertmaynard (approved):

cpp/examples/basic_io.cpp
madsbk: Good catch, fixed

cpp/include/kvikio/error.hpp
madsbk: I don't think the docs are correct. I now just print the raw value of `_nbytes`. Let's fix this in a follow-up when we know the intended behavior :)

cpp/include/kvikio/error.hpp
madsbk: I like the line numbers. In a follow-up, I will look at a more user friendly API for the deferred arguments and errors.

cpp/examples/basic_io.cpp
```diff
@@ -195,39 +195,40 @@ int main()
...
-      f_handle.read_async(c_dev, &io_size, &f_off, &d_off, &bytes_done, stream);
+      *bytes_done_p = 0;
+      f_handle.read_async(c_dev, &io_size, &f_off, &d_off, bytes_done_p, stream);
       check(cudaStreamSynchronize(stream) == cudaSuccess);
-      check(bytes_done == SIZE);
-      cout << "File stream Read : " << bytes_done << endl;
-      kvikio::buffer_deregister(c_dev);
+      CUFILE_CHECK_STREAM_IO(bytes_done_p);
+      check(*bytes_done_p == SIZE);
+      cout << "File async read : " << *bytes_done_p << endl;
```
wence-, line 229:
```suggestion
      cout << "File async read : " << *bytes_done_p << endl;
      check(cudaHostFree((void*)bytes_done_p) == cudaSuccess);
```

Let's clean up this allocation once we're done.

cpp/include/kvikio/file_handle.hpp
```diff
@@ -497,28 +497,36 @@ class FileHandle {
...
+   the
+   * actual size can be set prior to the stream I/O execution.
+   * @param file_offset Pointer to offset in the file from which to read. Unless otherwise set using
+   cuFileStreamRegister API, this value will not be evaluated until execution time.
+   * @param devPtr_offset Pointer to the offset relative to the bufPtr_base pointer from which to
+   * write. Unless otherwise set using cuFileStreamRegister API, this value will not be evaluated
+   * until execution time.
+   * @param bytes_read Pointer to the bytes read from file. This pointer should be a non-NULL value
+   * and *bytes_read set to 0. The bytes_read memory should be allocated with cuMemHostAlloc/malloc/
+   * mmap or registered with cuMemHostRegister.
```
wence-, line 522: I know you have just copied the cufile docs here, but this sentence makes no sense to me. AIUI, just plain `malloc` and `mmap` are completely unknown to the driver. So does this mean use `cuMemHostAlloc` or `cuMemAlloc` or `cuMemMap`? Or am I fully even more confused than I thought?

wence- (approved):
Some tiny fixes, LGTM!

cpp/include/kvikio/file_handle.hpp
madsbk: I guess it means that it has to be allocated on the heap (not stack)?

madsbk:
Thanks @robertmaynard and @wence-

cpp/include/kvikio/file_handle.hpp
tell-rebanta: That's correct.

PR #272, opened 2023-08-30
async: fall back to blocking

madsbk:
`read_async()` and `write_async()` now works in compat mode by falling back to the blocking implementation

### DIFF
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -506,10 +506,10 @@ class FileHandle {
    * When running CUDA v12.1 or older, this function falls back to use `.read()` after
    * `stream` has been synchronized.
    *
-   * The arguments have the same meaning as in `.read()` but some of them are deferred. That is,
-   * the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p` will not be evaluated
-   * until execution time. Notice, this behavior can be changed using cuFile's cuFileStreamRegister
-   * API.
+   * The arguments have the same meaning as in `.read()` but some of them are deferred.
+   * That is, the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p`
+   * will not be evaluated until execution time. Notice, this behavior can be changed
+   * using cuFile's cuFileStreamRegister API.
    *
    * @param devPtr_base Base address of buffer in device memory. For registered buffers,
    * `devPtr_base` must remain set to the base address used in the `buffer_register` call.
@@ -560,9 +560,10 @@ class FileHandle {
    * When running CUDA v12.1 or older, this function falls back to use `.read()` after
    * `stream` has been synchronized.
    *
-   * The arguments have the same meaning as in `.write()` but some of them are deferred. That is,
-   * the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p` will not be evaluated until execution
-   * time. Notice, this behavior can be changed using cuFile's cuFileStreamRegister API.
+   * The arguments have the same meaning as in `.write()` but some of them are deferred.
+   * That is, the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p`
+   * will not be evaluated until execution time. Notice, this behavior can be changed
+   * using cuFile's cuFileStreamRegister API.
    *
    * @param devPtr_base Base address of buffer in device memory. For registered buffers,
    * `devPtr_base` must remain set to the base address used in the `buffer_register` call.
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -507,8 +507,9 @@ class FileHandle {
    * `stream` has been synchronized.
    *
    * The arguments have the same meaning as in `.read()` but some of them are deferred. That is,
-   * the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p` will not be evaluated until execution
-   * time. Notice, this behavior can be changed using cuFile's cuFileStreamRegister API.
+   * the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p` will not be evaluated
+   * until execution time. Notice, this behavior can be changed using cuFile's cuFileStreamRegister
+   * API.
    *
    * @param devPtr_base Base address of buffer in device memory. For registered buffers,
    * `devPtr_base` must remain set to the base address used in the `buffer_register` call.
@@ -517,9 +518,9 @@ class FileHandle {
    * Later the actual size can be set prior to the stream I/O execution.
    * @param file_offset_p Pointer to offset in the file from which to read. Unless otherwise set
    * using cuFileStreamRegister API, this value will not be evaluated until execution time.
-   * @param devPtr_offset_p Pointer to the offset relative to the bufPtr_base pointer from which to
-   * write. Unless otherwise set using cuFileStreamRegister API, this value will not be evaluated
-   * until execution time.
+   * @param devPtr_offset_p Pointer to the offset relative to the bufPtr_base from which to write.
+   * Unless otherwise set using cuFileStreamRegister API, this value will not be evaluated until
+   * execution time.
    * @param bytes_read_p Pointer to the bytes read from file. This pointer should be a non-NULL
    * value and *bytes_read_p set to 0. The bytes_read_p memory should be allocated with
    * cuMemHostAlloc/malloc/mmap or registered with cuMemHostRegister. After successful execution of
@@ -538,7 +539,7 @@ class FileHandle {
                   CUstream stream)
   {
 #ifdef KVIKIO_CUFILE_STREAM_API_FOUND
-    if (kvikio::is_batch_and_stream_available() && !kvikio::defaults::compat_mode()) {
+    if (kvikio::is_batch_and_stream_available() && !_compat_mode) {
       CUFILE_TRY(cuFileAPI::instance().ReadAsync(
         _handle, devPtr_base, size_p, file_offset_p, devPtr_offset_p, bytes_read_p, stream));
       return;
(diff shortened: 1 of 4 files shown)


cpp/include/kvikio/file_handle.hpp
```diff
@@ -502,45 +503,51 @@ class FileHandle {
...
+   * When running CUDA v12.1 or older, this function falls back to use `.read()` after
+   * `stream` has been synchronized.
+   *
    * The arguments have the same meaning as in `.read()` but some of them are deferred. That is,
    * the values of `size`, `file_offset` and `devPtr_offset` will not be evaluated until execution
    * time. Notice, this behavior can be changed using cuFile's cuFileStreamRegister API.
    *
-   * @param devPtr_base Base address of buffer in device memory. For registered buffers,
+   * @param devPtr_base_p Base address of buffer in device memory. For registered buffers,
    * `devPtr_base` must remain set to the base address used in the `buffer_register` call.
```
wence-, line 514:
```suggestion
   * `devPtr_base_p` must remain set to the base address used in the `buffer_register` call.
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -502,45 +503,51 @@ class FileHandle {
...
-   * the actual size can be set prior to the stream I/O execution.
-   * @param file_offset Pointer to offset in the file from which to read. Unless otherwise set using
-   * cuFileStreamRegister API, this value will not be evaluated until execution time.
-   * @param devPtr_offset Pointer to the offset relative to the bufPtr_base pointer from which to
+   * @param size_p Pointer to size in bytes to read. If the exact size is not known at the time of
+   * I/O submission, then you must set it to the maximum possible I/O size for that stream I/O.
+   * Later the actual size can be set prior to the stream I/O execution.
+   * @param file_offset_p Pointer to offset in the file from which to read. Unless otherwise set
+   * using cuFileStreamRegister API, this value will not be evaluated until execution time.
+   * @param devPtr_offset_p Pointer to the offset relative to the bufPtr_base pointer from which to
```
wence-, line 520:
```suggestion
   * @param devPtr_offset_p Pointer to the offset relative to devPtr_base_p from which to
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -502,45 +503,51 @@ class FileHandle {
...
    * write. Unless otherwise set using cuFileStreamRegister API, this value will not be evaluated
    * until execution time.
-   * @param bytes_read Pointer to the bytes read from file. This pointer should be a non-NULL value
-   * and *bytes_read set to 0. The bytes_read memory should be allocated with cuMemHostAlloc/malloc/
-   * mmap or registered with cuMemHostRegister.
-   * After successful execution of the operation in the stream, the value *bytes_read will contain
-   * either:
+   * @param bytes_read_p Pointer to the bytes read from file. This pointer should be a non-NULL
+   * value and *bytes_read_p set to 0. The bytes_read_p memory should be allocated with
+   * cuMemHostAlloc/malloc/ mmap or registered with cuMemHostRegister. After successful execution of
```
wence-, line 525:
```suggestion
   * cuMemHostAlloc/malloc/mmap or registered with cuMemHostRegister. After successful execution of
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -502,45 +503,51 @@ class FileHandle {
    * This is an asynchronous version of `.read()`, which will be executed in sequence
    * for the specified stream.
    *
+   * When running CUDA v12.1 or older, this function falls back to use `.read()` after
+   * `stream` has been synchronized.
+   *
    * The arguments have the same meaning as in `.read()` but some of them are deferred. That is,
    * the values of `size`, `file_offset` and `devPtr_offset` will not be evaluated until execution
```
wence-, line 510:
```suggestion
   * the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p` will not be evaluated until execution
```

This is not the case in fallback mode, right?

cpp/include/kvikio/file_handle.hpp
```diff
@@ -502,45 +503,51 @@ class FileHandle {
...
-      _handle, devPtr_base, size, file_offset, devPtr_offset, bytes_read, stream));
-    return;
-#else
-    throw CUfileException("cuFile's stream API isn't available, please build with CUDA v12.2+.");
+    if (kvikio::is_batch_and_stream_available() && !kvikio::defaults::compat_mode()) {
+      CUFILE_TRY(cuFileAPI::instance().ReadAsync(
+        _handle, devPtr_base, size_p, file_offset_p, devPtr_offset_p, bytes_read_p, stream));
+      return;
+    }
 #endif
```
wence-, line 546: Perhaps prefer

```
#ifdef ...
  async_impl;
#else
  sync_impl;
#endif
```
To avoid the reader having to spot the early return in the true case? Or do you want the fallthrough to the sync case to happen even if the stream API is found but we're running in compat_mode?

cpp/include/kvikio/file_handle.hpp
```diff
@@ -549,44 +556,51 @@ class FileHandle {
...
+   * When running CUDA v12.1 or older, this function falls back to use `.read()` after
+   * `stream` has been synchronized.
+   *
    * The arguments have the same meaning as in `.write()` but some of them are deferred. That is,
    * the values of `size`, `file_offset` and `devPtr_offset` will not be evaluated until execution
    * time. Notice, this behavior can be changed using cuFile's cuFileStreamRegister API.
    *
-   * @param devPtr_base Base address of buffer in device memory. For registered buffers,
+   * @param devPtr_base_p Base address of buffer in device memory. For registered buffers,
    * `devPtr_base` must remain set to the base address used in the `buffer_register` call.
```
wence-, line 567:
```suggestion
   * `devPtr_base_p` must remain set to the base address used in the `buffer_register` call.
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -549,44 +556,51 @@ class FileHandle {
    * This is an asynchronous version of `.write()`, which will be executed in sequence
    * for the specified stream.
    *
+   * When running CUDA v12.1 or older, this function falls back to use `.read()` after
+   * `stream` has been synchronized.
+   *
    * The arguments have the same meaning as in `.write()` but some of them are deferred. That is,
    * the values of `size`, `file_offset` and `devPtr_offset` will not be evaluated until execution
```
wence-, line 563:
```suggestion
   * the values pointed to by `size_p`, `file_offset_p` and `devPtr_offset_p` will not be evaluated until execution
```
As for `read`, this is only true if we're not in fallback mode, right?

cpp/include/kvikio/file_handle.hpp
```diff
@@ -549,44 +556,51 @@ class FileHandle {
...
-   * submission, then you must set it to the maximum possible I/O size for that stream I/O. Later
-   * the actual size can be set prior to the stream I/O execution.
-   * @param file_offset Pointer to offset in the file from which to read. Unless otherwise set
+   * @param size_p Pointer to size in bytes to read. If the exact size is not known at the time of
+   * I/O submission, then you must set it to the maximum possible I/O size for that stream I/O.
+   * Later the actual size can be set prior to the stream I/O execution.
+   * @param file_offset_p Pointer to offset in the file from which to read. Unless otherwise set
    * using cuFileStreamRegister API, this value will not be evaluated until execution time.
-   * @param devPtr_offset Pointer to the offset relative to the bufPtr_base pointer from which to
+   * @param devPtr_offset_p Pointer to the offset relative to the bufPtr_base pointer from which to
```
wence-, line 573:
```suggestion
   * @param devPtr_offset_p Pointer to the offset relative to devPtr_base_p from which to
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -549,44 +556,51 @@ class FileHandle {
...
-    return;
-#else
-    throw CUfileException("cuFile's stream API isn't available, please build with CUDA v12.2+.");
+    if (kvikio::is_batch_and_stream_available() && !kvikio::defaults::compat_mode()) {
+      CUFILE_TRY(cuFileAPI::instance().WriteAsync(
+        _handle, devPtr_base, size_p, file_offset_p, devPtr_offset_p, bytes_written_p, stream));
+      return;
+    }
 #endif
+
```
wence-, line 600: As with the `read` case, put this fallback in an `#else` block? Same rationale, so if that isn't the intended behaviour, feel free to ignore in the same way.

wence- (changes requested):
Some small docstring fixes/queries.

cpp/include/kvikio/file_handle.hpp
madsbk: > Or do you want the fallthrough to the sync case to happen even if the stream API is found but we're running in compat_mode?

Yes, in compat_mode we shouldn't call cuFile

cpp/include/kvikio/file_handle.hpp
madsbk: True, in fallback mode we synchronize immediately but I guess it is a implementation detail?

wence-: Perhaps then I don't understand what it means for an argument to be "deferred" in the cufile API.

How I interpret what is going on is is something like:

```
void *devPtr_base = nullptr;
...
handle = read_async(devPtr_base, ...); // this call doesn't touch the pointer for now.
...
// some time later
*devPtr_base = some_valid_pointer;
// We are in control of when the read is actually submitted, so we can ensure that the
// pointer is set up correctly.
handle.submit(); 
```

But it seems like what you're saying, things are more like:
```
void *devPtr_base = nullptr;
...
handle = read_async(devPtr_base, ...); // this call doesn't (necessarily) touch the pointer immediately
...
// this might not have worked, because the work might already have run?
*devPtr_base = some_valid_pointer;
handle.sync();
```

Certainly the fallback "stream-sync + read" mode in cuda < 12.2 behaves more like the latter: we must arrange for all the pointers to be valid before calling `read_async` in fallback mode, because there's no chance for us to update them later.

Or do I have the wrong end of the stick?

madsbk: I think the goal of the deferred arguments in cuFile is the ability to feed the output of one async cuda function into the next async function without sync.
```cpp
ssize_t *bytes_read = malloc(sizeof(ssize_t));
read_async(..., bytes_read, stream);
cudaMemCpyAsync(..., *bytes_read, stream);
write_async(..., size_p=bytes_read, stream);
```

wence-: Ah OK, so it's not "set up a plan and fill in the details later". In which case, I withdraw my worries.

madsbk:
Thanks @wence-, anything else?

wence- (approved):
All good!

PR #275, opened 2023-09-04
Async I/O using by-value arguments

madsbk:
Implements a version of `read_async` and `write_async` that takes by-value arguments and returns a `StreamFuture`. 

This is to make the common case easy where the user knows the file and buffer size.

cc. @tell-rebanta


cpp/include/kvikio/stream.hpp
```diff
@@ -0,0 +1,136 @@
...
+   * @brief Return the number of bytes precessed by the future operation.
+   *
+   * Synchronize the associated CUDA stream.
+   *
+   * @return Number of bytes read or written by the future operation.
+   */
+  std::size_t check_bytes_done() const
+  {
+    CUDA_DRIVER_TRY(cudaAPI::instance().StreamSynchronize(_stream));
+    CUFILE_CHECK_STREAM_IO(_ref.bytes_done_p);
```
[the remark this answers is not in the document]
madsbk: I am not sure I follow, `bytes_done_p` should point to `bytes_done` so they should aways be sync?

To make the relationship more clear, I have added:
```cpp
/* 
 * The goal of this class is twofold:
 *   - Have `read_async` and `write_async` return an object that clearly associates the function
 *     arguments with the CUDA stream used. This is useful because the current validity of the
 *     arguments depends on the stream.
 *   - Support of by-value arguments. In many cases, a user will use `read_async` and `write_async`
 *     like most other asynchronous CUDA functions that take by-value arguments.
 *
 * To support by-value arguments, we allocate the arguments on the heap (malloc `ArgByVal`) and have
 * the by-reference arguments (`ArgByRef`) points into `ArgByVal`. This way, the `read_async` and
 * `write_async` can call `.get_args()` to get the by-reference arguments required by cuFile's
 * stream API. However, this also means that the caller of `read_async` and `write_async` MUST keep
 * the returned `StreamFuture` alive until the operations is done otherwise `StreamFuture` will free
 * `ArgByVal` before cuFile had a change to use them!
 */
```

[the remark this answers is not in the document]
tell-rebanta: Got it. Thanks for the explanation.

tell-rebanta (approved):

cpp/include/kvikio/stream.hpp
```diff
@@ -0,0 +1,150 @@
...
+ *     arguments depends on the stream.
+ *   - Support of by-value arguments. In many cases, a user will use `read_async` and `write_async`
+ *     like most other asynchronous CUDA functions that take by-value arguments.
+ *
+ * To support by-value arguments, we allocate the arguments on the heap (malloc `ArgByVal`) and have
+ * the by-reference arguments (`ArgByRef`) points into `ArgByVal`. This way, the `read_async` and
+ * `write_async` can call `.get_args()` to get the by-reference arguments required by cuFile's
+ * stream API. However, this also means that the caller of `read_async` and `write_async` MUST keep
+ * the returned `StreamFuture` alive until the operations is done otherwise `StreamFuture` will free
+ * `ArgByVal` before cuFile had a change to use them!
```
wence-, line 48: **💡 suggestion:** Could we make this safer by having `~StreamFuture` sync the stored stream?

cpp/include/kvikio/stream.hpp
```diff
@@ -0,0 +1,150 @@
...
+    off_t file_offset;
+    off_t devPtr_offset;
+    ssize_t bytes_done;
+  };
+  struct ArgByRef {
+    std::size_t* size_p;
+    off_t* file_offset_p;
+    off_t* devPtr_offset_p;
+    ssize_t* bytes_done_p;
+  };
```
wence-, line 63: **❓ question:** ‏Since these are not exposed to the outside world (I think/hope), should they rather be `private`?

cpp/include/kvikio/stream.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+ private:
+  void* _devPtr_base{nullptr};
+  CUstream _stream{nullptr};
+  ArgByVal* _val{nullptr};
+  ArgByRef _ref{nullptr};
```
wence-, line 69: **❓ question:** ‏Do we actually need both of these stored in the class, does it suffice instead to do something like:

```c++
class StreamFuture {
  private:
     struct data {
        std::size_t size;
        off_t file_offset;
        off_t devptr_offset;
        ssize_t bytes_done;
    }
    data* _values{};
  public:
    StreamFuture(...args) 
    {
        _values = static_cast<data *>(...);
        _values->size = size;
        _values->...;
     }
     std::tuple<...> get_args() { 
        return {&_values->size, &_values->file_offset, ...};
    }
}
```
Or is there some subtlety I am missing?

wence- (changes requested):
Thanks @madsbk, wondering if we can simplify things a little, and make the interface slightly safer to use (and avoid use-after-free footguns one normally has with raw pointers).

cpp/include/kvikio/stream.hpp
madsbk: fixed

cpp/include/kvikio/stream.hpp
madsbk: Good point, done

madsbk:
@wence- I think I have addressed all your comments.

cpp/include/kvikio/stream.hpp
```diff
@@ -0,0 +1,158 @@
...
+   */
+  ~StreamFuture() noexcept
+  {
+    if (_val != nullptr) {
+      try {
+        check_bytes_done();
+      } catch (const kvikio::CUfileException& e) {
+        std::cerr << e.what() << std::endl;
+      }
+      free(_val);
```
wence-, line 153: **nitpick (non-blocking):** ‏
```suggestion
      std::free(_val);
```
?

wence- (approved):
Thanks @madsbk, apologies for the delay

PR #276, opened 2023-09-05
Zarr Compressor Benchmark

madsbk:


python/benchmarks/zarr-compressors.py
```diff
@@ -0,0 +1,187 @@
...
+def run(args, compressor: Compressor):
+    src = create_src_data(args, compressor)
+    dst = numpy.empty_like(src)  # Notice, if src is a cupy array dst is as well
+    t0 = clock()
+    a = compressor.codec.encode(src)
+    encode_time = clock() - t0
+
+    t0 = clock()
+    compressor.codec.decode(a, out=dst)
+    decode_time = clock() - t0
```
akshaysubr, line 62: This probably needs some GPU synchronization before measuring end time, doesn't it?

python/benchmarks/zarr-compressors.py
```diff
@@ -0,0 +1,187 @@
...
+@dataclass
+class Compressor:
+    device: Device
+    codec: Codec
+
+
+compressors = {
+    "lz4-default": Compressor(device=Device.CPU, codec=numcodecs.LZ4()),
+    "lz4-blosc": Compressor(
+        device=Device.CPU, codec=numcodecs.blosc.Blosc(cname="lz4")
```
akshaysubr, line 35: The default for Blosc is `shuffle=SHUFFLE` which changes the compression ratio a fair bit compared to LZ4. Might be good to indicate that in the str naming?

python/benchmarks/zarr-compressors.py
```diff
@@ -0,0 +1,187 @@
...
+    ),
+    "lz4-nvcomp": Compressor(device=Device.GPU, codec=kvikio.zarr.LZ4()),
+    "snappy-nvcomp": Compressor(device=Device.GPU, codec=kvikio.zarr.Snappy()),
+    "cascaded-nvcomp": Compressor(device=Device.GPU, codec=kvikio.zarr.Cascaded()),
+    "gdeflate-nvcomp": Compressor(device=Device.GPU, codec=kvikio.zarr.Gdeflate()),
+    "bitcomp-nvcomp": Compressor(device=Device.GPU, codec=kvikio.zarr.Bitcomp()),
```
akshaysubr, line 41: Bitcomp requires the dtype which currently isn't an option here but is an option in `BitcompManager`. Might be good to expose that and add it here if possible.

akshaysubr (commented):
Wanted to chime in with a couple of minor comments that I had as I was going through this PR

PR #277, opened 2023-09-07
Update doxygen to 1.9.1

vyasr:
I selected this version as it is what ships with Ubuntu 22.04. I also ran doxygen -u to update the Doxyfile.


raydouglass (approved):

madsbk (approved):
Thanks @vyasr

PR #280, opened 2023-09-18
Update to clang 16.0.6.

bdice:
This PR updates kvikio to use clang 16.0.6. The previous version has some minor formatting issues affecting several RAPIDS repos.


madsbk (approved):
Thanks @bdice

PR #290, opened 2023-09-28
Use short tag in devcontainer versions.

bdice:
This fixes an issue with the devcontainers workflow referencing containers with the incorrect versions.


bdice:
Replacing this with #292. This fix is needed to unblock CI.

PR #293, opened 2023-09-28
Improve performance of nvCOMP batch codec.

Alexey-Kamenev:
This PR contains a set of performance-related improvements for the batch nvCOMP codec.
The short summary of the changes:
* Replaced multiple calls to CUDA  `memcpyAsync` with a single call to a CUDA kernel.
* Removed redundant memory allocations and copies (some of them are the result of the previous change).
* Vectorized loops, removed redundant loops.

As a result, decompression throughput on ERA5 data increased from **4** GB/s to **31** GB/s for LZ4 algorithm. For highly-compressible data from [nvCOMP benchmark](https://github.com/NVIDIA/nvcomp/blob/main/doc/Benchmarks.md), the increase is even higher: from **6** GB/s to about **110** GB/s. Other algorithms, such as GDeflate, show performance improvements as well.

Compression throughput was also improved, though the main target was decompression (compress once - decompress many kind of scenario).

Limitations: 
* these improvements are available only when directly using the codec's batch methods, such as `decode_batch` while passing reasonably-sized batches to saturate the GPU. That means these changes will not be available (for now) to `zarr` users as `zarr` serializes the calls into a sequence of `decode` calls.
* to get maximum performance, users should use equal-sized chunks (this is the default behavior in most of the cases anyway, such as `zarr`).

### DIFF
--- a/python/kvikio/_lib/libnvcomp_ll.pyx
+++ b/python/kvikio/_lib/libnvcomp_ll.pyx
@@ -399,7 +399,7 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
         self._set_chunk_size_header_kernel = cupy.ElementwiseKernel(
             "uint64 uncomp_chunk_size",
             "uint64 comp_chunk_ptr",
-            "((unsigned int *)comp_chunk_ptr)[0] = uncomp_chunk_size",
+            "((unsigned int *)comp_chunk_ptr)[0] = (unsigned int)uncomp_chunk_size",
             "set_chunk_size_header",
             no_return=True,
         )
--- a/python/kvikio/_lib/libnvcomp_ll.pyx
+++ b/python/kvikio/_lib/libnvcomp_ll.pyx
@@ -387,6 +387,14 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
         self.options = nvcompBatchedLZ4Opts_t(data_type)
         self.has_header = has_header
 
+        # Note on LZ4 header structure: numcodecs LZ4 codec prepends
+        # a 4-byte (uint32_t) header to each compressed chunk.
+        # The header stores the size of the original (uncompressed) data:
+        # https://github.com/zarr-developers/numcodecs/blob/cb155432e36536e17a2d054c8c24b7bf6f4a7347/numcodecs/lz4.pyx#L89
+        #
+        # The following CUDA kernels read / write chunk header by
+        # casting the chunk pointer to a pointer to unsigned int.
+
         # CUDA kernel that copies uncompressed chunk size from the chunk header.
         self._get_size_from_header_kernel = cupy.ElementwiseKernel(
             "uint64 comp_chunk_ptr",
--- a/python/kvikio/_lib/libnvcomp_ll.pyx
+++ b/python/kvikio/_lib/libnvcomp_ll.pyx
@@ -143,19 +143,14 @@ class nvCompBatchAlgorithm(ABC):
             The number of chunks to compress.
         temp_buf: cp.ndarray
             The temporary GPU workspace.
-        comp_chunks: np.ndarray[uintp]
+        comp_chunks: cp.ndarray[uintp]
             (output) The list of pointers on the GPU, to the output location for each
             compressed batch item.
-        comp_chunk_sizes: np.ndarray[uint64]
+        comp_chunk_sizes: cp.ndarray[uint64]
             (output) The compressed size in bytes of each chunk.
         stream: cp.cuda.Stream
             CUDA stream.
         """
-
-        # nvCOMP requires comp_chunks pointers container and
-        # comp_chunk_sizes to be in GPU memory.
-        comp_chunks_d = cupy.array(comp_chunks, dtype=cupy.uintp)
-        comp_chunk_sizes_d = cupy.empty_like(comp_chunk_sizes)
 
         err = self._compress(
             uncomp_chunks,
@@ -163,14 +158,12 @@ class nvCompBatchAlgorithm(ABC):
             max_uncomp_chunk_bytes,
             batch_size,
             temp_buf,
-            comp_chunks_d,
-            comp_chunk_sizes_d,
+            comp_chunks,
+            comp_chunk_sizes,
             stream,
         )
         if err != nvcompStatus_t.nvcompSuccess:
             raise RuntimeError(f"Compression failed, error: {nvCompStatus(err)!r}.")
-        # Copy resulting compressed chunk sizes back to the host buffer.
-        comp_chunk_sizes[:] = comp_chunk_sizes_d.get()
 
     @abstractmethod
     def _compress(
@@ -238,9 +231,9 @@ class nvCompBatchAlgorithm(ABC):
 
         Parameters
         ----------
-        comp_chunks: np.ndarray[uintp]
+        comp_chunks: cp.ndarray[uintp]
             The pointers on the GPU, to compressed batched items.
-        comp_chunk_sizes: np.ndarray[uint64]
+        comp_chunk_sizes: cp.ndarray[uint64]
             The size in bytes of each compressed batch item.
         stream: cp.cuda.Stream
             CUDA stream.
@@ -256,15 +249,13 @@ class nvCompBatchAlgorithm(ABC):
         batch_size = len(comp_chunks)
 
         # nvCOMP requires all buffers to be in GPU memory.
-        comp_chunks_d = cupy.array(comp_chunks, dtype=cupy.uintp)
-        comp_chunk_sizes_d = cupy.array(comp_chunk_sizes, dtype=cupy.uint64)
-        uncomp_chunk_sizes_d = cupy.empty_like(comp_chunk_sizes_d)
+        uncomp_chunk_sizes = cupy.empty_like(comp_chunk_sizes)
 
         err = self._get_decomp_size(
-            comp_chunks_d,
-            comp_chunk_sizes_d,
+            comp_chunks,
+            comp_chunk_sizes,
             batch_size,
-            uncomp_chunk_sizes_d,
+            uncomp_chunk_sizes,
             stream,
         )
         if err != nvcompStatus_t.nvcompSuccess:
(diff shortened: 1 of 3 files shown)


python/kvikio/nvcomp_codec.py
```diff
@@ -66,18 +66,18 @@ def encode(self, buf):
...
         return self.encode_batch([buf])[0]
 
-    def encode_batch(self, bufs: List[Any]) -> List[Any]:
+    def encode_batch(self, bufs: Sequence[Any]) -> Sequence[Any]:
         """Encode data in `bufs` using nvCOMP.
 
         Parameters
         ----------
-        bufs : List[buffer-like].
+        bufs : Sequence[buffer-like].
```
madsbk, line 74:
```suggestion
        bufs
```

python/kvikio/nvcomp_codec.py
```diff
@@ -66,18 +66,18 @@ def encode(self, buf):
...
             Data to be encoded. Each buffer in the list may be any object
             supporting the new-style buffer protocol.
 
         Returns
         -------
-        enc : List[buffer-like]
+        enc : Sequence[buffer-like]
             List of encoded buffers. Each buffer may be any object supporting
             the new-style buffer protocol.
```
madsbk, line 82:
```suggestion
            List of encoded buffers. Each buffer may be any object supporting
            the new-style buffer protocol.
```

python/kvikio/nvcomp_codec.py
```diff
@@ -145,23 +150,23 @@ def decode(self, buf, out=None):
...
-        self, bufs: List[Any], out: Optional[List[Any]] = None
-    ) -> List[Any]:
+        self, bufs: Sequence[Any], out: Optional[Sequence[Any]] = None
+    ) -> Sequence[Any]:
         """Decode data in `bufs` using nvCOMP.
 
         Parameters
         ----------
-        bufs : List[buffer-like]
+        bufs : Sequence[buffer-like]
```
madsbk, line 159:
```suggestion
        bufs
```

python/kvikio/nvcomp_codec.py
```diff
@@ -145,23 +150,23 @@ def decode(self, buf, out=None):
...
             Encoded data. Each buffer in the list may be any object
             supporting the new-style buffer protocol.
-        out : List[buffer-like], optional
+        out : Sequence[buffer-like], optional
```
madsbk, line 162:
```suggestion
        out
```

python/kvikio/nvcomp_codec.py
```diff
@@ -145,23 +150,23 @@ def decode(self, buf, out=None):
...
             List of writeable buffers to store decoded data.
             N.B. if provided, each buffer must be exactly the right size
             to store the decoded data.
 
         Returns
         -------
-        dec : List[buffer-like]
+        dec : Sequence[buffer-like]
             List of decoded buffers. Each buffer may be any object supporting
             the new-style buffer protocol.
```
madsbk, line 171:
```suggestion
            List of decoded buffers. Each buffer may be any object supporting
            the new-style buffer protocol.
```

madsbk (commented):
Some nitpick - no need to have the argument types in the doc when using type hints

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Just a quick note that I've tried to follow the style from `numcodecs.abc.Codec` even though I think it's too verbose (i.e. I agree with your comment).

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Done.

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Done.

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Done.

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Done.

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Done.

madsbk (approved):
Looks good, thanks @Alexey-Kamenev

python/kvikio/nvcomp_codec.py
```diff
@@ -176,34 +179,36 @@ def decode_batch(
...
-            cp.empty(size, dtype=cp.uint8) for size in uncomp_chunk_sizes_h
-        ]
-        uncomp_chunk_ptrs = cp.array(
-            [c.data.ptr for c in uncomp_chunks], dtype=cp.uintp
+        # For performance reasons, we use max_chunk_size so we can create
+        # a rectangular array with the same pointer increments.
+        uncomp_chunks = cp.empty((num_chunks, max_chunk_size), dtype=cp.uint8)
+        p_start = uncomp_chunks.data.ptr
+        uncomp_chunk_ptrs = p_start + (
+            cp.arange(0, num_chunks * max_chunk_size, max_chunk_size)
```
wence-, line 211: **nitpick (non-blocking):** ‏Maybe explicitly set the dtype here:
```suggestion
        uncomp_chunk_ptrs = cupy.uint64(p_start) + (
            cp.arange(0, num_chunks * max_chunk_size, max_chunk_size, dtype=cupy.uint64)
```
Rationale: I am always worried about mixed-kind casting (there are some circumstances, I think not here because `p_start` is just a python integer), but notice that `cupy.arange(10, dtype="uint64") + cupy.arange(10, dtype="int64")` has dtype `float64`!

python/kvikio/nvcomp_codec.py
```diff
@@ -226,20 +231,23 @@ def decode_batch(
...
 
+        # If all chunks are the same size, we can just return uncomp_chunks.
+        if is_equal_chunks and out is None:
+            return cp.asnumpy(uncomp_chunks) if is_host_buffer else uncomp_chunks
+
         res = []
         for i in range(num_chunks):
-            ret = uncomp_chunks[i]
-            if out is not None and out[i] is not None:
+            ret = uncomp_chunks[i, : uncomp_chunk_sizes[i].item()]
```
wence-, line 240: **polish:** ‏Avoid excessive d2h synchronisation by copying `uncomp_chunk_sizes` to the host in one go before this loop?

python/kvikio/nvcomp_codec.py
```diff
@@ -226,20 +231,23 @@ def decode_batch(
...
+            if out is None or out[i] is None:
+                res.append(cp.asnumpy(ret) if is_host_buffer else ret)
+            else:
                 o = ensure_contiguous_ndarray_like(out[i])
                 if hasattr(o, "__cuda_array_interface__"):
                     cp.copyto(o, ret.view(dtype=o.dtype), casting="no")
                 else:
+                    # TODO(akamenev): remove redundant H2H copy?
                     np.copyto(o, cp.asnumpy(ret.view(dtype=o.dtype)), casting="no")
```
wence-, line 249: **polish:** ‏addressing todo
```suggestion
                    cp.asnumpy(ret.view(dtype=o.dtype), out=o, stream=self._stream)
```

With a `self._stream.synchronize()` at the end of the loop.

wence- (commented):
Thanks, some tiny opportunities for cleanups, looks great!

madsbk:
@Alexey-Kamenev Notice, I took the liberty to run `pre-commit run --all-file` to get CI running

python/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -407,6 +387,23 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
         self.options = nvcompBatchedLZ4Opts_t(data_type)
         self.has_header = has_header
 
+        # CUDA kernel that copies uncompressed chunk size from the chunk header.
+        self._get_size_from_header_kernel = cupy.ElementwiseKernel(
+            "uint64 comp_chunk_ptr",
+            "uint64 uncomp_chunk_size",
+            "uncomp_chunk_size = *((unsigned int *)comp_chunk_ptr)",
```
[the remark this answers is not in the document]
Alexey-Kamenev: I have to use 32-bit pointer because that's how the header data is stored in [numcodecs LZ4 codec](https://github.com/zarr-developers/numcodecs/blob/cb155432e36536e17a2d054c8c24b7bf6f4a7347/numcodecs/lz4.pyx#L89). The header is just a `uint32_t` field that stores original, uncompressed data size. Unfortunately, that makes the header structure essentially frozen - there is no versioning or metadata info which would allow extending the header in the future, if needed.
I agree it may confuse readers, so I'm open to any suggestions to make this code more readable (add a comment?). Agree that ideally should also use fixed width `uint32_t` type instead of `unsigned int` (integer types are confusing in C++ standard), but, as you mentioned, it's not currently possible in cupy.

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Great catch - fixed! The code worked because I got lucky :) with `p_start` being an integer, as you mentioned.

Alexey-Kamenev: Also, in case of adding a scalar, the resulting type remains `uint64`:
```
assert (p_start + cp.arange(0, num_chunks * max_chunk_size, max_chunk_size, dtype=cp.uint64)).dtype == cp.uint64
```

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Done.

python/kvikio/nvcomp_codec.py
Alexey-Kamenev: Done.

python/kvikio/_lib/libnvcomp_ll.pyx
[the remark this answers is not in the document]
wence-: Thanks! I think I was confused because I was trying to find the allocation of `comp_chunk_ptr` and I _thought_ it was an allocation uint64 entries, but I probably looked in the wrong place. So, 

**suggestion:** ‏Perhaps add a comment here indicating which types these have to match up with?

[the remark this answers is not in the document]
Alexey-Kamenev: Done.

wence- (approved):
I think this loks good, thanks!

Issue #297, opened 2023-10-04
Error when loading Zarr store with LZ4 compression using numcodecs or nvCOMP


weiji14:
Hi there,

I've been trying out the new `kvikio.zarr.open_cupy_array` feature at https://docs.rapids.ai/api/kvikio/23.10/zarr (added in #267), and have been encountering some issues with the LZ4 decoding. A little unsure if this belongs here in kvikIO or on [numcodecs](https://github.com/zarr-developers/numcodecs), but I'll post it here first.

Minimal example using CPU LZ4 decoder:

```python
import cupy as cp
import zarr
import kvikio.zarr

za = zarr.create(
    shape=(10,), fill_value=1, meta_array=cp.empty(shape=()), compressor=kvikio.zarr.LZ4()
)
zarr.save("rand10.zarr", cp.asnumpy(za[:]), zarr_version=2)

zg = kvikio.zarr.open_cupy_array(
    store="rand10.zarr", compressor=kvikio.zarr.CompatCompressor.lz4(), mode="r", path="/"
)
print(zg.compressor)  # Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
zg[:]
```

yields this error:

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[8], line 1
----> 1 zg[:]

File ~/mambaforge/envs/kvikio-env/lib/python3.10/site-packages/zarr/core.py:844, in Array.__getitem__(self, selection)
    842     result = self.get_orthogonal_selection(pure_selection, fields=fields)
    843 else:
--> 844     result = self.get_basic_selection(pure_selection, fields=fields)
    845 return result

File ~/mambaforge/envs/kvikio-env/lib/python3.10/site-packages/zarr/core.py:970, in Array.get_basic_selection(self, selection, out, fields)
    968     return self._get_basic_selection_zd(selection=selection, out=out, fields=fields)
    969 else:
--> 970     return self._get_basic_selection_nd(selection=selection, out=out, fields=fields)

File ~/mambaforge/envs/kvikio-env/lib/python3.10/site-packages/zarr/core.py:1012, in Array._get_basic_selection_nd(self, selection, out, fields)
   1006 def _get_basic_selection_nd(self, selection, out=None, fields=None):
   1007     # implementation of basic selection for array with at least one dimension
   1008 
   1009     # setup indexer
   1010     indexer = BasicIndexer(selection, self)
[...]

PR #298, opened 2023-10-05
Unify the CUDA Codecs

madsbk:
It this PR we introduce `CudaCodec`, which is a base class for all CUDA Condecs/Compressors.
This makes it possible to detect if an user tries to open a Zarr file using an incompatible compressor (see https://github.com/rapidsai/kvikio/issues/297). 

Additionally, `kvikio.zarr.open_cupy_array()` now handles `mode="a"`

### DIFF
--- a/python/kvikio/numcodecs.py
+++ b/python/kvikio/numcodecs.py
@@ -1,0 +1,60 @@
+# Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
+
+"""
+This module implements CUDA compression and transformation codecs for Numcodecs.
+See <https://numcodecs.readthedocs.io/en/stable/>
+"""
+
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import Any, Optional, Union
+
+import cupy.typing
+import numpy.typing
+from numcodecs.abc import Codec
+
+# TODO: replace `ANY` with `collections.abc.Buffer` from PEP-688
+# when it becomes available.
+BufferLike = Union[cupy.typing.NDArray, numpy.typing.ArrayLike, Any]
+
+
+class CudaCodec(Codec):
+    """Abstract base class for CUDA codecs"""
+
+    @abstractmethod
+    def encode(self, buf: BufferLike) -> cupy.typing.NDArray:
+        """Encode `buf` using CUDA.
+
+        Parameters
+        ----------
+        buf
+            A numpy array like object such as numpy.ndarray, cupy.ndarray,
+            or any object exporting a buffer interface.
+
+        Returns
+        -------
+        The compressed buffer wrapped in a CuPy array
+        """
+
+    @abstractmethod
+    def decode(self, buf: BufferLike, out: Optional[BufferLike] = None) -> BufferLike:
+        """Encode `buf` using CUDA.
+
+        Parameters
+        ----------
+        buf
+            A numpy array like object such as numpy.ndarray, cupy.ndarray,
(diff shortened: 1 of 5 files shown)


python/kvikio/zarr.py
```diff
@@ -171,7 +175,7 @@ def getitems(
         return ret
 
 
-class NVCompCompressor(Codec):
+class NVCompCompressor(CudaCodec):
     """Abstract base class for nvCOMP compressors
```
wence-, line 179: **nitpick:** Is this really an abstract base class now?

python/kvikio/zarr.py
```diff
@@ -359,29 +333,44 @@ def open_cupy_array(
...
+                )
+            elif not isinstance(ret.compressor, CudaCodec):
+                raise ValueError(
+                    "The Zarr file was written using a non-CUDA compatible "
+                    f"compressor, {ret.compressor}, please use something "
+                    "like kvikio.zarr.CompatCompressor"
+                )
+            return ret
+
+    # At this point, we known that we are writing a new array ("w", "w-", "a")
```
wence-, line 373: **suggestion:** The type-checker only complains about an invalid mode argument if it can deduce a type for the argument that is not `Any`. So it is possible to pass an invalid value and fall through here without having a valid mode argument. Should we explicitly assert that the mode is in the set of valid modes?

python/tests/test_zarr.py
```diff
@@ -212,11 +212,13 @@ def test_compressor_config_overwrite(tmp_path, xp, algo):
     numpy.testing.assert_array_equal(z[:], range(10))
 
 
-def test_open_cupy_array(tmp_path):
+@pytest.mark.parametrize("write_mode", ["w", "w-", "a"])
+@pytest.mark.parametrize("read_mode", ["r", "r+", "a"])
+def test_open_cupy_array(tmp_path, write_mode, read_mode):
```
wence-, line 217: **polish:** ‏Test invalid modes and assert that an appropriate error is raised

wence- (changes requested):
Apologies for the delay @madsbk, just a few small suggestions

python/kvikio/zarr.py
madsbk: good point, done

python/tests/test_zarr.py
madsbk: Done

python/kvikio/numcodecs.py
```diff
@@ -0,0 +1,64 @@
...
+
+        Parameters
+        ----------
+        buf
+            A numpy array like object such as numpy.ndarray, cupy.ndarray,
+            or any object exporting a buffer interface.
+        out
+            A numpy array like object such as numpy.ndarray, cupy.ndarray,
+            or any object exporting a buffer interface. If provided, this buffer must
+            be exactly the right size to store the decoded data.
```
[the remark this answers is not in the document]
madsbk: > Does this raise if the size is not correct? What happens if the provided buffer is bigger than required?

It is up to the implementer to check for incorrect sizes. Our two nvCOMP codecs do this through CuPy/NumPy coping.

python/kvikio/zarr.py
wence-: Ah, missed that, thanks.

python/kvikio/numcodecs.py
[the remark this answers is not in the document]
wence-: OK, thanks.

wence- (approved):
Thanks Mads

PR #301, opened 2023-10-10
Update rapids-cmake functions to non-deprecated signatures

robertmaynard:
Update to use non deprecated signatures for `rapids_export` functions


bdice (approved):

robertmaynard:
@bdice @madsbk Any reason why I can't merge this? We will need this to unblock 23.04 builds

bdice:
I agree with your assessment and went ahead and merged.

PR #321, opened 2023-11-25
Add timer to basic_io example

yncxcw:
Add time to run some benchmark locally

### DIFF
--- a/cpp/examples/basic_io.cpp
+++ b/cpp/examples/basic_io.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <chrono>
 #include <iostream>
 
 #include <cuda_runtime_api.h>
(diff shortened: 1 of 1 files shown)


cpp/examples/basic_io.cpp
```diff
@@ -27,6 +28,27 @@
...
+  ~Timer()
+  {
+    auto end = std::chrono::high_resolution_clock::now();
+    auto start_ms =
+      std::chrono::time_point_cast<std::chrono::microseconds>(start).time_since_epoch().count();
+    auto end_ms =
+      std::chrono::time_point_cast<std::chrono::microseconds>(end).time_since_epoch().count();
+
+    std::cout << "Benchmark " << name << " takes: " << end_ms - start_ms << " microseconds"
+              << std::endl;
```
madsbk, line 44: What about a less verbose version like:
```suggestion
    std::cout << "(" << end_ms - start_ms << " us)" << std::endl;
```

and remove the `name` argument?

madsbk (commented):
Thanks @yncxcw, looks good. I only have a minor suggestion.

cpp/examples/basic_io.cpp
yncxcw: Sure, done

madsbk (approved):
Thanks @yncxcw, looks good

PR #328, opened 2024-01-11
Support CUDA 12.2

jameslamb:
## Description

* switches to CUDA 12.2.2 for building conda packages and wheels
* adds new tests running against CUDA 12.2.2

### Notes for Reviewers

This is part of ongoing work to build and test packages against CUDA 12.2.2 across all of RAPIDS.

For more details see:

* https://github.com/rapidsai/build-planning/issues/7
* https://github.com/rapidsai/shared-workflows/pull/166

Planning a second round of PRs to revert these references back to a proper `branch-24.{nn}` release branch of `shared-workflows` once https://github.com/rapidsai/shared-workflows/pull/166 is merged.

*(created with `rapids-reviser`)*


vyasr (changes requested):
[I'm seeing cufile being pulled in arm 12.0 testing](https://github.com/rapidsai/kvikio/actions/runs/7802997515/job/21282060836?pr=328#step:7:291). I think we want to disable that, right?

bdice:
The aarch64 / libcufile part needed a bit more work. I took a stronger approach in f45ac7b, which removes `libcufile-dev` from aarch64 completely. I don't think we have a very good way to support "only aarch64 and CUDA 12.2+" in our build/test workflows yet, because we only build once with CEC enabled and we would need two aarch64 builds to support CUDA 12.0/12.1 without libcufile and 12.2+ with libcufile. I think it's best to wait until a later time to handle this.

AyodeAwe (approved):

vyasr (commented):
I'm trying to think through exactly what cases are OK here and which aren't. kvikio dlopens libcufile, so in principle we ought to be able to put the runtime protections in place to avoid trying to load libcufile even if it is installed into an environment where it is not supported (CUDA 12.0, aarch64). We don't have to solve that in this PR (we can merge 12.2 support without aarch64 libcufile support) but I do think we should be able to enable this in a follow-up without waiting to drop CUDA 12.0/12.1. We can just allow installation of libcufile but make sure the library isn't loaded at runtime.

If we're OK with that result, can we just open an issue to track that as future work?

vyasr (approved):
Sorry meant to approve based on my previous comment.

bdice:
Great. You're right, that sounds plausible. I hadn't considered that this could be a runtime guard because it's dlopen'd. https://github.com/rapidsai/kvikio/issues/339

Merging this for now.

PR #330, opened 2024-01-16
Use libcufile in aarch64 CUDA 12 builds.

bdice:
CUDA 12.2 has conda-forge packages of `libcufile` and `libcufile-dev`. This PR updates `dependencies.yaml` so that environments can be generated on ARM.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -188,7 +188,6 @@ dependencies:
         matrices:
           - matrix:
               cuda: "12.*"
-              arch: x86_64
             packages:
               - libcufile
               - libcufile-dev


jakirkham:
Thanks Bradley! 🙏

Do we want to update lines like these as well?

https://github.com/rapidsai/kvikio/blob/f8f58581224082cbec98fb00a7b224abe98d3381/conda/recipes/libkvikio/meta.yaml#L49-L55

bdice:
@jakirkham Good question. I think I would prefer to put those recipe changes in #328, since we aren't enabling CUDA 12.2 workflows yet. Maybe this PR should be closed and ported into #328...?

bdice:
I'm closing this in favor of merging changes to #328, which will include the conda recipe changes for libcufile.

jakirkham:
No strong feelings either way

@jameslamb what do you think?

jameslamb:
ha no prob, yeah looks like @bdice pulled them into #328 already

PR #342, opened 2024-02-13
Defer cuFile feature checks until finding kvikio package

### LINKED ISSUE #341: libkvikio conda packages are built with INTERFACE_COMPILE_DEFINITIONS that require CUDA 12.2 for cuFile Stream APIs
This issue describes a build failure seen in cudf's devcontainers with CUDA 12.0 when using the libkvikio conda packages that are now built with CUDA 12.2, as of #328. The build failures look like this:
```
  In file included from /home/coder/.conda/envs/rapids/include/kvikio/defaults.hpp:26,
                   from /home/coder/.conda/envs/rapids/include/kvikio/buffer.hpp:23,
                   from /home/coder/.conda/envs/rapids/include/kvikio/file_handle.hpp:31,
                   from /home/coder/cudf/cpp/src/io/utilities/datasource.cpp:26:
  /home/coder/.conda/envs/rapids/include/kvikio/shim/cufile.hpp:59:12: error: 'cuFileReadAsync' was not declared in this scope; did you mean 'cuFileRead'?
     59 |   decltype(cuFileReadAsync)* ReadAsync{nullptr};
        |            ^~~~~~~~~~~~~~~
        |            cuFileRead
```

The libkvikio conda packages contain this in `lib/cmake/kvikio/kvikio-targets.cmake`:

```
set_target_properties(kvikio::kvikio PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "KVIKIO_CUFILE_FOUND;KVIKIO_CUFILE_BATCH_API_FOUND;KVIKIO_CUFILE_STREAM_API_FOUND"
  INTERFACE_COMPILE_FEATURES "cxx_std_17"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Threads::Threads;CUDA::toolkit;cufile::cuFile_interface;dl"
)
```

In other words, the `INTERFACE_COMPILE_DEFINITIONS` are being set based on the packages available during the libkvikio conda build, which might not be the same packages/versions as when libkvikio is actually being used. Because we built libkvikio with CUDA 12.2 and then tried to use it with CUDA 12.0 devcontainers, the build failed to find the cuFile Stream APIs that were introduced in CUDA 12.2.

I am opening a PR that proposes to defer these definitions until the call to `find_package`, which will use the exact cuFile features present (if cuFile is available at all) when building a package like cudf that depends on kvikio.

bdice:
This PR closes #341.

The kvikio `INTERFACE_COMPILE_DEFINITIONS` were being set based on the packages available during the libkvikio conda build (e.g. CUDA 12.2 since #328), which might not be the same packages/versions as when libkvikio is actually being used (e.g. to build libcudf with CUDA 12.0). Because we built libkvikio with CUDA 12.2 and then tried to use it with CUDA 12.0 devcontainers, the build failed to find the cuFile Stream APIs that were introduced in CUDA 12.2.

This PR defers these definitions until the call to `find_package`, which will then use the exact cuFile features present (if cuFile is available at all) when building a package like cudf that depends on kvikio. The libkvikio example/test binary is built with the cuFile features available at build time, for use in the `libkvikio-tests` conda package. However, this test binary will still be compatible with a runtime where cuFile is unavailable or is version 12.0, as it is dlopen-ing the library and has runtime checks for the batch/stream features it tries to use.

I did local testing of this PR with cudf devcontainers. I tested both 12.0 and 12.2 to reproduce (and fix) the failure, and also tested clean builds of libcudf after removing `libcufile` (to test when cuFile is not found). All seems to work as intended.


cpp/CMakeLists.txt
```diff
@@ -145,11 +135,52 @@ rapids_export(
...
+    target_compile_definitions(kvikio::kvikio INTERFACE KVIKIO_CUFILE_STREAM_API_FOUND)
+  endif()
+endif()
+
+]=]
+)
+
 # build export targets
 rapids_export(
   BUILD kvikio
```
bdice, line 180: Do we need any changes in the `rapids_export(INSTALL kvikio ...)` command above? It doesn't seem that we do, from my local testing. I just want to be sure I'm not missing something.

jakirkham:
Thanks Bradley! 🙏

How would this affect the Python packages built on CUDA 12.2 (especially when they are used on CUDA 12.0)?

cpp/CMakeLists.txt
robertmaynard: Yes both the `BUILD` and `INSTALL` export commands should have the same final code block.

cpp/CMakeLists.txt
```diff
@@ -145,11 +135,52 @@ rapids_export(
...
 
+set(build_code_string
+    [=[
+
+# Find cuFile and determine which features are supported
+find_package(cuFile)
+if(NOT cuFile_FOUND)
+  message(WARNING "KvikIO: Building KvikIO without cuFile")
+else()
+  file(READ "${cuFile_INCLUDE_DIRS}/cufile.h" CUFILE_H_STR)
```
robertmaynard, line 146: This doesn't seem necessary. Instead you can propagate this information from the current CMake execution into the final code string.

Something like:
```cmake

set(code_string [=[
set(cuFile_BATCH_API_FOUND @cuFile_BATCH_API_FOUND@)
set(cuFile_STREAM_API_FOUND @cuFile_STREAM_API_FOUND@)

# Enable supported cuFile features in KvikIO
if(cuFile_FOUND)
  target_link_libraries(kvikio::kvikio INTERFACE cufile::cuFile_interface)
  target_compile_definitions(kvikio::kvikio INTERFACE KVIKIO_CUFILE_FOUND)
  if(cuFile_BATCH_API_FOUND)
    target_compile_definitions(kvikio::kvikio INTERFACE KVIKIO_CUFILE_BATCH_API_FOUND)
  endif()
  if(cuFile_STREAM_API_FOUND)
    target_compile_definitions(kvikio::kvikio INTERFACE KVIKIO_CUFILE_STREAM_API_FOUND)
  endif()
endif()
]=])
string(CONFIGURE "${code_string}" code_string @ONLY) 

rapids_export(
  INSTALL kvikio
  EXPORT_SET kvikio-exports
  GLOBAL_TARGETS kvikio
  NAMESPACE kvikio::
  DOCUMENTATION doc_string
  FINAL_CODE_BLOCK code_string
)
rapids_export(
  INSTALL kvikio
  EXPORT_SET kvikio-exports
  GLOBAL_TARGETS kvikio
  NAMESPACE kvikio::
  DOCUMENTATION doc_string
  FINAL_CODE_BLOCK code_string
)
```

robertmaynard (changes requested):

cpp/CMakeLists.txt
```diff
@@ -145,11 +135,52 @@ rapids_export(
...
+  if(cuFileReadAsync_location EQUAL "-1")
+    set(cuFile_STREAM_API_FOUND FALSE)
+  else()
+    set(cuFile_STREAM_API_FOUND TRUE)
+  endif()
+  message(STATUS "KvikIO: Found cuFile's Stream API: ${cuFile_STREAM_API_FOUND}")
+endif()
+
+# Enable supported cuFile features in KvikIO
+if(cuFile_FOUND)
```
robertmaynard, line 164: You are going to keep adding these defines, includes, link libraries each time `find_package` is called. We want to do this as rarely as possible,

One option is to wrap the whole check in something along the lines of:
```cmake
get_property(already_set_kvikio DIRECTORY PROPERTY kvikio_already_set_defines SET)
if(NOT already_set_kvikio)
  set_property(DIRECTORY PROPERTY kvikio_already_set_defines "ON")
   find_package(cuFile)
   ...
endif()
```

robertmaynard (changes requested):

jakirkham:
> How would this affect the Python packages built on CUDA 12.2 (especially when they are used on CUDA 12.0)?

To clarify, what happens if cuDF Python is build against KvikIO on CUDA 12.2 and then runs on CUDA 12.0?

Will we start making use of new symbols from cuFile on CUDA 12.2 that are not available on CUDA 12.0? How do we handle this?

robertmaynard (approved):

bdice:
> To clarify, what happens if cuDF Python is build against KvikIO on CUDA 12.2 and then runs on CUDA 12.0?
>
> Will we start making use of new symbols from cuFile on CUDA 12.2 that are not available on CUDA 12.0? How do we handle this?

The actual calls to cuFile are guarded by dlopen and conditional checks. Building any library that depends on libkvikio with 12.2 and running on 12.0 will still be safe. This PR just makes it possible to build a library _depending_ on libkvikio with different CUDA 12.x versions (12.0 or newer), rather than only the CUDA version that was used to build the libkvikio package (which is now 12.2).

jakirkham:
Are we sure? It looks compile time defined

https://github.com/rapidsai/kvikio/blob/e5bc18440afc344e4b463c1c1bc945b87116ce8a/cpp/include/kvikio/shim/cufile.hpp#L93-L99

jakirkham:
Should add am ok going ahead with the build improvement

Am more saying there are other issues we might encounter next and maybe we should get out ahead of them

bdice:
> Are we sure? It looks compile time defined

Correct. This PR defers the definition of those macros to the libcudf compilation rather than the libkvikio "compilation" (which is just using CMake to package libkvikio for conda, since libkvikio is header-only).

The problem this solves is that libkvikio conda packages built with CUDA 12.2 assumed the batch/stream features were _always_ available, and defined these macros even when building libcudf with CUDA 12.0 (cuFile from CUDA 12.0 doesn't have batch/stream features). That caused compilation errors.

If you build libcudf on CUDA 12.0, the batch/stream features aren't built into the library at all (no macros).
If you build libcudf on CUDA 12.2, the batch/stream features will be enabled by the macros and use runtime compatibility checks to determine whether to call those features (12.0 no, 12.2+ yes).

cpp/CMakeLists.txt
```diff
@@ -137,12 +127,56 @@ Provide targets for KvikIO: C++ bindings for cuFile.
...
 )
 
+set(final_code_string
+    [=[
+get_property(already_set_kvikio DIRECTORY PROPERTY kvikio_already_set_defines SET)
+if(NOT already_set_kvikio)
+  set_property(DIRECTORY PROPERTY kvikio_already_set_defines "ON")
+
+  # Find cuFile and determine which features are supported
+  find_package(cuFile)
```
vyasr, line 137:
```suggestion
  find_package(cuFile QUIET)
```

cpp/CMakeLists.txt
```diff
@@ -137,12 +127,56 @@ Provide targets for KvikIO: C++ bindings for cuFile.
...
+  if(NOT cuFile_FOUND)
+    message(WARNING "KvikIO: Building KvikIO without cuFile")
```
vyasr, line 139: It might be helpful to to consumers to word this warning and other messages below such that it's clear that the warning comes from using kvikio headers when building their own library (as opposed to kvikio being built by CPM, for instance).

cpp/examples/CMakeLists.txt
```diff
@@ -20,6 +20,18 @@ set_target_properties(BASIC_IO_TEST PROPERTIES INSTALL_RPATH "\$ORIGIN/../../lib
 target_include_directories(BASIC_IO_TEST PRIVATE ../include ${cuFile_INCLUDE_DIRS})
 target_link_libraries(BASIC_IO_TEST PRIVATE kvikio CUDA::cudart)
 
+# Enable supported cuFile features in KvikIO examples
+if(cuFile_FOUND)
```
vyasr, line 24: If we end up needing this in another place it would be good to extract this logic, but fine with the duplication for now.

vyasr (approved):
A couple of small suggestions, but nothing critical.

cpp/CMakeLists.txt
bdice: Can you suggest something specific? I don't understand your goal.

bdice: The messages here begin with `KvikIO: ` so it should be clear where this comes from.

vyasr: Hmm I was thinking of trying to differentiate building kvikio itself from building some pkg against kvikio, but I am also struggling to come up with good wording. Perhaps something like this?
```suggestion
    message(WARNING "${PROJECT_NAME}: Using KvikIO headers without cuFile")
```

jakirkham:
> The problem this solves is that libkvikio conda packages built with CUDA 12.2 assumed the batch/stream features were always available, and defined these macros even when building libcudf with CUDA 12.0 (cuFile from CUDA 12.0 doesn't have batch/stream features). That caused compilation errors. By deferring this check until find_package(KvikIO) is called, we ensure that the library that depends on libkvikio has build-time compatibility rules that are similar to CEC runtime compatibility rules.

Yep this is fine

> If you build libcudf on CUDA 12.0, the batch/stream features aren't built into the library at all (no macros).
If you build libcudf on CUDA 12.2, the batch/stream features will be enabled by the macros and use runtime compatibility checks to determine whether to call those features (12.0 no, 12.2+ yes).

This is what I was worried about. Though it sounds like it shouldn't be an issue

cpp/CMakeLists.txt
bdice: This will never be called except when building some package against kvikio. I think the current message is clear enough and don't really want to do variable injection here...

bdice: The messages are immediately followed by the CPM message about finding kvikio, so it's clear that it's happening during the find_package call.

vyasr: I don't follow why you're concerned about "variable injection" here. There's nothing magical here.

The confusing case will be if CPM is adding kvikio from source instead of being found as an installed package, in which case you'll see any messages that occur as part of kvikio's build as well as these messages that occur when the built kvikio is found.

PR #346, opened 2024-02-23
Bump to nvcomp 3.0.6.

bdice:
This PR bumps nvcomp to 3.0.6. This is needed as a hotfix for https://github.com/rapidsai/cudf/issues/15096.

Depends on:
- https://github.com/conda-forge/nvcomp-feedstock/pull/14
- https://github.com/rapidsai/rapids-cmake/pull/542

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -173,7 +173,7 @@ dependencies:
     common:
       - output_types: conda
         packages:
-          - nvcomp==3.0.5
+          - nvcomp==3.0.6
     specific:
       - output_types: conda
         matrices:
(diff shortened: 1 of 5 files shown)


dependencies.yaml
```diff
@@ -283,8 +282,3 @@ dependencies:
         packages:
           - *dask
           - distributed>=2022.05.2
-  notebooks:
-    common:
-      - output_types: conda
-        packages:
-          - cudf==24.2.*
```
bdice, line 290: There were no references to `cudf` in the notebooks, so I removed this. This dependency was causing devcontainer builds to fail. This has been a pain in the past because it introduced a circular dependency between kvikio and cudf, but it seems safe to remove now.

bdice:
I checked the CI jobs and they used `nvcomp==3.0.6`. This should be good to go.

raydouglass (approved):

PR #356, opened 2024-03-25
Prevent path conflict in builds

AyodeAwe:
This will make builds fail when there are path conflicts


madsbk (approved):
LGTM, thanks @AyodeAwe

raydouglass (approved):

PR #364, opened 2024-04-06
Add deflate codec

ap--:
Hello everyone,

We've been using nvcomp's deflate decompression in https://github.com/pfizer-opensource/bigwig-loader to load data from a legacy genome track format and it would make distribution of our code a lot easier if kvikio would ship the deflate codec wrappers too.

This PR exposes nvcomp's deflate algorithm in kvikio and adds it to the nvcomp_codec tests.

Cheers,
Andreas


madsbk (approved):
Looks good, thanks @ap-- 

PS: ran `pre-commit run --all-file` to fix styles

PR #372, opened 2024-04-25
Refactor: move cufile header checks to the shim layer

madsbk:
Moving all the `#ifdef KVIKIO_CUFILE_FOUND` to the shim layer

### DIFF
--- a/cpp/include/kvikio/shim/cufile.hpp
+++ b/cpp/include/kvikio/shim/cufile.hpp
@@ -44,21 +44,16 @@ class cuFileAPI {
   decltype(cuFileDriverSetPollMode)* DriverSetPollMode{nullptr};
   decltype(cuFileDriverSetMaxCacheSize)* DriverSetMaxCacheSize{nullptr};
   decltype(cuFileDriverSetMaxPinnedMemSize)* DriverSetMaxPinnedMemSize{nullptr};
-
-#ifdef KVIKIO_CUFILE_BATCH_API_FOUND
   decltype(cuFileBatchIOSetUp)* BatchIOSetUp{nullptr};
   decltype(cuFileBatchIOSubmit)* BatchIOSubmit{nullptr};
   decltype(cuFileBatchIOGetStatus)* BatchIOGetStatus{nullptr};
   decltype(cuFileBatchIOCancel)* BatchIOCancel{nullptr};
   decltype(cuFileBatchIODestroy)* BatchIODestroy{nullptr};
-#endif
-
-#ifdef KVIKIO_CUFILE_STREAM_API_FOUND
   decltype(cuFileReadAsync)* ReadAsync{nullptr};
   decltype(cuFileWriteAsync)* WriteAsync{nullptr};
   decltype(cuFileStreamRegister)* StreamRegister{nullptr};
   decltype(cuFileStreamDeregister)* StreamDeregister{nullptr};
-#endif
+
   bool stream_available = false;
 
  private:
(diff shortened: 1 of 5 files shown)


cpp/include/kvikio/shim/cufile_h_wrapper.hpp
```diff
@@ -25,11 +28,49 @@
...
+ssize_t cuFileWrite(...);
+CUfileError_t cuFileBufRegister(...);
+CUfileError_t cuFileBufDeregister(...);
+CUfileError_t cuFileDriverOpen(...);
+CUfileError_t cuFileDriverClose(...);
+CUfileError_t cuFileDriverGetProperties(...);
+CUfileError_t cuFileDriverSetPollMode(...);
+CUfileError_t cuFileDriverSetMaxCacheSize(...);
+CUfileError_t cuFileDriverSetMaxPinnedMemSize(...);
+
```
wence-, line 73: OK, so this stubs out just enough of the cufile API that you can _compile_ without `KVIKIO_CUFILE_FOUND`, but then always produces an error when using?

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -128,10 +122,16 @@ class cuFileAPI {
                 << std::endl;
     }
   }
+#else
+  cuFileAPI() { throw std::runtime_error("KvikIO not compiled with cuFile.h"); }
+  ~cuFileAPI() = default;
```
wence-, line 127:
```suggestion
```
Unnecessary `default`?

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -128,10 +122,16 @@ class cuFileAPI {
                 << std::endl;
     }
   }
+#else
+  cuFileAPI() { throw std::runtime_error("KvikIO not compiled with cuFile.h"); }
```
wence-, line 126:
```suggestion
  cuFileAPI() { throw std::runtime_error(CUFILE_ERRSTR(0)); }
```

To OAOO the error message.

cpp/include/kvikio/shim/cufile_h_wrapper.hpp
```diff
@@ -25,11 +28,49 @@
...
+enum CUfileFileHandleType { CU_FILE_HANDLE_TYPE_OPAQUE_FD = 1 };
+
+struct CUfileDescr_t {
+  enum CUfileFileHandleType type;
+  struct handle_t {
+    int fd;
+  } handle;
+};
+
+const char* cufileop_status_error(...) { return "KvikIO not compiled with cuFile.h"; };
```
wence-, line 60:
```suggestion
static inline const char* cufileop_status_error(CUfileOpError err) { return CUFILE_ERRSTR(err); };
```
`static inline` for ODR, and `CUFILE_ERRSTR` for OAOO for the error string.

cpp/include/kvikio/shim/cufile_h_wrapper.hpp
```diff
@@ -25,11 +28,49 @@
...
+CUfileError_t cuFileHandleRegister(...);
+CUfileError_t cuFileHandleDeregister(...);
+ssize_t cuFileRead(...);
+ssize_t cuFileWrite(...);
+CUfileError_t cuFileBufRegister(...);
```
wence-, line 65: The lack of definitions of these is not a problem because the linker will never look for these symbols because the "real" function calls are made through the shim instance?

wence- (approved):
Some minor queries and maybe a fixup or two, looks good.

cpp/include/kvikio/shim/cufile_h_wrapper.hpp
madsbk: yes

cpp/include/kvikio/shim/cufile_h_wrapper.hpp
madsbk: yes, added your question as a comment

PR #374, opened 2024-04-29
POSIX IO: use async memcpy

### LINKED ISSUE #311: Use async `cuMemcpy` APIs instead of synchronous ones
https://github.com/rapidsai/kvikio/blob/branch-23.12/cpp/include/kvikio/posix_io.hpp#L179-L185

Currently, it looks like synchronous `cuMemcpy` APIs are being used which effectively synchronize the entire device. If someone is using streams explicitly or is using multiple threads with per thread default stream, this will cause unnecessary synchronization.

Instead, we should probably move to using the async `cuMemcpy` APIs and then synchronizing the stream explicitly

madsbk:
Closes #311

### DIFF
--- a/cpp/include/kvikio/posix_io.hpp
+++ b/cpp/include/kvikio/posix_io.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


cpp/include/kvikio/posix_io.hpp
```diff
@@ -33,7 +35,60 @@ inline constexpr std::size_t posix_bounce_buffer_size = 2 << 23;  // 16 MiB
...
+ private:
+  std::map<std::pair<CUcontext, std::thread::id>, CUstream> _streams;
+
+ public:
+  StreamsByThread() = default;
+  ~StreamsByThread() noexcept
+  {
+    for (auto& [_, stream] : _streams) {
+      try {
+        CUDA_DRIVER_TRY(cudaAPI::instance().StreamDestroy(stream));
```
vuule, line 53: Doesn't this segfault when destroying the singleton? We had to let the libcudf stream pool "leak" to avoid this.

madsbk: Not on my workstation or in CI. Maybe it is because we are using a newer CTK. Do you have a link to the libcudf issue?

vuule (approved):

cpp/include/kvikio/posix_io.hpp
madsbk: Thanks for checking

PR #375, opened 2024-05-01
Enable Warnings as errors for Python tests

mroeschke:
As part of https://github.com/rapidsai/build-planning/issues/26, warnings from the Python test suite will cause a test to fail


madsbk (approved):
Looks good, thanks @mroeschke

wence- (approved):

Issue #204, opened 2023-04-25
Tracking support of new cuFile features 


madsbk:
Meta Issue to track support of new cuFile features.

- [x] [Batch IO (C++)](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-batch-api-functional-specification):  https://github.com/rapidsai/kvikio/pull/220
- [ ] [Batch IO (Python)](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-batch-api-functional-specification)
- [x] [Async IO (C++](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-stream-api): https://github.com/rapidsai/kvikio/pull/259
- [x] [Async IO (Python)](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-stream-api)

jakirkham:
C++ support for Batch IO was done in PR ( https://github.com/rapidsai/kvikio/pull/220 ), right? Or is this about Python support?

madsbk:
> C++ support for Batch IO was done in PR ( #220 ), right? Or is this about Python support?

Yes, updated the issue

fstrug:
Is there any timeline for batch IO support in python?

madsbk:
> Is there any timeline for batch IO support in python?

No, not at the moment but we could prioritize it. Do you have a particular use case in mind?

fstrug:
> Do you have a particular use case in mind?

We are working on developing a tool for high energy particle physicists that will use the GPU to read and store data directly on the GPU for later use in an analysis. Our data is stored row-wise, so the bytes for any column are divided into many small baskets that are spread throughout the length of the file. To get a column of data out of the file to an array, we are performing many small `CuFile.preads()` (~300 reads of 10^5 bytes) at different offsets. With the CuFile API calls being FIFO, it seems that a batch API call would be the performant way to launch these reads.

madsbk:
Yes, sounds like the batch API could be useful. 

Currently with `CuFile.preads()`, are you using the [thread pool](https://docs.rapids.ai/api/kvikio/nightly/runtime_settings/#thread-pool-kvikio-nthreads) be setting `KVIKIO_NTHREADS` or calling `kvikio.default.num_threads_reset()`?

fstrug:
[...]

PR #376, opened 2024-05-05
Initial Python Interface for cufile Async IO

K-Wu:
Hi there,

Thanks for this great repository! I want to use the cuFile async IO in my research project and noticed this kvikio repo. However, the initial support has been done in #259 and tracked in #204, but the Python interface hasn't been done yet. So I exported the write_async and read_async to the CuFile Python class and added test case. This will be very helpful for my project where I want to do the PyTorch training computation and simultaneously load tensors from the SSDs. I created this PR because hopefully, it could be helpful for your repository as well as keeping the Python interface current.

Please let me know your thoughts. Thank you.

Best Regards,
Kun

### DIFF
--- a/python/kvikio/_lib/kvikio_cxx_api.pxd
+++ b/python/kvikio/_lib/kvikio_cxx_api.pxd
@@ -4,17 +4,16 @@
 # distutils: language = c++
 # cython: language_level=3
 
-cdef extern from "driver_types.h":
-    ctypedef struct CUstream_st:
-        pass
-    ctypedef CUstream_st *cudaStream_t
-
 from posix cimport fcntl
 
 from libcpp cimport bool
 from libcpp.string cimport string
 from libcpp.utility cimport pair
 from libcpp.vector cimport vector
+
+
+cdef extern from "cuda.h":
+    ctypedef void* CUstream
 
 
 cdef extern from "<future>" namespace "std" nogil:
@@ -114,12 +113,12 @@ cdef extern from "<kvikio/file_handle.hpp>" namespace "kvikio" nogil:
             size_t size,
             size_t file_offset,
             size_t devPtr_offset,
-            cudaStream_t stream
+            CUstream stream
         ) except +
         StreamFuture write_async(
             void* devPtr_base,
             size_t size,
             size_t file_offset,
             size_t devPtr_offset,
-            cudaStream_t stream
+            CUstream stream
         ) except +
--- a/python/kvikio/_lib/kvikio_cxx_api.pxd
+++ b/python/kvikio/_lib/kvikio_cxx_api.pxd
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
 # distutils: language = c++
--- a/python/kvikio/_lib/kvikio_cxx_api.pxd
+++ b/python/kvikio/_lib/kvikio_cxx_api.pxd
@@ -4,6 +4,12 @@
 # distutils: language = c++
 # cython: language_level=3
 
+cdef extern from "driver_types.h":
+    # Adapted from https://github.com/andersbll/cudarray/blob/a2cffbb1434db9a7e6ed83211300d23d47630d2e/cudarray/wrap/cudart.pxd#L16
+    ctypedef struct CUstream_st:
+        pass
+    ctypedef CUstream_st *cudaStream_t
+
 from posix cimport fcntl
 
 from libcpp cimport bool
@@ -17,6 +23,10 @@ cdef extern from "<future>" namespace "std" nogil:
         future() except +
         T get() except +
 
+cdef extern from "<kvikio/stream.hpp>" namespace "kvikio" nogil:
+    cdef cppclass StreamFuture:
+        StreamFuture() except +
+        size_t check_bytes_done() except +
 
 cdef extern from "<kvikio/utils.hpp>" namespace "kvikio" nogil:
     bool is_future_done[T](const T& future) except +
@@ -96,4 +106,18 @@ cdef extern from "<kvikio/file_handle.hpp>" namespace "kvikio" nogil:
             size_t size,
             size_t file_offset,
             size_t devPtr_offset
+        ) except +
+        StreamFuture read_async(
+            void* devPtr_base,
+            size_t size,
+            size_t file_offset,
+            size_t devPtr_offset,
+            cudaStream_t stream
+        ) except +
+        StreamFuture write_async(
+            void* devPtr_base,
+            size_t size,
+            size_t file_offset,
+            size_t devPtr_offset,
+            cudaStream_t stream
         ) except +
--- a/python/kvikio/_lib/kvikio_cxx_api.pxd
+++ b/python/kvikio/_lib/kvikio_cxx_api.pxd
@@ -1,4 +1,4 @@
-# Copyright (c) 2021-2024, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
 # distutils: language = c++
--- a/python/kvikio/_lib/kvikio_cxx_api.pxd
+++ b/python/kvikio/_lib/kvikio_cxx_api.pxd
@@ -1,11 +1,10 @@
-# Copyright (c) 2021-2023, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
 # distutils: language = c++
 # cython: language_level=3
 
 cdef extern from "driver_types.h":
-    # Adapted from https://github.com/andersbll/cudarray/blob/a2cffbb1434db9a7e6ed83211300d23d47630d2e/cudarray/wrap/cudart.pxd#L16
     ctypedef struct CUstream_st:
         pass
     ctypedef CUstream_st *cudaStream_t
@@ -23,10 +22,13 @@ cdef extern from "<future>" namespace "std" nogil:
         future() except +
         T get() except +
 
+
 cdef extern from "<kvikio/stream.hpp>" namespace "kvikio" nogil:
     cdef cppclass StreamFuture:
         StreamFuture() except +
+        StreamFuture(StreamFuture&&) except +
         size_t check_bytes_done() except +
+
 
 cdef extern from "<kvikio/utils.hpp>" namespace "kvikio" nogil:
     bool is_future_done[T](const T& future) except +
--- a/python/kvikio/_lib/libkvikio.pyx
+++ b/python/kvikio/_lib/libkvikio.pyx
@@ -12,16 +12,10 @@ from libcpp.utility cimport move, pair
 
 from . cimport kvikio_cxx_api
 from .arr cimport Array
-from .kvikio_cxx_api cimport (
-    FileHandle,
-    StreamFuture,
-    cudaStream_t,
-    future,
-    is_future_done,
-)
+from .kvikio_cxx_api cimport CUstream, FileHandle, StreamFuture, future, is_future_done
 
 
-cdef class PyStreamFuture:
+cdef class IOFutureStream:
     """Wrap a C++ StreamFuture in a Python object"""
     cdef StreamFuture _handle
 
@@ -29,9 +23,9 @@ cdef class PyStreamFuture:
         return self._handle.check_bytes_done()
 
 
-cdef PyStreamFuture _wrap_stream_future(StreamFuture &fut):
+cdef IOFutureStream _wrap_stream_future(StreamFuture &fut):
     """Wrap a C++ future (of a `size_t`) in a `IOFuture` instance"""
-    ret = PyStreamFuture()
+    ret = IOFutureStream()
     ret._handle = move(fut)
     return ret
 
@@ -187,8 +181,8 @@ cdef class CuFile:
         )
 
     def read_async(self, buf, size: Optional[int], file_offset: int, dev_offset: int,
-                   st: uintptr_t) -> PyStreamFuture:
-        stream = <cudaStream_t>st
+                   st: uintptr_t) -> IOFutureStream:
+        stream = <CUstream>st
         cdef pair[uintptr_t, size_t] info = _parse_buffer(buf, size, False)
         # TODO: return StreamFuture
         return _wrap_stream_future(self._handle.read_async(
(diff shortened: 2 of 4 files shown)


K-Wu:
In this implementation, cuda stream's pointer is passed into the Python interface, which can be retrieved from the stream object in pycuda, cuda-python, torch. The stream objects in the three packages are interoperable as well. You may check the following gist on how to create a stream object based on another from another package and retrieve the address from these objects.

https://gist.github.com/K-Wu/178703e4803716d61c94df8938c3da9b

K-Wu:
The failing tests in [conda-python-tests](https://github.com/rapidsai/kvikio/actions/runs/8965107867/job/24618313344?pr=376#logs) are caused by torch being not installed. How shall I fix it? Maybe adding some flags to indicate this test requires GPU and pytorch?

madsbk:
> The failing tests in [conda-python-tests](https://github.com/rapidsai/kvikio/actions/runs/8965107867/job/24618313344?pr=376#logs) are caused by torch being not installed. How shall I fix it? Maybe adding some flags to indicate this test requires Nvidia GPU and pytorch? Instead of pytorch, pycuda or cuda-python can also do the job.

If we need pytorch to create a new stream, I suggest that we use `stream = cupy.cuda.Stream()` and `stream.ptr`. KvikIO depend on cupy already.

python/kvikio/_lib/libkvikio.pyx
```diff
@@ -12,7 +12,28 @@ from libcpp.utility cimport move, pair
...
+from .kvikio_cxx_api cimport (
+    FileHandle,
+    StreamFuture,
+    cudaStream_t,
+    future,
+    is_future_done,
+)
+
+
+cdef class PyStreamFuture:
```
madsbk, line 24: I think we should rename this to something like `IOFutureStream`. Or maybe, we can extend `IOFuture` to also have an optional stream?

python/kvikio/cufile.py
```diff
@@ -7,6 +7,23 @@
 from ._lib import libkvikio  # type: ignore
 
 
+class PyStreamFuture:
```
madsbk, line 10: Important to include the lifetime info both here and in the read/write functions. Something like we have in the C++ API:
```
   * The arguments have the same meaning as in `.read()` but returns a `StreamFuture` object
   * that the caller must keep alive until all data has been read from disk. One way to do this,
   * is by calling `StreamFuture.check_bytes_done()`, which will synchronize the associated stream
   * and return the number of bytes read.
```

python/kvikio/cufile.py
```diff
@@ -260,8 +277,74 @@ def write(
         """
         return self.pwrite(buf, size, file_offset, task_size).get()
 
+    def read_async(
```
madsbk, line 280: I suggest we rename this to `raw_read_async`

python/kvikio/cufile.py
```diff
@@ -260,8 +277,74 @@ def write(
...
+            CUDA stream to perform the read operation asynchronously.
+
+        Returns
+        -------
+        int
+            The size of bytes that were successfully read.
+        """
+        return self._handle.read_async(buf, size, file_offset, dev_offset, stream)
+
+    def write_async(
```
madsbk, line 311: I suggest we rename this to `raw_write_async`

madsbk (changes requested):
Thanks @K-Wu, nice work!

K-Wu:
@madsbk Hi Mads, thanks for this very quick response and detailed review! I will address your comments tonight.

Best Regards,
Kun

python/kvikio/_lib/libkvikio.pyx
K-Wu: Good points and I've done the rename! The extended class seems a bit overloaded. Maybe we can think about how to combine the two classes in later PRs? I am slacking here :D

python/kvikio/_lib/kvikio_cxx_api.pxd
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2021-2023, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION. All rights reserved.
```
madsbk, line 1:
```suggestion
# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
```

python/kvikio/cufile.py
```diff
@@ -260,8 +280,74 @@ def write(
...
+        self,
+        buf,
+        stream,
+        size: Optional[int] = None,
+        file_offset: int = 0,
+        dev_offset: int = 0,
+    ) -> IOFutureStream:
+        """Reads specified bytes from the file into the device memory asynchronously
+
+        This is a low-level version of `.read` that doesn't use threads and
```
madsbk, line 293:
```suggestion
        This is an async version of `.raw_read` that doesn't use threads and
```

python/kvikio/cufile.py
```diff
@@ -260,8 +280,74 @@ def write(
...
+        self,
+        buf,
+        stream,
+        size: Optional[int] = None,
+        file_offset: int = 0,
+        dev_offset: int = 0,
+    ) -> IOFutureStream:
+        """Writes specified bytes from the device memory into the file asynchronously
+
+        This is a low-level version of `.write` that doesn't use threads and
```
madsbk, line 324:
```suggestion
        This is an async version of `.raw_write` that doesn't use threads and
```

python/tests/test_async_io.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2022-2024, NVIDIA CORPORATION. All rights reserved.
```
madsbk, line 1:
```suggestion
# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
```

python/tests/test_async_io.py
```diff
@@ -0,0 +1,49 @@
...
+            # Write file
+            a = cupy.arange(size)
+            f = kvikio.CuFile(filename, "w")
+            assert not f.closed
+            assert check_bit_flags(f.open_flags(), os.O_WRONLY)
+            assert f.raw_write_async(a, stream.ptr).check_bytes_done() == a.nbytes
+
+            # Try to read file opened in write-only mode
+            with pytest.raises(RuntimeError, match="unsupported file open flags"):
+                f.raw_read_async(a, stream.ptr).check_bytes_done()
```
madsbk, line 38:
```suggestion
            future_stream = f.raw_read_async(a, stream.ptr)
            with pytest.raises(RuntimeError, match="unsupported file open flags"):
                # The exception is raised when we check for completion.
                future_stream.check_bytes_done()
```

python/tests/test_async_io.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def check_bit_flags(x: int, y: int) -> bool:
+    """Check that the bits set in `y` is also set in `x`"""
+    return x & y == y
+
+
+@pytest.mark.parametrize("size", [1, 10, 100, 1000, 1024, 4096, 4096 * 10])
+@pytest.mark.parametrize("nthreads", [1, 3, 4, 16])
+@pytest.mark.parametrize("tasksize", [199, 1024])
```
madsbk, line 20: Since `raw_read_async` and `raw_write_async` doesn't use the thread pool, we don't need `nthreads` and `tasksize`. 

Also, currently the C++ API's `read_async` and `write_async` doesn't use the `gds_threshold` argument.

python/tests/test_async_io.py
```diff
@@ -0,0 +1,49 @@
...
+            # Close file
+            f.close()
+            assert f.closed
+
+            # Read file into a new array and compare
+            b = cupy.empty_like(a)
+            f = kvikio.CuFile(filename, "r")
+            assert check_bit_flags(f.open_flags(), os.O_RDONLY)
+            assert f.raw_read_async(b, stream.ptr).check_bytes_done() == b.nbytes
+            assert all(a == b)
```
madsbk, line 49: I would be good to test a stream pipeline. 
So instead of calling `check_bytes_done()` immediately after `raw_read_async`, add a new `raw_write_async` that  writes `b` to a new file using `stream.ptr` and then call `check_bytes_done()` (and check the written data was correct)

madsbk (changes requested):
Looks good, I only have some minor suggestions

python/kvikio/_lib/kvikio_cxx_api.pxd
K-Wu: This change cannot be done because it will be overwritten by the pre-commit hook with the original text before change

python/tests/test_async_io.py
K-Wu: Comments addressed. However, the exception is actually raised when .raw_read_async was triggered. I added in the comment that this is the expected behavior.

K-Wu:
Hi @madsbk, thanks for the review! Please let me know if further actions need to be done.

Also, shall I update the branch by clicking the "Update branch" button? It seems this repo used squash merge to deal with PRs so it seems okay. I know some people prefer rebase when pulling the main branch.

python/kvikio/_lib/kvikio_cxx_api.pxd
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2021-2023, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
```
madsbk, line 1:
```suggestion
# Copyright (c) 2021-2024, NVIDIA CORPORATION. All rights reserved.
```

python/kvikio/cufile.py
```diff
@@ -260,8 +280,76 @@ def write(
...
+        Parameters
+        ----------
+        buf: buffer-like or array-like
+            Device buffer to read into.
+        size: int, optional
+            Size in bytes to read.
+        file_offset: int, optional
+            Offset in the file to read from.
+        stream: cuda.Stream, optional
+            CUDA stream to perform the read operation asynchronously.
```
madsbk, line 305: order to match the function def. Also,  I don't think `stream` is optional?

python/kvikio/cufile.py
```diff
@@ -260,8 +280,76 @@ def write(
...
+        Parameters
+        ----------
+        buf: buffer-like or array-like
+            Device buffer to write to.
+        size: int, optional
+            Size in bytes to write.
+        file_offset: int, optional
+            Offset in the file to write from.
+        stream: cuda.Stream, optional
+            CUDA stream to perform the write operation asynchronously.
```
madsbk, line 337: order to match the function def. Also,  I don't think `stream` is optional?

python/kvikio/cufile.py
```diff
@@ -7,6 +7,26 @@
...
+class IOFutureStream:
+    """Future for CuFile async IO
+
+    This class shouldn't be used directly, instead non-blocking async IO operations
+    such as `CuFile.raw_read_async` and `CuFile.raw_write_async` returns an instance
+    of this class.
+
+    The instance must be kept alive alive until all data has been read from disk. One
+    way to do this, is by calling `StreamFuture.check_bytes_done()`, which will
+    synchronize the associated stream and return the number of bytes read."""
```
madsbk, line 19:
```suggestion
    synchronize the associated stream and return the number of bytes read.
    """
```

python/kvikio/cufile.py
```diff
@@ -260,8 +280,76 @@ def write(
...
+        size: int, optional
+            Size in bytes to write.
+        file_offset: int, optional
+            Offset in the file to write from.
+        stream: cuda.Stream, optional
+            CUDA stream to perform the write operation asynchronously.
+
+        Returns
+        -------
+        IOFuture
```
madsbk, line 341:
```suggestion
        IOFutureStream
```

Also, add something like:
```
The instance must be kept alive alive until all data has been written to disk. One
    way to do this, is by calling `StreamFuture.check_bytes_done()`, which will
    synchronize the associated stream and return the number of bytes written."""
```

python/kvikio/cufile.py
```diff
@@ -260,8 +280,76 @@ def write(
...
+        size: int, optional
+            Size in bytes to read.
+        file_offset: int, optional
+            Offset in the file to read from.
+        stream: cuda.Stream, optional
+            CUDA stream to perform the read operation asynchronously.
+
+        Returns
+        -------
+        IOFuture
```
madsbk, line 309:
```suggestion
        IOFutureStream
```

Also, add something like:
```
The instance must be kept alive alive until all data has been read from disk. One
    way to do this, is by calling `StreamFuture.check_bytes_done()`, which will
    synchronize the associated stream and return the number of bytes read."""
```

madsbk (changes requested):
Nitpick

K-Wu:
Done addressing the comments

python/kvikio/cufile.py
```diff
@@ -260,8 +281,82 @@ def write(
...
+        file_offset: int, optional
+            Offset in the file to read from.
+
+        Returns
+        -------
+        IOFutureStream
+            Future that when executed ".check_bytes_done()" returns the size of bytes
+            that were successfully read. The instance must be kept alive alive until
+            all data has been read from disk. One way to do this, is by calling
+            `StreamFuture.check_bytes_done()`, which will synchronize the associated
```
madsbk, line 314:
```suggestion
            `IOFutureStream.check_bytes_done()`, which will synchronize the associated
```

python/kvikio/cufile.py
```diff
@@ -260,8 +281,82 @@ def write(
...
+        file_offset: int, optional
+            Offset in the file to write from.
+
+        Returns
+        -------
+        IOFutureStream
+            Future that when executed ".check_bytes_done()" returns the size of bytes
+            that were successfully written. The instance must be kept alive alive until
+            all data has been written to disk. One way to do this, is by calling
+            `StreamFuture.check_bytes_done()`, which will synchronize the associated
```
madsbk, line 349:
```suggestion
            `IOFutureStream.check_bytes_done()`, which will synchronize the associated
```

python/kvikio/cufile.py
```diff
@@ -260,8 +281,82 @@ def write(
...
+        size: int, optional
+            Size in bytes to read.
+        file_offset: int, optional
+            Offset in the file to read from.
+
+        Returns
+        -------
+        IOFutureStream
+            Future that when executed ".check_bytes_done()" returns the size of bytes
+            that were successfully read. The instance must be kept alive alive until
```
madsbk, line 312:
```suggestion
            that were successfully read. The instance must be kept alive until
```

python/kvikio/cufile.py
```diff
@@ -260,8 +281,82 @@ def write(
...
+        size: int, optional
+            Size in bytes to write.
+        file_offset: int, optional
+            Offset in the file to write from.
+
+        Returns
+        -------
+        IOFutureStream
+            Future that when executed ".check_bytes_done()" returns the size of bytes
+            that were successfully written. The instance must be kept alive alive until
```
madsbk, line 347:
```suggestion
            that were successfully written. The instance must be kept alive until
```

K-Wu:
@madsbk Hi Mads, thank you for your further reviews and edits!

madsbk (approved):

madsbk:
Thanks @K-Wu

Issue #378, opened 2024-05-14
Conda package isn't build with GDS support!


madsbk:
### Reproducer

```
mamba create -n kvikio-env -c rapidsai-nightly -c conda-forge python=3.11 cuda-version=12.2 kvikio
conda activate kvikio-env
pip install dask

# Run the benchmark with compatibility mode disabled, which forces KvikIO to use cuFile
KVIKIO_COMPAT_MODE=off python python/benchmarks/single-node-io.py
```
```
Roundtrip benchmark
----------------------------------
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   WARNING - cuFile compat mode   
         GDS not enabled          
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
GPU               | Unknown (install pynvml)
GPU Memory Total  | Unknown (install pynvml)
BAR1 Memory Total | Unknown (install pynvml)
GDS driver        | N/A (Compatibility Mode)
GDS config.json   | /etc/cufile.json
----------------------------------
nbytes            | 10485760 bytes (10.00 MiB)
4K aligned        | True
pre-reg-buf       | True
directory         | /tmp/tmp71d24zvz
nthreads          | 1
nruns             | 1
==================================
Traceback (most recent call last):
  File "/home/mkristensen/repos/kvikio/python/benchmarks/single-node-io.py", line 429, in <module>
    main(args)
  File "/home/mkristensen/repos/kvikio/python/benchmarks/single-node-io.py", line 326, in main
    read, write = API[api](args)
  File "/home/mkristensen/repos/kvikio/python/benchmarks/single-node-io.py", line 47, in run_cufile
    res = f.write(data)
  File "/home/mkristensen/apps/miniforge3/envs/kvikio-env2404/lib/python3.10/site-packages/kvikio/cufile.py", line 261, in write
    return self.pwrite(buf, size, file_offset, task_size).get()
  File "/home/mkristensen/apps/miniforge3/envs/kvikio-env2404/lib/python3.10/site-packages/kvikio/cufile.py", line 33, in get
    return self._handle.get()
  File "libkvikio.pyx", line 25, in kvikio._lib.libkvikio.IOFuture.get
RuntimeError: KvikIO not compiled with cuFile.h
```

The latest version that works is `24.02`, the issue must have started somewhere between version `24.02` and `24.04`. 

cc. @aslobodaNV

KyleFromNVIDIA:
[...]

PR #379, opened 2024-05-14
Move cuFile linking to kvikio target

KyleFromNVIDIA:
The exported `kvikio` target has an `INTERFACE` dependency on cuFile. Add this dependency in the local buildsystem too for consistency, and remove the usage from the examples. This will allow all targets that link against `kvikio` to automatically get cuFile, including the Python modules.

Fixes: #378

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -86,14 +86,13 @@ add_library(kvikio::kvikio ALIAS kvikio)
 
 # Enable supported cuFile features in KvikIO
 #
-# While the code block below (in the value of final_code_string) sets this information
-# when KvikIO is imported from a kvikio-config.cmake file, this code block is intended to
-# be used by projects that include KvikIO's source directory in their own CMake build.
+# While the code block below (in the value of final_code_string) sets this information when KvikIO
+# is imported from a kvikio-config.cmake file, this code block is intended to be used by projects
+# that include KvikIO's source directory in their own CMake build.
 #
-# Normally we would just set the below without using $<BUILD_LOCAL_INTERFACE:...>, and
-# without the final_code_string, but in this case we want to conditionally set these
-# things at import time, not export time, since KvikIO is a header-only library that can
-# adapt to different build environments.
+# Normally we would just set the below without using $<BUILD_LOCAL_INTERFACE:...>, and without the
+# final_code_string, but in this case we want to conditionally set these things at import time, not
+# export time, since KvikIO is a header-only library that can adapt to different build environments.
 if(cuFile_FOUND)
   target_link_libraries(kvikio INTERFACE $<BUILD_LOCAL_INTERFACE:cufile::cuFile_interface>)
   target_compile_definitions(kvikio INTERFACE $<BUILD_LOCAL_INTERFACE:KVIKIO_CUFILE_FOUND>)
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -85,7 +85,7 @@ add_library(kvikio INTERFACE)
 add_library(kvikio::kvikio ALIAS kvikio)
 
 # Enable supported cuFile features in KvikIO
-if(FALSE)
+if(cuFile_FOUND)
   target_link_libraries(kvikio INTERFACE $<BUILD_LOCAL_INTERFACE:cufile::cuFile_interface>)
   target_compile_definitions(kvikio INTERFACE $<BUILD_LOCAL_INTERFACE:KVIKIO_CUFILE_FOUND>)
   if(cuFile_BATCH_API_FOUND)
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -85,7 +85,7 @@ add_library(kvikio INTERFACE)
 add_library(kvikio::kvikio ALIAS kvikio)
 
 # Enable supported cuFile features in KvikIO
-if(cuFile_FOUND)
+if(FALSE)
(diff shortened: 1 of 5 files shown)


bdice (approved):
Yes, kvikio's cuFile dependency should be handled the same way we handle cuFile packages in libkvikio. This looks right to me. Thanks!

I've only ever checked this for libkvikio because that's what we use in libcudf.

KyleFromNVIDIA:
What I don't understand is why this doesn't already work. `libkvikio` has a `run` dependency on cuFile already:

https://github.com/rapidsai/kvikio/blob/6ed7bccb42d82eb1ac60a98dfd460a7785881949/conda/recipes/libkvikio/meta.yaml#L77-L85

KyleFromNVIDIA:
...and it seems this change doesn't work either. I need to go back to the drawing board.

jakirkham:
Thanks Kyle! 🙏

Think `libkvikio` is header only. So adding `libcufile-dev` doesn't solve things for other libraries, which will need to add `libcufile-dev` themselves

Am curious if there are more details on what is not working with this change. Would also have expected this to work

KyleFromNVIDIA:
I'm still getting the same error:

```
  File "/home/kyedwards/development/kvikio/src/python/benchmarks/single-node-io.py", line 429, in <module>
    main(args)
  File "/home/kyedwards/development/kvikio/src/python/benchmarks/single-node-io.py", line 326, in main
    read, write = API[api](args)
                  ^^^^^^^^^^^^^^
  File "/home/kyedwards/development/kvikio/src/python/benchmarks/single-node-io.py", line 42, in run_cufile
    kvikio.memory_register(data)
  File "/home/kyedwards/anaconda3/envs/kvikio-env/lib/python3.11/site-packages/kvikio/__init__.py", line 10, in memory_register
    return libkvikio.memory_register(buf)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libkvikio.pyx", line 58, in kvikio._lib.libkvikio.memory_register
RuntimeError: KvikIO not compiled with cuFile.h
```

jakirkham:
Could you please share a bit more on how that is coming up? Is that one of the GHA jobs here? Or is that a local build? If so, are there any details?

<hr>

Taking a look at [one of the CUDA 12.2 `x86_64` GHA jobs]( https://github.com/rapidsai/kvikio/actions/runs/9084882174/job/24967129961?pr=379#step:7:630 ), it appears to be doing the right thing:

```
 -- Found cuFile: /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/libcufile.so
  -- Found CUDAToolkit: /opt/conda/conda-bld/_build_env/targets/x86_64-linux/include (found version "12.2.140")
  -- Found cuFile Batch API: TRUE
  -- Found cuFile Stream API: TRUE
```

Same story with [one of the CUDA 11.8 `x86_64` GHA jobs]( https://github.com/rapidsai/kvikio/actions/runs/9084882174/job/24967129706?pr=379#step:7:618 ):

```
  -- Found cuFile: /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/libcufile.so
  -- Found CUDAToolkit: /usr/local/cuda/targets/x86_64-linux/include (found version "11.8.89")
  -- Found cuFile Batch API: TRUE
  -- Found cuFile Stream API: FALSE
```

Note that cuFile's Stream API was added in CUDA 12.2. So it is expected to be disabled in CUDA 11.8

Admittedly haven't checked all of the jobs. So could be missing things

madsbk:
It looks like cmake dosn't run [`FINAL_CODE_BLOCK` ](https://github.com/rapidsai/kvikio/blob/branch-24.06/cpp/CMakeLists.txt#L130-L190) when building the Python library. 

Reverting https://github.com/rapidsai/kvikio/pull/342 fixes the issue.

KyleFromNVIDIA:
Local testing seems to indicate that this fix works. Ready for review.

cpp/CMakeLists.txt
```diff
@@ -84,6 +84,22 @@ endif()
 add_library(kvikio INTERFACE)
 add_library(kvikio::kvikio ALIAS kvikio)
 
+# Enable supported cuFile features in KvikIO
```
bdice, line 87: I think it would be helpful to provide some comments on how this section differs from the final code block below. It makes sense to me right now, but it will be hard to remember the intention of each piece of code later on.

KyleFromNVIDIA: I've added a thorough explanation.

bdice (approved):
Wonderful! Thank you for the additional comments.

robertmaynard (approved):

vyasr:
Something seems off here. I agree that the changes are good for FetchContent/add_subdirectory like use cases, but IIUC we were observing issues with conda Python builds which should be finding libkvikio's config, right? Why aren't they?

bdice:
> Something seems off here. I agree that the changes are good for FetchContent/add_subdirectory like use cases, but IIUC we were observing issues with conda Python builds which should be finding libkvikio's config, right? Why aren't they?

This was discussed offline and is being fixed in https://github.com/rapidsai/kvikio/pull/381.

PR #382, opened 2024-05-21
Set default value for PARALLEL_LEVEL to nproc

jjacobelli:
The `PARALLEL_LEVEL` variable is used in CI to set parallel level while building. Currently this variable is set on the GHA runners to `nproc`, but as this is a RAPIDS specific variable, this variable will soon be removed from the runners, so RAPIDS projects must explicitly set this variable to `nproc`. This PR is setting the default value of `PARALLEL_LEVEL` to `nproc` to match what is done by other RAPIDS projects


KyleFromNVIDIA (approved):
Approved, see my question at https://github.com/rapidsai/rmm/pull/1566#discussion_r1608465529 which also applies here.

jjacobelli:
Closed in favor of https://github.com/rapidsai/gha-tools/pull/108

PR #385, opened 2024-05-28
remove unnecessary 'setuptools' dependency

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/62.

It looks like this some of the `kvikio` conda package recipe has an unnecessary dependency on `setuptools`. I suspect that's left over from before the project was cut over to `scikit-build-core`.

This proposes removing it.

## Notes for Reviewers

How I confirmed there were no direct uses of `setuptools` in  this project:

```shell
git grep -i setuptools
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,6 @@ The Python packages depends on the following packages:
 
 * Cython
 * Pip
-* Setuptools
 
 For testing:
 * pytest
--- a/conda/recipes/kvikio/meta.yaml
+++ b/conda/recipes/kvikio/meta.yaml
@@ -54,7 +54,6 @@ requirements:
     - {{ stdlib("c") }}
   host:
     - python
-    - setuptools
     - pip
     - cython >=3.0.0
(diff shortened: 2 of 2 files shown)


madsbk:
There is a reference to setuptools here: https://github.com/rapidsai/kvikio/blob/branch-24.08/ci/release/update-version.sh#L27

jameslamb:
> There is a reference to setuptools here: 

Thanks for checking!

I believe those `update-version.sh` scripts are run manually by ops when cutting new release branches (@raydouglass or @AyodeAwe could confirm), and that they don't rely on projects' dependency metadata (conda recipes, `pyproject.toml`, `dependencies.yaml`) to provide it.

jameslamb:
Thanks for your attention to detail @madsbk ... based on @AyodeAwe 's 👍🏻 on https://github.com/rapidsai/kvikio/pull/385#issuecomment-2137468368, I think this is safe to merge and the `setuptools` reference in `update-version.sh` should remain where it is.

PR #389, opened 2024-06-05
use rapids-build-backend

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/31
Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89

Proposes introducing `rapids-build-backend` as this project's build backend, to reduce the complexity of various CI/build scripts.

### What about the ongoing work to add wheels?

If #369 is merged before this, this should be updated to ensure it doesn't break wheel builds.

But opening this one separately so we can keep making forward progress on rolling out `rapids-build-backend` across RAPIDS.


bdice (approved):
Looks good to me. Thank you!

PR #397, opened 2024-06-27
Avoid --find-links in wheel jobs

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/69.

Proposes a stricter pattern for passing a `libkvikio` wheel from the `wheel-build-cpp` job that produced it into the `wheel-build-python` job wanting to use it (as a build dependency of `kvikio`). This change improves the likelihood that issues with the `libkivkio` wheels will be caught in CI.

## Notes for Reviewers

See https://github.com/rapidsai/rmm/pull/1586 and https://github.com/rapidsai/build-planning/issues/69#issuecomment-2174556118 for details on how I tested this approach.

I didn't propose this on #369 because I hadn't quite finished testing the approach yet. But I do think we should do this for all of the C++ wheels (https://github.com/rapidsai/build-planning/issues/33).

### DIFF
--- a/ci/build_wheel_python.sh
+++ b/ci/build_wheel_python.sh
@@ -17,7 +17,14 @@ CPP_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="libkvikio_${RAPIDS_PY_CUDA_SUFFIX}" rapid
 
 cd "${package_dir}"
 
-python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check --find-links ${CPP_WHEELHOUSE}
+# ensure 'kvikio wheel builds always use the 'libkvikio' just built in the same CI run
+#
+# using env variable PIP_CONSTRAINT is necessary to ensure the constraints
+# are used when creating the isolated build environment
+echo "libkvikio-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${CPP_WHEELHOUSE}/libkvikio_*.whl)" > ./constraints.txt
+
+PIP_CONSTRAINT="${PWD}/constraints.txt" \
+    python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
 
 mkdir -p final_dist
 python -m auditwheel repair -w final_dist dist/*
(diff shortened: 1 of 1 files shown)


ci/build_wheel_python.sh
```diff
@@ -17,7 +17,14 @@ CPP_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="libkvikio_${RAPIDS_PY_CUDA_SUFFIX}" rapid
...
 
-python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check --find-links ${CPP_WHEELHOUSE}
+# ensure 'kvikio' wheel builds always use the 'libkvikio' just built in the same CI run
+#
+# using env variable PIP_CONSTRAINT is necessary to ensure the constraints
+# are used when creating the isolated build environment
+echo "libkvikio-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${CPP_WHEELHOUSE}/libkvikio_*.whl)" > ./constraints.txt
+
+PIP_CONSTRAINT="${PWD}/constraints.txt" \
+    python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
```
jakirkham, line 27: Am curious whether just adding `--no-index` would be sufficient for us

```suggestion
python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check --no-index --find-links ${CPP_WHEELHOUSE}
```

jameslamb: Thanks for the suggestion! I hadn't considered this.

But I wouldn't expect that to be sufficient here. We're building with build isolation, and this `${CPP_WHEELHOUSE}` directory doesn't contain wheels for other build dependencies like `scikit-build-core` and `rapids-build-backend`.

So to use `--no-index --find-links ${CPP_WHEELHOUSE}` we'd have to also do something like `pip download` the other build dependencies into that local directory. I think the approach with constraints is simpler for as long as we want to use build isolation (which, per https://github.com/rapidsai/build-planning/issues/69#issuecomment-2187054141, might not be fore long).

jakirkham: Discussed offline with James. We decided James will write up an issue for follow up and we will skip this suggestion in this PR

jameslamb: Thanks for talking through it with me. Here's the issue: https://github.com/rapidsai/build-planning/issues/79

msarahan (approved):

PR #398, opened 2024-07-02
Use verify-alpha-spec hook

KyleFromNVIDIA:
With the deployment of rapids-build-backend, we need to make sure our dependencies have alpha specs.

Contributes to https://github.com/rapidsai/build-planning/issues/31


jameslamb (approved):
Thanks. For this and all the other similar PRs you'll open across RAPIDS, can you please link them back to https://github.com/rapidsai/build-planning/issues/31 by mentioning that issue in the description?

Issue #383, opened 2024-05-22
MacOS doesn't have O_DIRECT, so I commented out this line in cpp/include/kvikio/file_handle.hpp:


qkoziol:
> > MacOS doesn't have O_DIRECT, so I commented out this line in cpp/include/kvikio/file_handle.hpp:
> > // if (o_direct) { file_flags |= O_DIRECT; }
> > After that, ./build.sh libkvikio worked :-)
> 
> @qkoziol, could you please open a separate issue for this so we don't lose track.

Yes, will do

_Originally posted by @qkoziol in https://github.com/rapidsai/kvikio/issues/373#issuecomment-2125670944_

PR #404, opened 2024-07-15
Use `#ifdef` to guard `O_DIRECT` usage

jakirkham:
Fixes https://github.com/rapidsai/kvikio/issues/383


jakirkham:
@qkoziol could you please try this and let us know if that addresses the issue?

jakirkham:
AIUI `O_DIRECT` is only needed with GDS. However if we do need equivalent behavior on other platforms, there are [other options]( https://github.com/facebook/rocksdb/wiki/Direct-IO ), but it may involve reworking the code here more substantially

cpp/include/kvikio/file_handle.hpp
```diff
@@ -66,7 +66,13 @@ inline int open_fd_parse_flags(const std::string& flags, bool o_direct)
     default: throw std::invalid_argument("Unknown file open flag");
   }
   file_flags |= O_CLOEXEC;
-  if (o_direct) { file_flags |= O_DIRECT; }
+  if (o_direct) {
+    #ifdef O_DIRECT
+    file_flags |= O_DIRECT;
+    #else
+    throw std::invalid_argument("'O_DIRECT' flag unsupported on this platform");
```
jakirkham, line 73: Currently this errors if `o_direct` is set and `O_DIRECT` is unavailable. Though another option would be to warn or simply silently suppress it

cpp/include/kvikio/file_handle.hpp
```diff
@@ -84,7 +91,21 @@ inline int open_fd(const std::string& file_path,
...
-  int fd = ::open(file_path.c_str(), open_fd_parse_flags(flags, o_direct), mode);
+  int fd = -1;
+  if (o_direct) {
+#if defined(O_DIRECT)
+    // On Linux, use `O_DIRECT`
+    fd = ::open(file_path.c_str(), open_fd_parse_flags(flags, true), mode);
+#elif defined(F_NOCACHE)
+    // On macOS, pass `F_NOCACHE` to `fcntl` after opening file
+    fd = ::open(file_path.c_str(), open_fd_parse_flags(flags, false), mode);
+    if (fd != -1) { fcntl(fd, F_NOCACHE, 1); }
```
jakirkham, line 102: Added similar logic for macOS. Would be good to check whether we need this

madsbk (approved):
Thanks @jakirkham, looks good.

cpp/include/kvikio/file_handle.hpp
wence-: Perhaps @madsbk can explain what we want `O_DIRECT` for? Are we doing something magic in this scenario that means we don't want the kernel filesystem layer handling caching?

madsbk: Yes, or rather, cuFile does some magic.

wence-: Since cufile is not supported on macos, we therefore do not want this part on macos.

madsbk: Good point, instead of this, we could have `detail::open_fd()` fail if `O_DIRECT` isn't available and then fall back to compatibility mode: 
https://github.com/rapidsai/kvikio/blob/branch-24.08/cpp/include/kvikio/file_handle.hpp#L167-L171

jakirkham:
Is this superseded by PR ( https://github.com/rapidsai/kvikio/pull/410 )?

madsbk:
> Is this superseded by PR ( #410 )?

No, we still need to guard the `O_DIRECT` usage. 
I pushed a change to enable compat mode when `open_fd()` fails

cpp/include/kvikio/file_handle.hpp
```diff
@@ -66,7 +67,13 @@ inline int open_fd_parse_flags(const std::string& flags, bool o_direct)
     default: throw std::invalid_argument("Unknown file open flag");
   }
   file_flags |= O_CLOEXEC;
-  if (o_direct) { file_flags |= O_DIRECT; }
+  if (o_direct) {
+#if defined(O_DIRECT)
+    file_flags |= O_DIRECT;
+#else
+    throw std::invalid_argument("'o_direct' flag unsupported on this platform");
```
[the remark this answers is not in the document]
madsbk: added

jakirkham:
Thanks Mads and Lawrence! 🙏

cpp/include/kvikio/file_handle.hpp
```diff
@@ -66,7 +70,13 @@ inline int open_fd_parse_flags(const std::string& flags, bool o_direct)
     default: throw std::invalid_argument("Unknown file open flag");
   }
   file_flags |= O_CLOEXEC;
-  if (o_direct) { file_flags |= O_DIRECT; }
+  if (o_direct) {
+#if defined(O_DIRECTT)
```
jakirkham, line 74: Looks like a small typo snuck in here. Fixing in PR: https://github.com/rapidsai/kvikio/pull/416

```suggestion
#if defined(O_DIRECT)
```

Issue #339, opened 2024-02-10
Support aarch64 + libcufile


bdice:
As of CUDA 12.2, there is a libcufile library for aarch64. In #328, we updated to CUDA 12.2 but decided to hold off on changing behavior around libcufile on aarch64. Because kvikio dlopens libcufile, it may be possible to have a runtime check that disables libcufile support on aarch64 with CUDA < 12.2 (or if the library is not installed). I haven't investigated this very far, but wanted to open this issue for tracking purposes.

See also: https://github.com/rapidsai/kvikio/pull/328#pullrequestreview-1873354307

We wouldn't be able to add `libcufile` as a hard dependency on aarch64 until dropping CUDA 12.0 / 12.1, but it should be possible to build a package that supports libcufile if it is installed by the user.

cc: @vyasr, thanks for the proposal. This sounds plausible.

vyasr:
Thanks for opening the issue Bradley. I'm pretty sure this is the right long-term answer here, but it's not a blocker for us getting CUDA 12.2 done.

Issue #393, opened 2024-06-13
Incompatibility with ARM


Marcus-M1999:
Hello,

It seems that there is an incompatibility with the ARM architecture. Is there anyway that this could be fixed? I was able to get this repo running by having a colleague build from source, however conda install from the stable and nightly repo doesn't work.
Specs:
 NVIDIA-SMI 550.54.15              Driver Version: 550.54.15      CUDA Version: 12.4 GPU: A6000

madsbk:
What errors are  you seeing? 
We do run CI tests on ARM; it is supposed to work :)

Marcus-M1999:
The kvikio.defaults.compat_mode() returns true even though the cufile.json is set to false. I tested the read & write times both are slower than or equal to the time it takes to save a CPU array. I'm testing by saving a sin wave in a cupy array with the blow code:
`cp.sin(cp.arange(0, 100000000, 0.1))`

jakirkham:
Is this still an issue after our offline discussion, Marcus?

Marcus-M1999:
The code seems to be working, there is a slight speed up for this workload. The only "bug" I would say is that we could only get it working after building the library from source. When trying to install via pip or conda it wouldn't work properly. I'm not sure what the issue could be, if it's working on your system then I'll try to dig deeper into the issue on my IGX.

jakirkham:
It might be worth retrying the pip & conda installs. AIUI there were a number of changes that occurred to the system between initially testing these and switching to building from source to test. It would be good to determine whether the system changes had any effect on the binary installs

jakirkham:
In 24.08 we enabled cuFile on ARM ( https://github.com/rapidsai/kvikio/pull/405 ), which was previously disabled

Does that help?

PR #405, opened 2024-07-16
Enable cuFile on ARM

jakirkham:
Fixes https://github.com/rapidsai/kvikio/issues/339

Builds CUDA 12.2+ ARM packages with cuFile and adds cuFile as a dependency. Continues to build CUDA 11 ARM packages without cuFile.

As cuFile already has the needed logic to `dlopen` cuFile at runtime (with fallbacks for when it is missing), there shouldn't be any additional handling needed for legacy CUDA versions with ARM support.


dependencies.yaml
```diff
@@ -207,12 +207,18 @@ dependencies:
         matrices:
           - matrix:
               arch: aarch64
+              cuda: "12.*"
```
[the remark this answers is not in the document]
jakirkham: Agreed. Had tried this before, but ran into some issues. Though this may have been caused by other changes. Appears to work now

madsbk (approved):
Nice, thanks @jakirkham !

dependencies.yaml
```diff
@@ -4,7 +4,7 @@ files:
     output: conda
     matrix:
       cuda: ["11.8", "12.5"]
-      arch: [x86_64]
+      arch: [aarch64, x86_64]
```
bdice, line 7: I would prefer to revert this. We don't supply ARM environment files for most RAPIDS repos. If you'd like to add ARM conda envs for all of RAPIDS, I would do that in a separate PR and track the effort in rapidsai/build-planning for all repos.
```suggestion
      arch: [x86_64]
```

jakirkham: Given there are other users asking about ARM specifically with KvikIO (and running into issues without more explicit guidance), the environments provide a value add for them. For example: https://github.com/rapidsai/kvikio/issues/393

dependencies.yaml
```diff
@@ -207,12 +207,18 @@ dependencies:
...
+              cuda: "12.*"
             packages:
+              - libcufile-dev
           - matrix:
-              cuda: "12.*"
               arch: x86_64
+              cuda: "12.*"
             packages:
               - libcufile-dev
+          - matrix:
```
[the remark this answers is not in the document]
msarahan: I see, I didn't realize that we had to be so explicit. Thanks for clarifying that.

msarahan (approved):

jakirkham:
Thanks all! 🙏

Anything else needed here?

dependencies.yaml
bdice: I suppose it's fine to include this. Maybe this should be done for all of RAPIDS, with a build-planning issue for tracking.

We don't have a good way to exclude (CUDA 11 + ARM), unless you create a separate file key entry for (CUDA 12 + ARM) That's not worth doing imo.

As we see increased ARM usage, it'll be somewhat helpful to have these environment files readily accessible. I have had to explain how to generate the ARM environments with `rapids-dependency-file-generator` before and it's a little complicated for new DFG users.

bdice (approved):

dependencies.yaml
jakirkham: Agree this is a good idea

Have raised issue: https://github.com/rapidsai/build-planning/issues/87

Please add anything additional to that

PR #408, opened 2024-07-22
Fetch and use BS thread_pool 

madsbk:
Instead of hard coding the header of [`BS::thread_pool`](https://github.com/bshoshany/thread-pool?tab=readme-ov-file), we now fetch the implementation. Done for cudf in https://github.com/rapidsai/cudf/pull/16210.


cpp/cmake/rapids_config.cmake
```diff
@@ -29,7 +29,7 @@ endif()
 if(NOT EXISTS "${CMAKE_CURRENT_BINARY_DIR}/KVIKIO_RAPIDS-${RAPIDS_VERSION_MAJOR_MINOR}.cmake")
   file(
     DOWNLOAD
-    "https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-${RAPIDS_VERSION_MAJOR_MINOR}/RAPIDS.cmake"
+    "https://raw.githubusercontent.com/KyleFromNVIDIA/rapids-cmake/bs-thread-pool/RAPIDS.cmake"
```
KyleFromNVIDIA, line 32: This is a temporary change to test https://github.com/rapidsai/rapids-cmake/pull/651.

robertmaynard: The setting of `rapids-cmake-repo` and `rapids-cmake-branch` need to occur outside of the conditional. Those variables are evaluated by `RAPIDS.cmake` on each run of cmake.

KyleFromNVIDIA: Fixed

bdice (approved):

PR #410, opened 2024-07-23
Don't open with `O_DIRECT` when in compat mode

madsbk:
Currently, `FileHandle` always opens the file twice - one with and one without the `O_DIRECT`  flag.
Now, when in compatibility mode, `FileHandle` only opens the file once (without the `O_DIRECT`  flag) .

### DIFF
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -165,13 +165,17 @@ class FileHandle {
       _compat_mode{compat_mode}
   {
     if (!_compat_mode) {
-      try {
-        _fd_direct_on = detail::open_fd(file_path, flags, true, mode);
-      } catch (const std::system_error&) {
-        _compat_mode = true;  // Fall back to compat mode if we cannot open the file with O_DIRECT
-      }
+      return;  // Nothing to do in compatibility mode
     }
 
+    // Try to open the file with the O_DIRECT flag. Fall back to compatibility mode, if it fails.
+    try {
+      _fd_direct_on = detail::open_fd(file_path, flags, true, mode);
+    } catch (const std::system_error&) {
+      _compat_mode = true;
+    }
+
+    // Create a cuFile handle, if not in compatibility mode
     if (!_compat_mode) {
       CUfileDescr_t desc{};  // It is important to set to zero!
       desc.type = CU_FILE_HANDLE_TYPE_OPAQUE_FD;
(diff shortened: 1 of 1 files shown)


bdice (approved):

wence- (approved):

cpp/include/kvikio/file_handle.hpp
```diff
@@ -164,12 +164,18 @@ class FileHandle {
...
+    // Try to open the file with the O_DIRECT flag. Fall back to compatibility mode, if it fails.
     try {
       _fd_direct_on = detail::open_fd(file_path, flags, true, mode);
     } catch (const std::system_error&) {
-      _compat_mode = true;  // Fall back to compat mode if we cannot open the file with O_DIRECT
+      _compat_mode = true;
     }
 
+    // Create a cuFile handle, if not in compatibility mode
     if (!_compat_mode) {
```
vuule, line 179: no need for this condition now that we return early in compat mode.

madsbk: Notice, if `_fd_direct_on = detail::open_fd(file_path, flags, true, mode);` fails, we set `_compat_mode = true` thus `_compat_mode`  might be `true` even though it was `false` initially

vuule (approved):

fstrug:
Would it be possible to add the ability for users to define whether when in compatibility mode the POSIX API is using the O_DIRECT flag or not? We want to evaluate the performance between kvikio's GDS and POSIX API, and it'd be useful to determine how caching might effect performance. Especially when trying to evaluate performance on an average of multiple runs ran on the same filepath.

jakirkham:
We still take a `mode` argument. Is it possible for you to pass `O_DIRECT` by `|` with the `mode`?

fstrug:
How would you do this? All I can see is 

https://github.com/rapidsai/kvikio/blob/007c0671004c33c46635c345060e11dcc612c26c/python/kvikio/kvikio/cufile.py#L84-L88

So like "rb | O_DIRECT"?

Also, maybe this doc string needs some clarification wrt this PR?
https://github.com/rapidsai/kvikio/blob/007c0671004c33c46635c345060e11dcc612c26c/python/kvikio/kvikio/cufile.py#L75-L77

jakirkham:
Ah I see. Yeah we don't expose this to Python currently. Could you please raise an issue about this?

PR #413, opened 2024-07-24
cufile-wrapper: include the cuda-wrapper

madsbk:
Fix the cufile-wrapper when `cuda.h` isn't available


qkoziol (approved):
Fixes the CPU-only build for legate.io

jakirkham (approved):
Thanks Mads! 🙏

PR #424, opened 2024-07-25
Benchmarks clean up

### LINKED ISSUE #176: Moving from `pynvml` to `nvidia-ml-py`
Currently this is making use of `pynvml` in a few places:

https://github.com/rapidsai/kvikio/blob/c152f633779d8c41791a3b76f9f678f926a9847d/python/benchmarks/single-node-io.py#L248-L250

However we would like to move to `nvidia-ml-py` in the future. Raising this issue to track this work

madsbk:
Closes #176

Since we didn't install the python benchmarks previously, this PR is non-breaking even though it moves and renames them.


jameslamb (approved):

wence- (approved):
Looks sane.

PR #426, opened 2024-07-31
[WIP] S3 support

madsbk:
**NB**: not ready for review, still very much prototyping.

Implements AWS S3 read support. 

Tasks
- [ ] Make the dependency of AWS SDK optional
- [ ] Make a shim layer
- [ ] Test with a libcudf io source 
- [ ] Docs


cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,327 @@
...
+    const std::size_t n = outcome.GetResult().GetContentLength();
+    if (n != size) {
+      throw std::runtime_error("S3 read of " + std::to_string(size) + " bytes failed, received " +
+                               std::to_string(n) + " bytes");
+    }
+    auto t1   = std::chrono::high_resolution_clock::now();
+    float dur = std::chrono::duration_cast<std::chrono::microseconds>(t1 - t0).count() / 1000000.0;
+    std::cout << "RemoteHandle::read_to_host() - buf: " << buf << ", size: " << size
+              << ", file_offset: " << file_offset << ", bw: " << size / dur / (2 << 20) << " MiB/s"
+              << std::endl;
```
madsbk, line 263: Remove before PR merging

python/kvikio/tests/test_aws_s3.py
```diff
@@ -0,0 +1,137 @@
...
+import socket
+import time
+from contextlib import contextmanager
+
+import pytest
+
+import kvikio
+
+# TODO: remove before PR merge. Trigger CI error if the remote module wasn't built
+import kvikio._lib.remote_handle  # isort: skip
```
madsbk, line 15: [debug] remove before PR merging

KyleFromNVIDIA:
Built the C++ wheel locally with https://github.com/rapidsai/ci-imgs/pull/168, and confirmed that it found `aws-sdk-cpp`, though I did not build the Python wheel or run the tests.

KyleFromNVIDIA:
https://github.com/rapidsai/ci-imgs/pull/168 has just been merged. This will be ready to test again once the images are rebuilt.

python/kvikio/tests/test_aws_s3.py
```diff
@@ -0,0 +1,137 @@
...
+
+moto = pytest.importorskip("moto", minversion="3.1.6")
+boto3 = pytest.importorskip("boto3")
```
KyleFromNVIDIA, line 18: I don't believe these should be `importorskip`. I understand the intention of not running the tests if S3 support is not available, but if it is available, these should be required.

python/kvikio/tests/test_aws_s3.py
```diff
@@ -0,0 +1,137 @@
...
+
+if not kvikio.is_remote_file_available():
+    pytest.skip(
+        "cannot test remote IO, please build KvikIO with with AWS S3 support",
+        allow_module_level=True,
+    )
+
+ThreadedMotoServer = pytest.importorskip("moto.server").ThreadedMotoServer
```
KyleFromNVIDIA, line 26: Same here, make this a regular import.

KyleFromNVIDIA (changes requested):

python/kvikio/tests/test_aws_s3.py
madsbk: I like to use `pytest.importorskip` to allow users run pytest without optional packages installed. 
However, we could introduce (in another PR) a dedicated test that tries to include all optional packages and fails if any of them isn't available.

KyleFromNVIDIA: Sure. My point is that in this particular case, if S3 support is present (as determined by `is_remote_file_available()` above), then `moto` and `boto3` should be required to test it. The test itself is optional, but if we're going to run it, then `moto` and `boto3` are required. (Note that they've both been added as optional dependencies in the `test` category already.)

dependencies.yaml
```diff
@@ -319,6 +320,14 @@ dependencies:
           - &dask dask>=2022.05.2
           - pytest
           - pytest-cov
+          - boto3>=1.21.21
+      - output_types: [requirements, pyproject]
+        packages:
+          - moto[server]>=4.0.8
+      - output_types: conda
+        packages:
+          - cudf==24.10.*,>=0.0.0a0
```
KyleFromNVIDIA, line 329: Note that CI is currently failing due to the `cudf` tests running without https://github.com/rapidsai/cudf/pull/16499 being merged yet. The other PR can't be merged until this one is. I think the way to resolve this is to temporarily remove the `cudf` dependency here and below, merge this PR, merge the other one, and then add the `cudf` dependency back.

python/kvikio/tests/test_aws_s3.py
madsbk: I am not sure, what if a user build/install kvikio with s3 support but doesn’t have `moto` or `boto3`? In this case, I just want pytest to skip the tests. The same goes for installations without `cupy`, `numpy`, or `zarr`.

dependencies.yaml
madsbk: Sounds good

python/kvikio/tests/test_aws_s3.py
KyleFromNVIDIA: Good point on that scenario. Not how I'd do it, but I suppose it's reasonable. I'll leave it to you to decide.

dependencies.yaml
KyleFromNVIDIA: I went ahead and commented it out with a note.

conda/recipes/libkvikio/meta.yaml
```diff
@@ -72,8 +74,10 @@ outputs:
     requirements:
       build:
         - cmake {{ cmake_version }}
+        - aws-sdk-cpp
```
madsbk, line 77:
```suggestion
```
@KyleFromNVIDIA do we `aws-sdk-cpp` here? Asking because `libcufile-dev` isn't need?

conda/recipes/libkvikio/meta.yaml
```diff
@@ -43,6 +43,7 @@ requirements:
     {% endif %}
     - ninja
     - {{ stdlib("c") }}
+    - aws-sdk-cpp
```
madsbk, line 46:
```suggestion
```
@KyleFromNVIDIA do we `aws-sdk-cpp` here? Asking because `libcufile-dev` isn't need?

python/kvikio/tests/test_aws_s3.py
bdice: Using `importorskip` like that makes our test suites prone to accidentally skipping tests. I tend to lean the same way as Kyle here: it is reasonable for the test suites to have dependencies, and we require those for the whole test suite to pass. Users missing a test dependency can skip the affected tests manually or install the needed packages if they’re looking for a green/passing result.

conda/recipes/libkvikio/meta.yaml
```diff
@@ -72,8 +74,10 @@ outputs:
     requirements:
       build:
         - cmake {{ cmake_version }}
+        - aws-sdk-cpp
       host:
         - cuda-version ={{ cuda_version }}
+        - aws-sdk-cpp
```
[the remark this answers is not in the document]
KyleFromNVIDIA: `libcufile-dev` is present under the `cuda` depencency set. I'm not sure of the best way to organize this. @vyasr any thoughts on this?

python/kvikio/tests/test_aws_s3.py
madsbk: OK, I see your points, let's make them `import`

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,327 @@
...
+    req.SetRange(byte_range.c_str());
+
+    // To write directly to `buf`, we register a "factory" that wraps a buffer as a output stream.
+    Aws::Utils::Stream::PreallocatedStreamBuf buf_stream(static_cast<unsigned char*>(buf), size);
+    req.SetResponseStreamFactory(
+      [&]() { return Aws::New<detail::BufferAsStream>("BufferAsStream", &buf_stream); });
+
+    Aws::S3::Model::GetObjectOutcome outcome = default_context.client().GetObject(req);
+    if (!outcome.IsSuccess()) {
+      const Aws::S3::S3Error& err = outcome.GetError();
```
mhaseeb123, line 251: _nit:_ let's use east consts throughout in this file.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,327 @@
...
+      throw std::runtime_error(err.GetExceptionName() + ": " + err.GetMessage());
+    }
+    const std::size_t n = outcome.GetResult().GetContentLength();
+    if (n != size) {
+      throw std::runtime_error("S3 read of " + std::to_string(size) + " bytes failed, received " +
+                               std::to_string(n) + " bytes");
+    }
+    auto t1   = std::chrono::high_resolution_clock::now();
+    float dur = std::chrono::duration_cast<std::chrono::microseconds>(t1 - t0).count() / 1000000.0;
```
mhaseeb123, line 260:
```suggestion
    float const dur = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::high_resolution_clock::now() - t0).count() / 1000000.0;
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,327 @@
...
+   *
+   * @param buf Pointer to host memory.
+   * @param size Number of bytes to read.
+   * @param file_offset File offset in bytes.
+   * @return Number of bytes read.
+   */
+  std::size_t read_to_host(void* buf, std::size_t size, std::size_t file_offset = 0)
+  {
+    KVIKIO_NVTX_FUNC_RANGE("AWS S3 receive", size);
+    auto t0 = std::chrono::high_resolution_clock::now();
```
mhaseeb123, line 234:
```suggestion
    auto const t0 = std::chrono::high_resolution_clock::now();
```

python/kvikio/tests/test_aws_s3.py
```diff
@@ -0,0 +1,137 @@
...
+import kvikio
+
+# TODO: remove before PR merge. Trigger CI error if the remote module wasn't built
+import kvikio._lib.remote_handle  # isort: skip
+
+if not kvikio.is_remote_file_available():
+    pytest.skip(
+        "cannot test remote IO, please build KvikIO with with AWS S3 support",
+        allow_module_level=True,
+    )
```
wence-, line 21:
```suggestion
pytestmark = pytest.mark.skipif(
    not kvikio.is_remote_file_available(), 
    reason="cannot test remote IO, please build KvikIO with with AWS S3 support",
)
```

I think?

wence- (changes requested):

cpp/include/kvikio/file_handle.hpp
```diff
@@ -286,7 +286,7 @@ class FileHandle {
    *
    * @return The number of bytes
    */
-  [[nodiscard]] inline std::size_t nbytes() const
+  [[nodiscard]] std::size_t nbytes() const
```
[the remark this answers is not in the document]
madsbk: Yes :)

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+  std::size_t read_to_host(void* buf, std::size_t size, std::size_t file_offset = 0)
+  {
+    KVIKIO_NVTX_FUNC_RANGE("AWS S3 receive", size);
+
+    auto& default_context = detail::S3Context::default_context();
+    Aws::S3::Model::GetObjectRequest req;
+    req.SetBucket(_bucket_name.c_str());
+    req.SetKey(_object_name.c_str());
+    const std::string byte_range =
+      "bytes=" + std::to_string(file_offset) + "-" + std::to_string(file_offset + size - 1);
```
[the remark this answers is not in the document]
madsbk: Heh, yeah very strange API :)

Don't you think string stream is a bit overkill here?

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+    req.SetRange(byte_range.c_str());
+
+    // To write directly to `buf`, we register a "factory" that wraps a buffer as a output stream.
+    Aws::Utils::Stream::PreallocatedStreamBuf buf_stream(static_cast<unsigned char*>(buf), size);
+    req.SetResponseStreamFactory(
+      [&]() { return Aws::New<detail::BufferAsStream>("BufferAsStream", &buf_stream); });
```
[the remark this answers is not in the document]
madsbk: The allocation comes from the user `(void* buf)`, this is just a wrapper to make the S3 SDK write directly to `buf`.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+    std::size_t cur_file_offset = convert_size2off(file_offset);
+    std::size_t byte_remaining  = convert_size2off(size);
+
+    while (byte_remaining > 0) {
+      const std::size_t nbytes_requested = std::min(posix_bounce_buffer_size, byte_remaining);
+      std::size_t nbytes_got             = nbytes_requested;
+      nbytes_got = read_to_host(alloc.get(), nbytes_requested, cur_file_offset);
+      CUDA_DRIVER_TRY(cudaAPI::instance().MemcpyHtoDAsync(devPtr, alloc.get(), nbytes_got, stream));
+      CUDA_DRIVER_TRY(cudaAPI::instance().StreamSynchronize(stream));
```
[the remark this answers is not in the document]
madsbk: Yes

python/kvikio/kvikio/remote_file.py
```diff
@@ -0,0 +1,124 @@
...
+            Name of the bucket.
+        object_name
+            Name of the object.
+        """
+        self._handle = _get_remote_remote_file_class().from_bucket_and_object(
+            bucket_name, object_name
+        )
+
+    @classmethod
+    def from_url(cls, url: str) -> RemoteFile:
```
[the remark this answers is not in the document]
madsbk: KvikIO doesn't depend on typing_extensions, let's do this in another PR.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+    const char* ep = getenv("AWS_ENDPOINT_URL");
+    if (ep != nullptr) { clientConfig.endpointOverride = ep; }
+
+    // We check authentication here to trigger an early exception.
+    auto provider = Aws::MakeShared<Aws::Auth::DefaultAWSCredentialsProviderChain>("check-creds");
+    auto creds    = provider->GetAWSCredentials();
+    if (creds.IsEmpty()) {
+      throw std::runtime_error(std::string("Failed authentication to ") + ep);
+    }
+    return std::make_shared<Aws::S3::S3Client>(Aws::S3::S3Client(clientConfig));
```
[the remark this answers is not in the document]
madsbk: It does not, fixed :)

cpp/include/kvikio/remote_handle.hpp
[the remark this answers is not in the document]
wence-: I meant the heap allocation of the `BufferAsStream` object (via `Aws::New`).

cpp/include/kvikio/remote_handle.hpp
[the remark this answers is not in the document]
wence-: I guess it doesn't matter for this, though not that `std::string::operator+` has quadratic behaviour.

cpp/include/kvikio/remote_handle.hpp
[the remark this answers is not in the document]
madsbk: I assume AWS will call `Aws::Free`?  I cannot find anything in the docs, but all `SetResponseStreamFactory` uses `Aws::New`. The call back must return `Aws::IOStream*`.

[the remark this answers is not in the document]
wence-: Thanks, it appears yes, see https://github.com/aws/aws-sdk-cpp/blob/c54ac9bf020230619d16f33713d33a2e64a840fe/src/aws-cpp-sdk-core/source/utils/stream/ResponseStream.cpp#L78

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -177,35 +171,50 @@ class RemoteHandle {
...
   RemoteHandle() noexcept = default;
 
+  /**
+   * @brief Construct from a bucket and object name pair.
+   *
+   * @param context The S3 context used for the connection to the remove server.
+   * @param bucket_and_object_name Name pair <bucket, object>.
+   */
+  RemoteHandle(std::shared_ptr<S3Context> context,
+               std::pair<std::string, std::string> bucket_and_object_name)
```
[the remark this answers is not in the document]
madsbk: I did this at first, but since we need the `if (!_context) { throw ...; }` check, I think it is easier to read when everything is in the body.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+  /**
+   * @brief Get the default context, which is created on first call.
+   *
+   * @return The default S3 context.
+   */
+  static S3Context& default_context()
+  {
+    static S3Context _default_context;
+    return _default_context;
```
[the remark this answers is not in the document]
madsbk: Removed

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+ * @param object_name The object name.
+ * @return Size of the file in bytes.
+ */
+inline std::size_t get_s3_file_size(const std::string& bucket_name, const std::string& object_name)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  Aws::S3::Model::HeadObjectRequest req;
+  req.SetBucket(bucket_name.c_str());
+  req.SetKey(object_name.c_str());
+  Aws::S3::Model::HeadObjectOutcome outcome = S3Context::default_context().client().HeadObject(req);
```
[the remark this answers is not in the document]
madsbk: Done

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+   *
+   * @return The new client
+   */
+  static std::shared_ptr<Aws::S3::S3Client> create_client()
+  {
+    S3Context::ensure_aws_s3_api_init();
+
+    // Read AWS_ENDPOINT_URL to overwrite endpoint
+    Aws::Client::ClientConfiguration clientConfig;
+    const char* ep = getenv("AWS_ENDPOINT_URL");
```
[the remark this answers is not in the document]
madsbk: `S3Context` can now be created with an existing S3 client.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,319 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cstdlib>
+#include <iostream>
+#include <memory>
+#include <stdexcept>
+#include <string>
+
```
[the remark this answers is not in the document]
madsbk: Good point,  done

dependencies.yaml
```diff
@@ -106,6 +106,7 @@ dependencies:
         packages:
           - c-compiler
           - cxx-compiler
+          - aws-sdk-cpp
```
wence-, line 109: *question*: Should we set a version constraint on this package? I suspect yes.

cpp/CMakeLists.txt
```diff
@@ -55,6 +55,29 @@ rapids_find_package(
   INSTALL_EXPORT_SET kvikio-exports
 )
 
+# If AWSSDK isn't found, the Cython module remote_handle.pyx isn't built and C++ users shouldn't
+# include <kvikio/remote_handle.hpp>
+rapids_find_package(
```
wence-, line 60: *question*: Perhaps not for this prototype, but if it is easy, perhaps yes.

Should building S3 support be a user-configurable option. That is, even if we can find the package, should we have an option (say `-DKVIKIO_HAVE_S3_SUPPORT=1`) to cmake so that the user can compile without this dependency if they are only doing local I/O?

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+#include <aws/s3/model/GetObjectRequest.h>
+#include <aws/s3/model/HeadObjectRequest.h>
+
+#include <kvikio/parallel_operation.hpp>
+#include <kvikio/posix_io.hpp>
+#include <kvikio/utils.hpp>
+
+#ifndef KVIKIO_AWS_SDK_FOUND
+#error "cannot include <kvikio/remote_handle.hpp>, configuration did not find AWS SDK"
+#endif
```
wence-, line 38: *issue*: For the compiler to even get here if we didn't find the SDK, you need to gate the `#include <aws/...>` lines in `#ifdef KVIKIO_AWS_SDK_FOUND` I think.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+inline std::pair<std::string, std::string> parse_s3_path(const std::string& path)
+{
+  if (path.empty()) { throw std::invalid_argument("The remote path cannot be an empty string."); }
+  if (path.size() < 5 || path.substr(0, 5) != "s3://") {
+    throw std::invalid_argument("The remote path must start with the S3 scheme (\"s3://\").");
+  }
+  std::string p = path.substr(5);
+  if (p.empty()) { throw std::invalid_argument("The remote path cannot be an empty string."); }
+  size_t pos = p.find_first_of('/');
+  if (pos == 0) { throw std::invalid_argument("The remote path does not contain a bucket name."); }
```
wence-, line 69: OK, so if the string is `s3:///object`, then we raise. But if it's `s3://bucket` we return `(bucket, "")`.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+  explicit BufferAsStream(std::streambuf* buf) : Base(buf) {}
+
+  ~BufferAsStream() override = default;
+};
+
+/**
+ * @brief Given a file path like "s3://<bucket>/<object>", return the name of the bucket and object.
+ *
+ * @param path S3 file path.
+ * @return Pair of strings: [bucket-name, object-name].
```
wence-, line 58: *nit*: Document the handling for `s3://bucket` and `s3://bucket/` As well as `@throws std::invalid_argument`

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+ *  - The SDK for C++ and its dependencies use C++ static objects, and the order of static object
+ *    destruction is not determined by the C++ standard. To avoid memory issues caused by the
+ *    nondeterministic order of static variable destruction, do not wrap `S3Context` in another
+ *    static object.
+ *  - Please construct and destruct `S3Context` from the same thread (use a dedicated thread if
+ *    necessary). This avoids problems in initializing the dependent Common RunTime C libraries.
+ */
+class S3Context {
+ private:
+  // We use a shared point since constructing a default `Aws::S3::S3Client` before calling
```
wence-, line 89:
```suggestion
  // We use a shared pointer since constructing a default `Aws::S3::S3Client` before calling
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+  // `Aws::InitAPI` is illegal.
+  std::shared_ptr<Aws::S3::S3Client> _client;
```
wence-, line 91: I guess we also need an owning pointer because in the case where we construct the S3Context from an endpoint, we need to keep the client alive.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+    if (!endpoint_override.empty()) {
+      config.endpointOverride = endpoint_override;
+    } else if (ep != nullptr && !std::string(ep).empty()) {
+      config.endpointOverride = ep;
+    }
+
+    // We check authentication here to trigger an early exception.
+    Aws::Auth::DefaultAWSCredentialsProviderChain provider;
+    if (provider.GetAWSCredentials().IsEmpty()) {
+      throw std::runtime_error("failed authentication to S3 server");
```
wence-, line 145: Mention in `@throws` in docstring?

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  /**
+   * @brief Get a reference to the S3 client.
+   *
+   * @return S3 client.
+   */
+  Aws::S3::S3Client& client() { return *_client; }
+
+  // No copy and move semantic
```
wence-, line 169: OK, so the user is always going to hand around const references to this?

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+
+/**
+ * @brief Handle of remote file (currently, only AWS S3 is supported).
+ */
+class RemoteHandle {
+ private:
+  std::string _bucket_name{};
+  std::string _object_name{};
+  std::size_t _nbytes{};
+  std::shared_ptr<S3Context> _context;
```
wence-, line 206: Can this be a non-owning `S3Context const & _context;` or does that cause too many problems?

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+  RemoteHandle(std::shared_ptr<S3Context> context, std::string bucket_name, std::string object_name)
+    : RemoteHandle(std::move(context),
+                   std::make_pair(std::move(bucket_name), std::move(object_name)))
+  {
+  }
+
+  /**
+   * @brief Construct from a remote path such as "s3://<bucket>/<object>".
+   *
+   * @param context The S3 context used for the connection to the remove server.
```
wence-, line 241:
```suggestion
   * @param context The S3 context used for the connection to the remote server.
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+  std::string _bucket_name{};
+  std::string _object_name{};
+  std::size_t _nbytes{};
+  std::shared_ptr<S3Context> _context;
+
+ public:
+  /**
+   * @brief Construct from a bucket and object name pair.
+   *
+   * @param context The S3 context used for the connection to the remove server.
```
wence-, line 212:
```suggestion
   * @param context The S3 context used for the connection to the remote server.
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,356 @@
...
+    _context     = std::move(context);
+    _bucket_name = std::move(bucket_and_object_name.first);
+    _object_name = std::move(bucket_and_object_name.second);
+    _nbytes      = _context->get_file_size(_bucket_name, _object_name);
+  }
+
+  /**
+   * @brief Construct from a bucket and object name.
+   *
+   * @param context The S3 context used for the connection to the remove server.
```
wence-, line 228:
```suggestion
   * @param context The S3 context used for the connection to the remote server.
```

wence- (commented):
Getting close I think, mostly small comments. Haven't gone through the python/cython yet...

cpp/CMakeLists.txt
madsbk: Good point, done

cpp/include/kvikio/remote_handle.hpp
madsbk: D'oh, fixed

cpp/include/kvikio/remote_handle.hpp
madsbk: Yeah, in this case the server would just response with object not found. 
But I have added an explicit check and added some tests

cpp/include/kvikio/remote_handle.hpp
madsbk: I think we can, it looks like all the relevant methods of `Aws::S3::S3Client ` is `const`.
But I think it is a bit too dangerous, particularly, for the Python crowd.

cpp/include/kvikio/remote_handle.hpp
madsbk: Heh, you don't like the _remove_ server?

cpp/include/kvikio/remote_handle.hpp
madsbk: Done, `remote_handle.hpp` is the first file in KvikIO to go east :)

cpp/include/kvikio/remote_handle.hpp
madsbk: yes

cpp/include/kvikio/remote_handle.hpp
madsbk: yes

dependencies.yaml
madsbk: Yes, we need https://github.com/aws/aws-sdk-cpp/pull/2710

cpp/include/kvikio/remote_handle.hpp
madsbk: We could support _move_ as long as we make `_shutdown_s3_api` non-const and sets it to `false` to avoid double `Aws::ShutdownAPI` calls

madsbk: but since `RemoteHandle` takes a shared pointer, I expect users to create the `S3Context` in a shared pointer as well

madsbk:
@wence- thanks for the review, I think I have addressed all of your suggestions

.github/workflows/pr.yaml
```diff
@@ -61,16 +61,17 @@ jobs:
...
+  # devcontainer:
+  #   secrets: inherit
+  #   uses: rapidsai/shared-workflows/.github/workflows/build-in-devcontainer.yaml@branch-24.10
+  #   with:
+  #     arch: '["amd64"]'
+  #     cuda: '["12.5"]'
+  #     build_command: |
+  #       sccache -z;
+  #       build-all --verbose;
+  #       sccache -s;
```
madsbk, line 74: uncomment before PR merge

python/kvikio/kvikio/remote_file.py
```diff
@@ -0,0 +1,148 @@
...
+
+def _get_remote_module():
+    """Get the remote module or raise an error"""
+    if not is_remote_file_available():
+        raise RuntimeError(
+            "RemoteFile not available, please build KvikIO with AWS S3 support"
+        )
+    import kvikio._lib.remote_handle
+
+    return kvikio._lib.remote_handle
```
wence-, line 29: These can't change their mind after calling, so perhaps trivial change:

```
from functools import cache

@cache
def is_remote_file_available() -> bool
   ...

@cache
def _get_remote_module()
```

python/kvikio/tests/test_aws_s3.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Fixture to set up moto server in separate process
+    """
+    with ensure_safe_environment_variables():
+        # Use fake aws credentials
+        os.environ["AWS_ACCESS_KEY_ID"] = "foobar_key"
+        os.environ["AWS_SECRET_ACCESS_KEY"] = "foobar_secret"
+        os.environ["AWS_SECURITY_TOKEN"] = "foobar_security_token"
+        os.environ["AWS_SESSION_TOKEN"] = "foobar_session_token"
+        os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
```
wence-, line 77: *nit* if you wanted you could use the `monkeypatch` fixture that pytest provides to handle this.

python/kvikio/tests/test_examples.py
```diff
@@ -26,3 +28,18 @@ def test_zarr_cupy_nvcomp(tmp_path, monkeypatch):
...
     import_module("zarr_cupy_nvcomp").main(tmp_path / "test-file")
+
+
+def test_aws_s3(monkeypatch):
+    """Test examples/aws_s3.py"""
+
+    if not kvikio.is_remote_file_available():
+        pytest.skip(
+            "cannot test remote IO, please build KvikIO with with AWS S3 support"
+        )
```
wence-, line 39:
```suggestion
@pytest.mark.skipif(
    not kvikio.is_remote_file_available(), 
    reason="KvikIO not built with AWS S3 support",
)
def test_aws_s3(monkeypatch):
    """Test examples/aws_s3.py"""

```

wence- (approved):
One issue in the dtor for `S3Context`, but otherwise only tiny suggestions.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,366 @@
...
+    Aws::Auth::DefaultAWSCredentialsProviderChain provider;
+    if (provider.GetAWSCredentials().IsEmpty()) {
+      throw std::runtime_error("failed authentication to S3 server");
+    }
+    _client = std::make_shared<Aws::S3::S3Client>(config);
+  }
+
+  ~S3Context() noexcept
+  {
+    if (_shutdown_s3_api) {
```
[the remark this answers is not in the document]
madsbk: Good catch, I will add some comments to make this clear

dependencies.yaml
```diff
@@ -315,6 +316,15 @@ dependencies:
...
           - pytest
           - pytest-cov
+          - boto3>=1.21.21
+      - output_types: [requirements, pyproject]
+        packages:
+          - moto[server]>=4.0.8
+      - output_types: conda
+        packages:
+          # TODO: Uncomment once https://github.com/rapidsai/cudf/pull/16499 is merged
+          #- cudf==24.10.*,>=0.0.0a0
```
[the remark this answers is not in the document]
madsbk: yes

python/kvikio/kvikio/benchmarks/aws_s3_io.py
```diff
@@ -0,0 +1,242 @@
...
+
+def start_s3_server(lifetime: int):
+    from moto.server import ThreadedMotoServer
+
+    # Silence the activity info from ThreadedMotoServer
+    sys.stderr = open("/dev/null", "w")
+    url = urlparse(os.environ["AWS_ENDPOINT_URL"])
+    server = ThreadedMotoServer(ip_address=url.hostname, port=url.port)
+    server.start()
+    time.sleep(lifetime)
```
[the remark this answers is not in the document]
madsbk: yes, added comments

python/kvikio/kvikio/benchmarks/aws_s3_io.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+@contextlib.contextmanager
+def local_s3_server(lifetime: int):
+    # Use fake aws credentials
+    os.environ["AWS_ACCESS_KEY_ID"] = "foobar_key"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "foobar_secret"
+    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
```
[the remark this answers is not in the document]
madsbk: I don't think it is worth depending on pytest in the benchmarks

python/kvikio/kvikio/benchmarks/aws_s3_io.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # Run each benchmark using the requested APIs
+    for api in args.api:
+        res = []
+        for elapsed in API[api](args):
+            res.append(elapsed)
+
+        def pprint_api_res(name, samples):
+            samples = [args.nbytes / s for s in samples]  # Convert to throughput
+            mean = statistics.mean(samples) if len(samples) > 1 else samples[0]
```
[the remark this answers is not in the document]
madsbk: Fixed

python/kvikio/kvikio/remote_file.py
madsbk: done

python/kvikio/tests/test_aws_s3.py
madsbk: Good idea, I just copied the test code from cudf

.github/workflows/pr.yaml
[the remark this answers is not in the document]
madsbk: Yes, it should be fixed before we merge

python/kvikio/kvikio/benchmarks/aws_s3_io.py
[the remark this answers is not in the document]
wence-: ah, didn't spot this was in benchmarks til after my review was done.

cpp/cmake/thirdparty/patches/aws-sdk-cpp/0001-Don-t-set-CMP0077-to-OLD.patch
```diff
@@ -0,0 +1,26 @@
+From 7b24166a73e422e65b725ffcb0acd20ab493fac0 Mon Sep 17 00:00:00 2001
```
mhaseeb123, line 1: Is this patch file intentially committed?

cpp/include/kvikio/bounce_buffer.hpp
```diff
@@ -78,12 +83,24 @@ class AllocRetain {
...
+  /**
+   * @brief Ensure the size of the retained allocations match `defaults::bounce_buffer_size()`
+   *
+   * NB: `_mutex` must be taken prior to calling this function.
+   */
+  void ensure_alloc_size()
   {
-    const std::lock_guard lock(_mutex);
-    if (_size != defaults::bounce_buffer_size()) {
+    const auto bounce_buffer_size = defaults::bounce_buffer_size();
```
mhaseeb123, line 93:
```suggestion
    auto const bounce_buffer_size = defaults::bounce_buffer_size();
```

madsbk:
@KyleFromNVIDIA, what is the status of the building work? Is it ready for review/merge?

python/kvikio/kvikio/remote_file.py
```diff
@@ -0,0 +1,151 @@
...
+        ret._handle = _get_remote_module().RemoteFile.from_url(context._handle, url)
+        return ret
+
+    def __enter__(self) -> RemoteFile:
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        pass
+
+    def nbytes(self) -> int:
```
bdice, line 105: Should this be a property?

python/kvikio/kvikio/remote_file.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Returns
+        -------
+        A newly opened remote file
+        """
+        ret = object.__new__(cls)
+        ret._handle = _get_remote_module().RemoteFile.from_url(context._handle, url)
+        return ret
+
+    def __enter__(self) -> RemoteFile:
```
bdice, line 99: What is the purpose of defining `__enter__` and `__exit__` to make this a context manager if there is no state/context being managed? They look like no-ops. I would expect that exiting the context manager should clean up and close connections.

python/kvikio/kvikio/remote_file.py
```diff
@@ -0,0 +1,151 @@
...
+        The number of bytes.
+        """
+        return self._handle.nbytes()
+
+    def read(self, buf, size: Optional[int] = None, file_offset: int = 0) -> int:
+        """Read from remote source into buffer (host or device memory) in parallel.
+
+        Parameters
+        ----------
+        buf: buffer-like or array-like
```
bdice, line 121: Missing a space before the type. See https://numpydoc.readthedocs.io/en/latest/format.html#parameters.
```suggestion
        buf : buffer-like or array-like
```

python/kvikio/kvikio/remote_file.py
```diff
@@ -0,0 +1,151 @@
...
+        The size of bytes that were successfully read.
+        """
+        return self.pread(buf, size, file_offset).get()
+
+    def pread(self, buf, size: Optional[int] = None, file_offset: int = 0) -> IOFuture:
+        """Read from remote source into buffer (host or device memory) in parallel.
+
+        Parameters
+        ----------
+        buf: buffer-like or array-like
```
bdice, line 139:
```suggestion
        buf : buffer-like or array-like
```

python/kvikio/tests/test_benchmarks.py
```diff
@@ -78,3 +80,60 @@ def test_zarr_io(run_cmd, tmp_path, api):
...
+    """Call pytest.skip() if cudf or its "libcudf_s3_io" option isn't available
+
+    See <https://github.com/rapidsai/cudf/pull/16499>
+    """
+    cudf = pytest.importorskip("cudf")
+    try:
+        cudf.get_option("libcudf_s3_io")
+    except KeyError:
+        pytest.skip(
+            """cudf doesn't has the "libcudf_s3_io" option, """
```
bdice, line 95: Suggesting a simpler quoting style here.
```suggestion
            "cudf doesn't have the 'libcudf_s3_io' option, "
```

python/kvikio/tests/test_benchmarks.py
```diff
@@ -78,3 +80,60 @@ def test_zarr_io(run_cmd, tmp_path, api):
...
+    ],
+)
+def test_aws_s3_io(run_cmd, api):
+    """Test benchmarks/aws_s3_io.py"""
+
+    if not kvikio.is_remote_file_available():
+        pytest.skip(
+            "cannot test remote IO, please build KvikIO with with AWS S3 support"
+        )
+    # Fail early if benchmark dependencies isn't available
```
bdice, line 116:
```suggestion
    # Fail early if benchmark dependencies aren't available
```

python/kvikio/tests/test_benchmarks.py
```diff
@@ -78,3 +80,60 @@ def test_zarr_io(run_cmd, tmp_path, api):
...
+    import boto3  # noqa: F401
+    import moto  # noqa: F401
+
+    if "cudf" in api:
+        skipif_libcudf_s3_io_option_is_not_available()
+
+    retcode = run_cmd(
+        cmd=[
+            sys.executable or "python",
```
bdice, line 125: I think `sys.executable` should always be defined? Are there cases where that isn't safe?

python/kvikio/tests/test_benchmarks.py
```diff
@@ -78,3 +80,60 @@ def test_zarr_io(run_cmd, tmp_path, api):
...
+            "-n",
+            "1000",
+            "-t",
+            "4",
+            "--api",
+            api,
+        ],
+        cwd=benchmarks_path,
+    )
+    assert retcode == 0
```
bdice, line 139: Should this assert anything about the stdout/stderr outputs?

bdice (changes requested):
I have some concerns about the handling of the `aws-sdk-cpp` dependency and how it will constrain our compatibility between RAPIDS and the conda-forge ecosystem. This dependency is used by many packages and it is tightly pinned, meaning that RAPIDS will need to track and continually migrate its dependency to align exactly with the conda-forge ecosystem. See threads in this review.

conda/recipes/kvikio/meta.yaml
```diff
@@ -64,11 +64,13 @@ requirements:
     - rapids-build-backend >=0.3.0,<0.4.0.dev0
     - scikit-build-core >=0.10.0
     - libkvikio ={{ version }}
+    - aws-sdk-cpp>=1.11.267
```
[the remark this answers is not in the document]
madsbk: Sorry my ignore here, but why do we need an exact match? Naively, I would have thought that a lower-limit would be more flexible? Do all packages in conda-forge needs to maintain this pinning if they use `aws-sdk-cpp`?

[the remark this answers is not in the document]
bdice: First I'll explain the migration logic. Then I'll answer the questions you posed about other options.

### Migrations and why they're needed

The `aws-sdk-cpp` release cadence is very frequent. Some of those releases ([there's a huge number of them](https://github.com/aws/aws-sdk-cpp/tags)) are picked up by conda-forge, but taking the new version requires "rebuilding the world" because `aws-sdk-cpp` is a core C++ dependency of many packages (see `mamba repoquery whoneeds -c conda-forge aws-sdk-cpp`). Updates happen via a conda-forge migration ([tracked here](https://conda-forge.org/status/)) which rebuild everything that depends on `aws-sdk-cpp`.

See the number of recent AWS-related rebuilds of libarrow, for instance:
https://github.com/conda-forge/arrow-cpp-feedstock/commits/main/recipe/meta.yaml

I'll ignore the `aws-crt-cpp` and other rebuilds, and only focus on `aws-sdk-cpp` because that's the dependency we are introducing here.

In the past 12 months, there have been 6 rebuilds of libarrow for new versions of `aws-sdk-cpp` ([1.11.379](https://github.com/conda-forge/arrow-cpp-feedstock/commit/8d5cd0e6e4b7150d99ea0b3c28f968500fdb9ba9), [1.11.329](https://github.com/conda-forge/arrow-cpp-feedstock/commit/1e381317006297d34d53bfda80cb427990a77532), [1.11.267](https://github.com/conda-forge/arrow-cpp-feedstock/commit/2da6505474c2724f24867a2b1434c5c9ad89cd49), [1.11.242](https://github.com/conda-forge/arrow-cpp-feedstock/commit/2fab9347fab92d72d8e8845f864545e83b92ec80), [1.11.210](https://github.com/conda-forge/arrow-cpp-feedstock/commit/e950ac4f208cddef50a79181389cf7f5a24e1fc8), [1.11.182](https://github.com/conda-forge/arrow-cpp-feedstock/commit/1608eaba13cfadc79580b6cc3f1ca7adb5dfffa7)).

Every time that conda-forge migrates to a new version, we must use the _same_ version so that RAPIDS (KvikIO specifically) can be installed alongside the latest conda-forge packages.

Try this:
```
mamba create -n test --dry-run libarrow
```

Note that `aws-sdk-cpp 1.11.379` is included in the environment because there has been a recent rebuild of `libarrow`.

### Can we use a lower limit (or no version pinning)?

If we were to use a lower limit (or no version pinning), kvikio would use the latest `aws-sdk-cpp` when it is built. That would impose a runtime pinning on the latest `aws-sdk-cpp`. However, if a conda-forge migration has not happened yet (or is incomplete), those `kvikio` packages would be incompatible with the latest conda-forge packages of things like libarrow. We need to match the migrators so we lag behind "latest `aws-sdk-cpp`" and instead align with "current version used by conda-forge". Typically conda-forge maintainers will start a migration as soon as a new `aws-sdk-cpp` is released, but that's not guaranteed to happen immediately. Sometimes migrations are [delayed](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5913#issuecomment-2119201067) or [skipped](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4883#issuecomment-1709403242), which would leave our packages unusable if we had picked up an `aws-sdk-cpp` version that conda-forge did not adopt with a migration.


### So what do we do?

RAPIDS is not built with conda-forge infrastructure, but aims to be compatible with the conda-forge ecosystem. We do not benefit from conda-forge's automation for version migrations but we must track them in order to be compatible. (The conda-forge bots that "rebuild the world" do not rebuild RAPIDS recipes, but maybe we could create such a tool.)

To remain compatible with conda-forge, we need to match the exact pinnings for core packages like `fmt`, `spdlog`, `aws-sdk-cpp`, and others. We have a lot of pain from matching conda-forge migrations for `fmt` and `spdlog` (see [Keith's warning](https://github.com/rapidsai/build-planning/issues/56#issuecomment-2288846888) and [this problem in cuspatial](https://github.com/rapidsai/cuspatial/pull/1453#issuecomment-2335527542) that comes from us being behind on spdlog/fmt migrations)

Our best course of action for now is to:
- pin `aws-sdk-cpp` to the exact version in this YAML file: https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/f291cfdaf8c328337cb8cd1f63c63caceeda8991/recipe/conda_build_config.yaml#L267-L268 (that's a permalink but we should track the latest as it changes)
  - we cannot pin more loosely -- `aws-sdk-cpp` imposes run-exports that pin very tightly, which typically indicates that there is no guarantee of stable ABI across versions
- manually update that version when conda-forge issues a migration
- consider adding tooling to help automate migrations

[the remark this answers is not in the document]
bdice: I opened https://github.com/rapidsai/build-planning/issues/100 to think more about conda-forge migration support in RAPIDS.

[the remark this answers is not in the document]
madsbk: Wow, thanks for the explanation @bdice! I now see why you were hesitant from the get-go 😊  

What if we move to use `libcurl` instead? They take API and ABI stability very serious: https://curl.se/libcurl/features.html#stableapi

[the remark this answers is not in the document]
madsbk: I was actually thinking of using libcurl from the start since we only need the very basic S3 operations. I don’t think it will be too hard to implement and it will make Azure Blob Storage support straightforward.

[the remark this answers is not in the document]
bdice: That might be a better option than `aws-sdk-cpp` if we can use `curl`. It seems like `curl` is already a dependency of some packages we depend on in RAPIDS, so that gives me more confidence in it. See `$ mamba repoquery whoneeds -c conda-forge curl`.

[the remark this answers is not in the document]
madsbk: Great, I will give it a try

GregoryKimball:
@madsbk Would you please share the status of this work? Are you planning to include AWS C++ SDK changes in 24.12?

madsbk:
> @madsbk Would you please share the status of this work? Are you planning to include AWS C++ SDK changes in 24.12?

The plan is to use libcurl instead of aws-s3-sdk to support S3. First step implemented by https://github.com/rapidsai/kvikio/pull/464.

vyasr:
Replaced by #479.

PR #432, opened 2024-08-07
Improve update-version.sh

bdice:
This PR removes the `UCX_PY` version HTTP call from `update-version.sh` because it is not used.


jameslamb (approved):
Hey nice! Glad we could get rid of one of these calls.

PR #437, opened 2024-08-15
Stream sync before calling cufile

madsbk:
Contrary to most of the non-async CUDA API, cuFile does not have the semantic of being ordered with respect to other work in the null stream. Since this is surprising to most users, we now synchronize on the null stream before calling cuFIle read and write by default. 

See [cuFile IO API Functional Specification](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-io-api-functional-specification) and [CUDA Concurrent Execution between Host and Device](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#concurrent-execution-between-host-and-device).


cpp/include/kvikio/file_handle.hpp
```diff
@@ -314,16 +314,26 @@ class FileHandle {
    * @param file_offset Offset in the file to read from.
    * @param devPtr_offset Offset relative to the `devPtr_base` pointer to read into.
    * This parameter should be used only with registered buffers.
+   * @param sync_null_stream Synchronize the CUDA null stream prior to calling cuFile. Contrary to
```
bdice, line 317: Do we mean "default stream" instead of "null stream" everywhere here? https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html

madsbk: Right, I have renamed to default stream. cuFile calls it the [null stream](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-io-api-functional-specification), but I agree default stream is less confusing.

bdice (approved):
This looks good!

Would an equivalent alternative be to use the async API and pass in the default stream? Is that possible with the current design? If so, I would explain in the docs how this new option behaves compared to that alternative (or simply direct users to an async API rather than add a new parameter if it is equivalent).

madsbk:
> This looks good!
> 
> Would an equivalent alternative be to use the async API and pass in the default stream? Is that possible with the current design? If so, I would explain in the docs how this new option behaves compared to that alternative (or simply direct users to an async API rather than add a new parameter if it is equivalent).

Thanks @bdice, the async API comes with its own quirks so it isn't really an equivalent alternative.

PR #438, opened 2024-08-15
`config_path()`: return a reference

madsbk:
Avoid a string copy

### DIFF
--- a/cpp/include/kvikio/cufile_config.hpp
+++ b/cpp/include/kvikio/cufile_config.hpp
@@ -22,12 +22,12 @@
 namespace kvikio {
 namespace detail {
 
-[[nodiscard]] inline const std::string lookup_config_path()
+[[nodiscard]] inline const char* lookup_config_path()
 {
   const char* env = std::getenv("CUFILE_ENV_PATH_JSON");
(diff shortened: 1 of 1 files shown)


bdice (approved):

cpp/include/kvikio/cufile_config.hpp
```diff
@@ -22,12 +22,12 @@
 namespace kvikio {
 namespace detail {
 
-[[nodiscard]] inline const std::string lookup_config_path()
+[[nodiscard]] inline const char* lookup_config_path()
 {
   const char* env = std::getenv("CUFILE_ENV_PATH_JSON");
   if (env != nullptr && std::filesystem::exists(env)) { return env; }
   if (std::filesystem::exists("/etc/cufile.json")) { return "/etc/cufile.json"; }
```
vuule, line 29: side question: could this be done once (e.g. init a local static var)?

madsbk: It is done once, `config_path()` calls it to assign a static variable

vuule: Got it. I assumed repeated calls because of the optimization.

PR #439, opened 2024-08-16
Fix several issues with wheel builds

KyleFromNVIDIA:
* Ensure correct directory layout in both wheels
* Add CMake entry point to libkvikio wheel
* Require kvikio wheel to use libkvikio wheel
* Exclude unneeded source files from wheel


cpp/cmake/rapids_config.cmake
```diff
@@ -26,10 +26,12 @@ else()
   )
 endif()
 
+set(rapids-cmake-repo KyleFromNVIDIA/rapids-cmake)
+set(rapids-cmake-branch nvcomp-nested-dir)
 if(NOT EXISTS "${CMAKE_CURRENT_BINARY_DIR}/KVIKIO_RAPIDS-${RAPIDS_VERSION_MAJOR_MINOR}.cmake")
   file(
     DOWNLOAD
-    "https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-${RAPIDS_VERSION_MAJOR_MINOR}/RAPIDS.cmake"
+    "https://raw.githubusercontent.com/KyleFromNVIDIA/rapids-cmake/nvcomp-nested-dir/RAPIDS.cmake"
```
KyleFromNVIDIA, line 34: This is to get the changes in rapidsai/rapids-cmake#675 and will be removed once that is merged.

python/libkvikio/pyproject.toml
```diff
@@ -55,3 +56,6 @@ requires = [
     "cmake>=3.26.4,!=3.30.0",
     "ninja",
 ] # This list was generated by `rapids-dependency-file-generator`. To make changes, edit ../../dependencies.yaml and run `rapids-dependency-file-generator`.
+
+[project.entry-points."cmake.prefix"]
+libkvikio = "libkvikio"
```
jameslamb, line 61: I didn't realize `scikit-build` allowed you to do this.

https://github.com/scikit-build/scikit-build-core/blob/8cbd92f430c20ad5e0d9725198a1397ab599437d/docs/cmakelists.md?plain=1#L83-L87

https://github.com/scikit-build/scikit-build-core/blob/8cbd92f430c20ad5e0d9725198a1397ab599437d/src/scikit_build_core/builder/builder.py#L129-L131

Interesting! If you know the history for why this re-uses the `project.entry-points` table (which I'd always thought was intended to provide CLI commands that call into Python code), instead of having `scikit-build-core`-specific configuration that just accepts a list of paths to append to `CMAKE_PREFIX_PATH`, I'd love to hear it.

python/kvikio/pyproject.toml
```diff
@@ -127,6 +127,7 @@ ninja.make-fallback = true
 sdist.exclude = ["*tests*"]
 sdist.reproducible = true
 wheel.packages = ["kvikio"]
+wheel.exclude = ["*.pyx", "*.pxd"]
```
jameslamb, line 130: oh nice! Makes sense to me that `kivkio` wouldn't need to include these sources, and that`libkvikio` would be the only one that needs the library and headers.

This is a good trick to keep in mind as we continue with other libraries for https://github.com/rapidsai/build-planning/issues/33.

jameslamb (approved):
these changes all make sense to me, thanks

python/libkvikio/pyproject.toml
KyleFromNVIDIA: CLI commands are just one use case for entry points. In reality they can be used for many different things.

The reason a `scikit-build-core`-specific configuration isn't used is because `CMAKE_PREFIX_PATH` isn't used when buiilding *this* package, but rather when building packages that consume it. Some sort of metadata has to be installed along with this package. Entry points are the way to do that.

jameslamb: ooooo I see. Thanks for explaining that!

KyleFromNVIDIA:
cc @msarahan

msarahan (approved):
good fixes! Thanks.

robertmaynard (approved):

python/kvikio/pyproject.toml
vyasr: I agree with not including pyx files, but we should be careful about not including pxd files. I _think_ this is currently fine for kvikio because none of kvikio's pxd files are intended for public consumption, but that is definitely not going to be generally true in RAPIDS. Most of the time we do want to include the pxd files.

ci/build_wheel_python.sh
```diff
@@ -24,6 +24,7 @@ cd "${package_dir}"
 echo "libkvikio-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${CPP_WHEELHOUSE}/libkvikio_*.whl)" > ./constraints.txt
 
 PIP_CONSTRAINT="${PWD}/constraints.txt" \
+SKBUILD_CMAKE_ARGS="-DFIND_KVIKIO_CPP=ON;-DCMAKE_INSTALL_INCLUDEDIR=kvikio/include;-DCMAKE_INSTALL_LIBDIR=kvikio/lib64" \
```
vyasr, line 27: Why do we need to specify the include and lib install dirs for the Python package? What gets installed to these directories, and why isn't it part of the libkvikio package instead?

KyleFromNVIDIA: Looks like it's currently just installing `nvcomp` and `gdeflate`. I'll see if there's some way to not have these present in the wheel.

KyleFromNVIDIA: I've prevented `nvcomp` from being installed and removed these arguments.

python/kvikio/pyproject.toml
KyleFromNVIDIA: I've put back the `*.pxd` files.

KyleFromNVIDIA: I've also opened https://github.com/rapidsai/ucxx/pull/270 to correct a mistake I made.

ci/build_wheel_python.sh
KyleFromNVIDIA: > I think installing libnvcomp*.so while avoiding installing the headers is the right choice because we don't want to make our wheels provide a complete installation of nvcomp that consumers could try to use.

That's what's currently happening here. `libnvcomp*.so` is located in the `kvikio/_lib` directory, and no headers are installed.

vyasr: Oh good right we do this correctly [here](https://github.com/rapidsai/kvikio/blob/branch-24.10/python/kvikio/CMakeLists.txt#L58). I forgot that I walked Michael through this in #369. This looks good, then.

vyasr (approved):

PR #441, opened 2024-08-19
Remove NumPy <2 pin

seberg:
This PR removes the NumPy<2 pin. This is desirable for RAPIDS projects that do not have a strong numpy/cupy dependency.
Not doing any math, they don't run into subtle issues, but the pin was still necessary briefly before NumPy 2 was initially released.

(Other RAPIDS projects rely heavily on CuPy/NumPy and need the soon to be released 13.3.0 to function properly, e.g. due to promotion related changes in NumPy.)


seberg:
Seems this is already using numpy 2 and working fine (i.e. kvikio doesn't run into the CuPy 13.2 issues, and doesn't depend on rmm or so limiting it).  So it seems this is probably ready.

wence- (approved):

jameslamb (approved):
Looks good to me, based on my understanding of https://github.com/rapidsai/ucx-py/pull/1064#discussion_r1725786361.

Please remove the language in the description about this being unsafe combined with `cupy < 13.3.0` before merging.

jakirkham:
Sebastian rewrote the description based on the feedback in the UCX-Py PR and updated it here as well

jakirkham (approved):

jakirkham:
Thanks Sebastian, James, and Lawrence! 🙏

PR #445, opened 2024-08-23
Update `build.sh`

madsbk:


build.sh
```diff
@@ -115,6 +115,8 @@ if hasArg -v; then
 fi
 if hasArg -g; then
     BUILD_TYPE=Debug
+    export SKBUILD_INSTALL_STRIP=false
+    export SKBUILD_CMAKE_BUILD_TYPE=""Debug""
```
bdice, line 119: Why double double quotes?
```suggestion
    export SKBUILD_CMAKE_BUILD_TYPE="Debug"
```

docs/source/install.rst
```diff
@@ -48,11 +48,11 @@ In order to setup a development environment run:
   # CUDA 12.5
   mamba env create --name kvikio-dev --file conda/environments/all_cuda-125_arch-x86_64.yaml
 
-To build and install the extension run:
+The Python library depend on the C++ library, thus we build and install both:
```
bdice, line 51:
```suggestion
The Python library depends on the C++ library, thus we build and install both:
```

bdice (approved):
Two comments, please resolve as you see fit. Approving.

build.sh
wence-: I just copied as-is from the scikit-build documentation

madsbk: It looks like a typo, at least other code, including in their own testing, just use `SKBUILD_CMAKE_BUILD_TYPE=Debug`.
Fixed

PR #447, opened 2024-08-26
Make the bounce buffer size configurable 

madsbk:


python/kvikio/tests/test_basic_io.py
```diff
@@ -256,3 +256,20 @@ def test_multiple_gpus(tmp_path, xp, gds_threshold):
                 with cupy.cuda.Device(0):
                     assert f.read(a1) == a1.nbytes
             assert bytes(a0) == bytes(a1)
+
+
+@pytest.mark.parametrize("size", [1, 10, 100, 1000])
+@pytest.mark.parametrize("tasksize", [1, 10, 100, 1000])
+@pytest.mark.parametrize("buffer_size", [1, 10, 100, 1000])
```
bdice, line 263: We need tests for the error case of buffer size 0.

python/kvikio/kvikio/defaults.py
```diff
@@ -200,3 +200,46 @@ def set_gds_threshold(nbytes: int):
...
+    Parameters
+    ----------
+    nbytes : int
+        The bounce buffer size in bytes.
+    """
+    kvikio._lib.defaults.bounce_buffer_size_reset(nbytes)
+
+
+@contextlib.contextmanager
+def set_bounce_buffer_size(nbytes: int):
```
bdice, line 232: What happens if a negative value is used? Does it wrap to a huge value? Do huge values cause errors? Let’s check the behavior and error appropriately.

python/kvikio/tests/test_basic_io.py
madsbk: Good point, added tests

bdice (approved):

PR #449, opened 2024-08-29
Update nvcomp to 4.0.1

vuule:
Update the nvCOMP version used for compression/decompression to 4.0.1.

See also:

https://github.com/rapidsai/rapids-cmake/pull/633
https://github.com/rapidsai/cudf/pull/16076

Depends on https://github.com/rapidsai/kvikio/pull/450.

PR #450, opened 2024-08-29
Use nvcomp defaults for algo options.

bdice:
This PR updates the nvcomp bindings to use nvcomp's defaults rather than hardcode the default options. The defaults changed in nvcomp 4.0.1, so this is needed for https://github.com/rapidsai/kvikio/pull/449.

### DIFF
--- a/python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
+++ b/python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
@@ -390,7 +390,7 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
         elif isinstance(data_type, int):
             self.options = nvcompBatchedLZ4Opts_t(data_type)
         else:
-            raise ValueError("Invalid value for data_type: {data_type}")
+            raise ValueError(f"Invalid value for data_type: {data_type}")
 
         self.has_header = has_header
 
--- a/python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
+++ b/python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
@@ -357,6 +357,7 @@ from kvikio._lib.nvcomp_ll_cxx_api cimport (
     nvcompBatchedLZ4CompressGetTempSize,
     nvcompBatchedLZ4DecompressAsync,
     nvcompBatchedLZ4DecompressGetTempSize,
+    nvcompBatchedLZ4DefaultOpts,
     nvcompBatchedLZ4GetDecompressSizeAsync,
     nvcompBatchedLZ4Opts_t,
 )
@@ -371,20 +372,26 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
 
     HEADER_SIZE_BYTES: size_t = sizeof(uint32_t)
 
-    def __init__(self, data_type: int = 0, has_header: bool = True):
+    def __init__(self, data_type: int = None, has_header: bool = True):
         """Initialize the codec.
 
         Parameters
         ----------
-        data_type: int
-            Source data type.
+        data_type: int or None
+            Source data type. If None, uses nvcomp default options.
         has_header: bool
             Whether the compressed data has a header.
             This enables data compatibility between numcodecs LZ4 codec,
             which has the header and nvCOMP LZ4 codec which does not
             require the header.
         """
-        self.options = nvcompBatchedLZ4Opts_t(data_type)
+        if data_type is None:
+            self.options = nvcompBatchedLZ4DefaultOpts
+        elif isinstance(data_type, int):
+            self.options = nvcompBatchedLZ4Opts_t(data_type)
+        else:
+            raise ValueError("Invalid value for data_type: {data_type}")
+
         self.has_header = has_header
 
         # Note on LZ4 header structure: numcodecs LZ4 codec prepends
@@ -621,6 +628,7 @@ from kvikio._lib.nvcomp_ll_cxx_api cimport (
     nvcompBatchedGdeflateCompressGetTempSize,
     nvcompBatchedGdeflateDecompressAsync,
     nvcompBatchedGdeflateDecompressGetTempSize,
+    nvcompBatchedGdeflateDefaultOpts,
     nvcompBatchedGdeflateGetDecompressSizeAsync,
     nvcompBatchedGdeflateOpts_t,
 )
@@ -633,8 +641,13 @@ class nvCompBatchAlgorithmGdeflate(nvCompBatchAlgorithm):
 
     options: nvcompBatchedGdeflateOpts_t
 
-    def __init__(self, algo: int = 0):
-        self.options = nvcompBatchedGdeflateOpts_t(algo)
+    def __init__(self, algo: int = None):
+        if algo is None:
+            self.options = nvcompBatchedGdeflateDefaultOpts
+        elif isinstance(algo, int):
+            self.options = nvcompBatchedGdeflateOpts_t(algo)
+        else:
+            raise ValueError("Invalid value for algo: {algo}")
 
     def _get_comp_temp_size(
         self,
@@ -756,6 +769,7 @@ from kvikio._lib.nvcomp_ll_cxx_api cimport (
     nvcompBatchedZstdCompressGetTempSize,
     nvcompBatchedZstdDecompressAsync,
     nvcompBatchedZstdDecompressGetTempSize,
+    nvcompBatchedZstdDefaultOpts,
     nvcompBatchedZstdGetDecompressSizeAsync,
     nvcompBatchedZstdOpts_t,
 )
@@ -769,7 +783,7 @@ class nvCompBatchAlgorithmZstd(nvCompBatchAlgorithm):
     options: nvcompBatchedZstdOpts_t
 
     def __init__(self):
-        self.options = nvcompBatchedZstdOpts_t(0)
+        self.options = nvcompBatchedZstdDefaultOpts
 
     def _get_comp_temp_size(
         self,
@@ -891,6 +905,7 @@ from kvikio._lib.nvcomp_ll_cxx_api cimport (
     nvcompBatchedSnappyCompressGetTempSize,
     nvcompBatchedSnappyDecompressAsync,
     nvcompBatchedSnappyDecompressGetTempSize,
+    nvcompBatchedSnappyDefaultOpts,
     nvcompBatchedSnappyGetDecompressSizeAsync,
     nvcompBatchedSnappyOpts_t,
 )
@@ -904,7 +919,7 @@ class nvCompBatchAlgorithmSnappy(nvCompBatchAlgorithm):
     options: nvcompBatchedSnappyOpts_t
 
     def __init__(self):
-        self.options = nvcompBatchedSnappyOpts_t(0)
+        self.options = nvcompBatchedSnappyDefaultOpts
 
     def _get_comp_temp_size(
         self,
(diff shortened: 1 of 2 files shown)


jakirkham:
@madsbk @Alexey-Kamenev @thomcom could you please look over these changes when you have a moment? 🙂

vuule (approved):
Semantically the change looks good to me, not qualified to evaluate the (Python) code quality.

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -371,20 +372,26 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
...
             which has the header and nvCOMP LZ4 codec which does not
             require the header.
         """
-        self.options = nvcompBatchedLZ4Opts_t(data_type)
+        if data_type is None:
+            self.options = nvcompBatchedLZ4DefaultOpts
+        elif isinstance(data_type, int):
+            self.options = nvcompBatchedLZ4Opts_t(data_type)
+        else:
+            raise ValueError("Invalid value for data_type: {data_type}")
```
Alexey-Kamenev, line 393: I think this is missing `f` to make it a proper f-string. Same in other places.

Alexey-Kamenev (approved):
LGTM with minor comment.

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
jakirkham: Agreed let's fix that

```suggestion
            raise ValueError(f"Invalid value for data_type: {data_type}")
```

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -633,8 +641,13 @@ class nvCompBatchAlgorithmGdeflate(nvCompBatchAlgorithm):
...
 
-    def __init__(self, algo: int = 0):
-        self.options = nvcompBatchedGdeflateOpts_t(algo)
+    def __init__(self, algo: int = None):
+        if algo is None:
+            self.options = nvcompBatchedGdeflateDefaultOpts
+        elif isinstance(algo, int):
+            self.options = nvcompBatchedGdeflateOpts_t(algo)
+        else:
+            raise ValueError("Invalid value for algo: {algo}")
```
jakirkham, line 650:
```suggestion
            raise ValueError(f"Invalid value for algo: {algo}")
```

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -1038,14 +1054,19 @@ class nvCompBatchAlgorithmDeflate(nvCompBatchAlgorithm):
...
 
-    def __init__(self, algo: int = 0):
-        self.options = nvcompBatchedDeflateOpts_t(algo)
+    def __init__(self, algo: int = None):
+        if algo is None:
+            self.options = nvcompBatchedDeflateDefaultOpts
+        elif isinstance(algo, int):
+            self.options = nvcompBatchedDeflateOpts_t(algo)
+        else:
+            raise ValueError("Invalid value for algo: {algo}")
```
jakirkham, line 1063:
```suggestion
            raise ValueError(f"Invalid value for algo: {algo}")
```

jakirkham (commented):
After Alexey observed the `f-string` typo. Saw a few more. Flagged them below

jakirkham:
As it appears we do tests constructing compressors with default arguments, this should already be covered in the existing test suite

https://github.com/rapidsai/kvikio/blob/3ce8bcc4da057f3c9fc7855d6297c73a448bc836/python/kvikio/tests/test_nvcomp.py#L195-L217

Though please let me know if I've missed anything here

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -371,20 +372,26 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
 
     HEADER_SIZE_BYTES: size_t = sizeof(uint32_t)
 
-    def __init__(self, data_type: int = 0, has_header: bool = True):
+    def __init__(self, data_type: int = None, has_header: bool = True):
```
madsbk, line 375:
```suggestion
    def __init__(self, data_type: int | None = None, has_header: bool = True):
```

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -633,8 +641,13 @@ class nvCompBatchAlgorithmGdeflate(nvCompBatchAlgorithm):
 
     options: nvcompBatchedGdeflateOpts_t
 
-    def __init__(self, algo: int = 0):
-        self.options = nvcompBatchedGdeflateOpts_t(algo)
+    def __init__(self, algo: int = None):
```
madsbk, line 644:
```suggestion
    def __init__(self, algo: int | None = None):
```

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -1038,14 +1054,19 @@ class nvCompBatchAlgorithmDeflate(nvCompBatchAlgorithm):
 
     options: nvcompBatchedDeflateOpts_t
 
-    def __init__(self, algo: int = 0):
-        self.options = nvcompBatchedDeflateOpts_t(algo)
+    def __init__(self, algo: int = None):
```
madsbk, line 1057:
```suggestion
    def __init__(self, algo: int | None = None):
```

madsbk (approved):
Looks good, I only have some type hint suggestions

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -371,20 +372,26 @@ class nvCompBatchAlgorithmLZ4(nvCompBatchAlgorithm):
...
             which has the header and nvCOMP LZ4 codec which does not
             require the header.
         """
-        self.options = nvcompBatchedLZ4Opts_t(data_type)
+        if data_type is None:
+            self.options = nvcompBatchedLZ4DefaultOpts
+        elif isinstance(data_type, int):
+            self.options = nvcompBatchedLZ4Opts_t(data_type)
+        else:
+            raise ValueError(f"Invalid value for data_type: {data_type}")
```
wence-, line 393: *nit (non-blocking)*: This is arguably a `TypeError`, not a `ValueError`. We've passed an incorrect type, not a correct type with an incorrect value.

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -633,8 +641,13 @@ class nvCompBatchAlgorithmGdeflate(nvCompBatchAlgorithm):
...
 
-    def __init__(self, algo: int = 0):
-        self.options = nvcompBatchedGdeflateOpts_t(algo)
+    def __init__(self, algo: int = None):
+        if algo is None:
+            self.options = nvcompBatchedGdeflateDefaultOpts
+        elif isinstance(algo, int):
+            self.options = nvcompBatchedGdeflateOpts_t(algo)
+        else:
+            raise ValueError(f"Invalid value for algo: {algo}")
```
wence-, line 650: *nit (non-blocking)*: similarly here `TypeError` not `ValueError`?

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
jakirkham: Think Cython currently doesn't handle `typing.Union`s ( https://github.com/cython/cython/pull/6206 ) and probably won't. Though it might handle the special case with `None` in the future ( https://github.com/cython/cython/issues/6254 ), it doesn't appear to be included yet

Reading [this Cython FAQ]( https://cython.readthedocs.io/en/latest/src/userguide/faq.html#why-does-a-function-with-cdef-d-parameters-accept-none ) it suggests to pick one of...

* `func(x: list = None)` (as Bradley has)
* `func(x: Optional[list])` (from locally testing found it needed to be `func(x: Optional[list] = None)`)

So which do you prefer?

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -1038,14 +1054,19 @@ class nvCompBatchAlgorithmDeflate(nvCompBatchAlgorithm):
...
 
-    def __init__(self, algo: int = 0):
-        self.options = nvcompBatchedDeflateOpts_t(algo)
+    def __init__(self, algo: int = None):
+        if algo is None:
+            self.options = nvcompBatchedDeflateDefaultOpts
+        elif isinstance(algo, int):
+            self.options = nvcompBatchedDeflateOpts_t(algo)
+        else:
+            raise ValueError(f"Invalid value for algo: {algo}")
```
wence-, line 1063: *nit*: ibid.

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
```diff
@@ -1038,14 +1054,19 @@ class nvCompBatchAlgorithmDeflate(nvCompBatchAlgorithm):
...
 
     def _get_comp_temp_size(
         self,
         size_t batch_size,
         size_t max_uncompressed_chunk_bytes,
-    ) -> (nvcompStatus_t, size_t):
+    ) -> tuple[nvcompStatus_t, size_t]:
```
wence-, line 1069: hmm, how did this ever work? I guess cython is less strict with parsing.

wence- (approved):
Non-blocking nits

jakirkham (commented):
Cython already checks whether the value is an `int` or `None` based on the function signature. So the exceptions can be dropped. If we wanted to be extra careful, we could add an `assert`. Though likely that is unneeded

Example of this behavior demonstrated below:

<details>

```python
In [1]: %load_ext cython

In [2]: %%cython
   ...: 
   ...: from typing import Optional
   ...: 
   ...: class Class:
   ...:     def __init__(self, v: Optional[int] = None):
   ...:         print(f"v = {v}")
   ...:         self.v = v
   ...: 

In [3]: Class()
v = None
Out[3]: <_cython_magic_96fdfa8e6c04d6dbb7c0c7398f039f6decfcbe5b.Class at 0x10524ad10>

In [4]: Class(1)
v = 1
Out[4]: <_cython_magic_96fdfa8e6c04d6dbb7c0c7398f039f6decfcbe5b.Class at 0x104f4c610>

In [5]: Class(3.14)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 Class(3.14)

TypeError: Argument 'v' has incorrect type (expected int, got float)
```

</details>

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
jakirkham: Both are valid, the type format we had before is called a `ctuple` and has been around in Cython longer

Support for `typing.Tuple` is relatively new

Here's [an example in the docs]( https://cython.readthedocs.io/en/3.0.x/src/userguide/language_basics.html#types )

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
jakirkham: Agreed. Though Cython generates this check for us. So we can drop it altogether

There's an example in the top review comment

https://github.com/rapidsai/kvikio/pull/450#pullrequestreview-2271498081

vuule:
Thanks everyone for the reviews! I'll wait another work day to make sure folks are happy with the way the comments like https://github.com/rapidsai/kvikio/pull/450#discussion_r1738036063 have been resolved.

bdice (commented):
Thanks for the reviews and fixes. I am happy with this.

python/kvikio/kvikio/_lib/libnvcomp_ll.pyx
madsbk: No strong opinion between the two, let's god with Bradley's approach :)

PR #454, opened 2024-09-02
Non-static bounce buffer option

### LINKED ISSUE #451: Allow task tuning within the same process
When benchmarking with the `KVIKIO_TASK_SIZE` and `KVIKIO_NTHREADS`, I observed that changes to these variable don't seem to take effect if kvikIO has already been initialized in a process. For instance, if I loop over `KVIKIO_NTHREADS` in python, all of the reads seem to use the default values. However, if I set  `KVIKIO_TASK_SIZE` and `KVIKIO_NTHREADS` before launching the python process, then they appear to work as intended.

Is it possible for `KVIKIO_TASK_SIZE` and `KVIKIO_NTHREADS` to be updated dynamically within a process?

### LINKED ISSUE #452: Verify behavior of `KVIKIO_BOUNCE_BUFFER_SIZE`
When attempting to analyze large task sizes, it appears that the size of a copy is always capped at 16 MiB. 

Even with `KVIKIO_BOUNCE_BUFFER_SIZE` of 40 MB and `KVIKIO_TASK_SIZE` of 40 MB, I observe copies capped at ~16 MiB in the profile. Is there a way to enable larger task sizes?

Image

madsbk:
closes #451 
closes #452

Also some refactor/clean up

### DIFF
--- a/cpp/doxygen/main_page.md
+++ b/cpp/doxygen/main_page.md
@@ -99,12 +99,12 @@ KvikIO splits parallel IO operations into multiple tasks. Set the environment va
 This setting can also be controlled by `defaults::task_size()` and `defaults::task_size_reset()`.
 
 #### GDS Threshold (KVIKIO_GDS_THRESHOLD)
-In order to improve performance of small IO, `.pread()` and `.pwrite()` implement a shortcut that circumvent the threadpool and use the POSIX backend directly. Set the environment variable `KVIKIO_GDS_THRESHOLD` to the minimum size (in bytes) to use GDS. If not set, the default value is 1048576 (1 MiB).
+To improve performance of small IO requests, `.pread()` and `.pwrite()` implement a shortcut that circumvents the threadpool and uses the POSIX backend directly. Set the environment variable `KVIKIO_GDS_THRESHOLD` to the minimum size (in bytes) to use GDS. If not set, the default value is 1048576 (1 MiB).
 
 This setting can also be controlled by `defaults::gds_threshold()` and `defaults::gds_threshold_reset()`.
 
 #### Size of the Bounce Buffer (KVIKIO_GDS_THRESHOLD)
-KvikIO might have to use an intermediate host buffer when copying between file and device memory. Set the environment variable ``KVIKIO_BOUNCE_BUFFER_SIZE`` to size (in bytes) of this "bounce" buffer. If not set, the default value is 16777216 (16 MiB).
+KvikIO might have to use an intermediate host buffer when copying between files and device memory. Set the environment variable ``KVIKIO_BOUNCE_BUFFER_SIZE`` to the size (in bytes) of this "bounce" buffer. If not set, the default value is 16777216 (16 MiB).
 
 This setting can also be controlled by `defaults::bounce_buffer_size()` and `defaults::bounce_buffer_size_reset()`.
 
--- a/cpp/doxygen/main_page.md
+++ b/cpp/doxygen/main_page.md
@@ -85,14 +85,28 @@ Set the environment variable `KVIKIO_COMPAT_MODE` to enable/disable compatibilit
   - when running in Windows Subsystem for Linux (WSL).
   - when `/run/udev` isn't readable, which typically happens when running inside a docker image not launched with `--volume /run/udev:/run/udev:ro`.
 
+This setting can also be controlled by `defaults::compat_mode()` and `defaults::compat_mode_reset()`.
+
+
 #### Thread Pool (KVIKIO_NTHREADS)
 KvikIO can use multiple threads for IO automatically. Set the environment variable `KVIKIO_NTHREADS` to the number of threads in the thread pool. If not set, the default value is 1.
+
+This setting can also be controlled by `defaults::thread_pool_nthreads()` and `defaults::thread_pool_nthreads_reset()`.
 
 #### Task Size (KVIKIO_TASK_SIZE)
 KvikIO splits parallel IO operations into multiple tasks. Set the environment variable `KVIKIO_TASK_SIZE` to the maximum task size (in bytes). If not set, the default value is 4194304 (4 MiB).
 
+This setting can also be controlled by `defaults::task_size()` and `defaults::task_size_reset()`.
+
 #### GDS Threshold (KVIKIO_GDS_THRESHOLD)
 In order to improve performance of small IO, `.pread()` and `.pwrite()` implement a shortcut that circumvent the threadpool and use the POSIX backend directly. Set the environment variable `KVIKIO_GDS_THRESHOLD` to the minimum size (in bytes) to use GDS. If not set, the default value is 1048576 (1 MiB).
+
+This setting can also be controlled by `defaults::gds_threshold()` and `defaults::gds_threshold_reset()`.
+
+#### Size of the Bounce Buffer (KVIKIO_GDS_THRESHOLD)
+KvikIO might have to use an intermediate host buffer when copying between file and device memory. Set the environment variable ``KVIKIO_BOUNCE_BUFFER_SIZE`` to size (in bytes) of this "bounce" buffer. If not set, the default value is 16777216 (16 MiB).
+
+This setting can also be controlled by `defaults::bounce_buffer_size()` and `defaults::bounce_buffer_size_reset()`.
 
 
 ## Example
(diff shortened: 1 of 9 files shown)


cpp/doxygen/main_page.md
```diff
@@ -85,15 +85,29 @@ Set the environment variable `KVIKIO_COMPAT_MODE` to enable/disable compatibilit
...
 
+This setting can also be controlled by `defaults::thread_pool_nthreads()` and `defaults::thread_pool_nthreads_reset()`.
+
 #### Task Size (KVIKIO_TASK_SIZE)
 KvikIO splits parallel IO operations into multiple tasks. Set the environment variable `KVIKIO_TASK_SIZE` to the maximum task size (in bytes). If not set, the default value is 4194304 (4 MiB).
 
+This setting can also be controlled by `defaults::task_size()` and `defaults::task_size_reset()`.
+
 #### GDS Threshold (KVIKIO_GDS_THRESHOLD)
 In order to improve performance of small IO, `.pread()` and `.pwrite()` implement a shortcut that circumvent the threadpool and use the POSIX backend directly. Set the environment variable `KVIKIO_GDS_THRESHOLD` to the minimum size (in bytes) to use GDS. If not set, the default value is 1048576 (1 MiB).
```
wence-, line 102:
```suggestion
To improve performance of small IO requests, `.pread()` and `.pwrite()` implement a shortcut that circumvents the threadpool and uses the POSIX backend directly. Set the environment variable `KVIKIO_GDS_THRESHOLD` to the minimum size (in bytes) to use GDS. If not set, the default value is 1048576 (1 MiB).
```

(My never ending crusade to eliminate "In order to" and "Due to the fact that" from written english ;))

cpp/doxygen/main_page.md
```diff
@@ -85,15 +85,29 @@ Set the environment variable `KVIKIO_COMPAT_MODE` to enable/disable compatibilit
...
 
+This setting can also be controlled by `defaults::gds_threshold()` and `defaults::gds_threshold_reset()`.
+
+#### Size of the Bounce Buffer (KVIKIO_GDS_THRESHOLD)
+KvikIO might have to use an intermediate host buffer when copying between file and device memory. Set the environment variable ``KVIKIO_BOUNCE_BUFFER_SIZE`` to size (in bytes) of this "bounce" buffer. If not set, the default value is 16777216 (16 MiB).
```
wence-, line 107:
```suggestion
KvikIO might have to use an intermediate host buffer when copying between files and device memory. Set the environment variable ``KVIKIO_BOUNCE_BUFFER_SIZE`` to the size (in bytes) of this "bounce" buffer. If not set, the default value is 16777216 (16 MiB).
```

cpp/include/kvikio/bounce_buffer.hpp
```diff
@@ -61,29 +60,49 @@ class AllocRetain {
...
+  // Notice, we do not clear the allocations at destruction thus the allocations leaks
+  // at exit. We do this because `AllocRetain::instance()` stores the allocations in a
+  // static stack that are destructed below main, which is not allowed in CUDA:
+  // <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization>
+  ~AllocRetain() noexcept = default;
+
+  /**
+   * @brief Free all retained allocations
+   *
+   * NB: The `_mutex` must be taken prior to calling this function, if not called from the dtor.
```
wence-, line 73:
```suggestion
   * NB: The `_mutex` must be taken prior to calling this function.
```

The dtor no longer calls this function, no?

cpp/include/kvikio/bounce_buffer.hpp
```diff
@@ -61,29 +60,49 @@ class AllocRetain {
...
     while (!_free_allocs.empty()) {
       CUDA_DRIVER_TRY(cudaAPI::instance().MemFreeHost(_free_allocs.top()));
       _free_allocs.pop();
     }
+    return ret;
   }
 
-  [[nodiscard]] Alloc get()
+  /**
+   * @brief Ensure the size of the retained allocations match `defaults::bounce_buffer_size()`
```
wence-, line 88:
```suggestion
   * @brief Ensure the sizes of the retained allocations match `defaults::bounce_buffer_size()`
```

python/kvikio/kvikio/buffer.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
+
+from kvikio._lib import buffer  # type: ignore
+
+
+def memory_register(buf) -> None:
+    """Register a device memory allocation to cuFile.
```
wence-, line 8:
```suggestion
    """Register a device memory allocation with cuFile.
```

python/kvikio/kvikio/buffer.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Warning
+    -------
+    This API is intended for usecases where the memory is used as streaming
```
wence-, line 12:
```suggestion
    This API is intended for usecases where the memory is used as a streaming
```

wence-: I am not quite sure why this is annotated as a "Warning" (rather than a "Notes"), can you explain?

python/kvikio/kvikio/buffer.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Parameters
+    ----------
+    buf: buffer-like or array-like
+        Device buffer to deregister .
+    """
+    buffer.memory_deregister(buf)
+
+
+def bounce_buffer_clear() -> int:
```
wence-, line 34: *nit*: Docstring.

python/kvikio/tests/test_basic_io.py
```diff
@@ -273,3 +274,18 @@ def test_different_bounce_buffer_sizes(tmp_path, size, tasksize, buffer_size):
...
+    with kvikio.defaults.set_compat_mode(True), kvikio.defaults.set_num_threads(1):
+        with kvikio.CuFile(filename, "w") as f:
+            with kvikio.defaults.set_bounce_buffer_size(1024):
+                f.write(cupy.arange(10))  # populate the bounce buffer
+                assert kvikio.buffer.bounce_buffer_clear() == 1024
+                f.write(cupy.arange(10))  # populate the bounce buffer
+            with kvikio.defaults.set_bounce_buffer_size(2048):
+                with kvikio.CuFile(filename, "w") as f:
+                    f.write(cupy.arange(10))  # populate the bounce buffer
+        assert kvikio.buffer.bounce_buffer_clear() == 2048
```
wence-, line 291: *question*: Is this indentation right? I thought that the context manager on line 288 would reset the bounce buffer size to the default (16MiB) size, and so this clear would return `0` (because when we exited the block we had to clear the bounce buffers).

python/kvikio/tests/test_basic_io.py
```diff
@@ -273,3 +274,18 @@ def test_different_bounce_buffer_sizes(tmp_path, size, tasksize, buffer_size):
...
+    """Test clearing the bounce buffer"""
+    filename = tmp_path / "test-file"
+    with kvikio.defaults.set_compat_mode(True), kvikio.defaults.set_num_threads(1):
+        with kvikio.CuFile(filename, "w") as f:
+            with kvikio.defaults.set_bounce_buffer_size(1024):
+                f.write(cupy.arange(10))  # populate the bounce buffer
+                assert kvikio.buffer.bounce_buffer_clear() == 1024
+                f.write(cupy.arange(10))  # populate the bounce buffer
+            with kvikio.defaults.set_bounce_buffer_size(2048):
+                with kvikio.CuFile(filename, "w") as f:
```
wence-, line 289: *issue*: filename is already open for writing (via line 283).

wence- (changes requested):

wence-:
Does this also close #452?

python/kvikio/tests/test_basic_io.py
madsbk: Fixed

wence- (approved):
Thanks Mads

python/kvikio/kvikio/buffer.py
madsbk: Done

python/kvikio/tests/test_basic_io.py
madsbk: The bounce buffer isn't cleared on context exit, added the following comment:
```python
            # Notice, since the bounce buffer size is only checked when the buffer
            # is used, we populate the bounce buffer in between we clear it.
```

madsbk:
Thanks for the review @wence-

PR #455, opened 2024-09-04
Skip cufile and nvtx3 when cuda isn't found

### LINKED ISSUE #444: Very Chatty CPU-only Build
When building KvikIO without CUDA, cmake spits out a great deal of scary looking but ultimately harmless warnings. Could these be silenced?
```sh
CMake Warning at arch-conda/cmake_build/_deps/kvikio-src/cpp/CMakeLists.txt:64 (message):
    Building KvikIO without cuFile

CMake Error at arch-conda/cmake_build/_deps/kvikio-src/cpp/cmake/thirdparty/get_nvtx.cmake:18 (include):
    include could not find requested file:
  
      $SRC_DIR/arch-conda/cmake_build/_deps/rapids-cmake-src/rapids-cmake/cpm/nvtx3.cmake
  Call Stack (most recent call first):
    arch-conda/cmake_build/_deps/kvikio-src/cpp/cmake/thirdparty/get_nvtx.cmake:25 (find_and_configure_nvtx)
    arch-conda/cmake_build/_deps/kvikio-src/cpp/CMakeLists.txt:83 (include)
  
  
  CMake Error at arch-conda/cmake_build/_deps/kvikio-src/cpp/cmake/thirdparty/get_thread_pool.cmake:18 (include):
    include could not find requested file:
  
      $SRC_DIR/arch-conda/cmake_build/_deps/rapids-cmake-src/rapids-cmake/cpm/bs_thread_pool.cmake
  Call Stack (most recent call first):
    arch-conda/cmake_build/_deps/kvikio-src/cpp/cmake/thirdparty/get_thread_pool.cmake:25 (find_and_configure_thread_pool)
    arch-conda/cmake_build/_deps/kvikio-src/cpp/CMakeLists.txt:84 (include)
  
 CMake Warning at arch-conda/cmake_build/_deps/kvikio-src/cpp/CMakeLists.txt:108 (message):
    Building KvikIO without CUDA
```
It seems that if `CUDAToolKit` is not found (or, perhaps the user passes an imaginary `KvikIO_USE_CUDA=OFF`), then KvikIO could:
1. Not search for these packages in the first place, or...
2. Search for them if it must, but error silently if they are not found

madsbk:
Closes #444


cpp/CMakeLists.txt
```diff
@@ -129,7 +133,7 @@ target_include_directories(
                    "$<INSTALL_INTERFACE:include>"
 )
 target_link_libraries(
-  kvikio INTERFACE Threads::Threads ${CMAKE_DL_LIBS} nvtx3::nvtx3-cpp BS::thread_pool
+  kvikio INTERFACE Threads::Threads BS::thread_pool ${CMAKE_DL_LIBS} ${NVTX3_LIBRARIES}
```
Jacobfaib, line 136:
```suggestion
  kvikio INTERFACE Threads::Threads BS::thread_pool ${CMAKE_DL_LIBS} $<TARGET_NAME_IF_EXISTS:nvtx3::nvtx3-cpp>
```
Slightly cleaner, and doesn't rely on `NVTX3_LIBRARIES` possibly being empty.

cpp/CMakeLists.txt
```diff
@@ -55,32 +55,36 @@ rapids_find_package(
...
-  if(cuFileReadAsync_location EQUAL "-1")
-    set(cuFile_STREAM_API_FOUND FALSE)
+if(CUDAToolkit_FOUND)
+  rapids_find_package(
+    cuFile
+    BUILD_EXPORT_SET kvikio-exports
+    INSTALL_EXPORT_SET kvikio-exports
+  )
+  if(NOT cuFile_FOUND)
+    message(WARNING "Building KvikIO without cuFile")
```
Jacobfaib, line 65: IMO, this is still a bit too verbose. As a user, I don't quite know what to do with this information. So I don't have cuFile, do I want it? If so, how do I get it? And what happens if I don't have it?

Also consider the case where I know I definitely don't want cuFile (for whatever reason), there is no way to silence this warning. Now suppose I want to run my cmake files using `-Werror` (to catch cmake warnings), I have no way of doing that with this warning.

Note that `CUDAToolKit` will automatically [come shipped with cuFile](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#imported-targets).

My suggestion would be either:
- Consider making this a hard error instead (i.e. `REQUIRED`).
- Or giving the user some way to silence the warning (`KvikIO_SILENCE_MISSING_CUFILE` or something like that).

madsbk: I see you point but let's add explicit options in another PR: https://github.com/rapidsai/kvikio/issues/459.
For now, I have added:
```cmake 
message(
  WARNING "Cannot find cuFile - KvikIO will still work but won't use GPUDirect Storage (GDS)"
)
```

KyleFromNVIDIA (approved):

PR #460, opened 2024-09-10
Address segmentation fault stemming from CUDA resource destruction after cudaDeviceReset

kingcrimsontianyu:


kingcrimsontianyu:
Simply let those CUDA resources leak

PR #462, opened 2024-09-10
Intentionally leak static CUDA resources to avoid crash (part 2)

kingcrimsontianyu:
The NVbench application `PARQUET_READER_NVBENCH` in libcudf currently crashes with the segmentation fault. To reproduce:

```
./PARQUET_READER_NVBENCH -d 0 -b 1 --run-once -a io_type=FILEPATH -a compression_type=SNAPPY -a cardinality=0 -a run_length=1
```
 
The root cause is that some (1) `thread_local`  objects on the main thread in `libcudf` and (2) `static` objects in `kvikio` are destroyed after `cudaDeviceReset()` in NVbench and upon program termination. These objects should simply be leaked, since their destructors making CUDA calls upon program termination constitutes UB in CUDA.

This simple PR is the kvikIO side of the fix. The other part is done here https://github.com/rapidsai/cudf/pull/16787.

### DIFF
--- a/cpp/include/kvikio/posix_io.hpp
+++ b/cpp/include/kvikio/posix_io.hpp
@@ -41,7 +41,7 @@ class StreamsByThread {
   std::map<std::pair<CUcontext, std::thread::id>, CUstream> _streams;
 
  public:
-  StreamsByThread() = default;
+  StreamsByThread()  = default;
   ~StreamsByThread() = default;
 
   static CUstream get(CUcontext ctx, std::thread::id thd_id)
--- a/cpp/include/kvikio/posix_io.hpp
+++ b/cpp/include/kvikio/posix_io.hpp
@@ -41,7 +41,12 @@ class StreamsByThread {
   std::map<std::pair<CUcontext, std::thread::id>, CUstream> _streams;
 
  public:
-  StreamsByThread()  = default;
+  StreamsByThread() = default;
+
+  // Here we intentionally do not destroy in the destructor the CUDA resources
+  // (e.g. CUstream) with static storage duration, but instead let them leak
+  // on program termination. This is to prevent undefined behavior in CUDA. See
+  // <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization>
   ~StreamsByThread() = default;
 
   static CUstream get(CUcontext ctx, std::thread::id thd_id)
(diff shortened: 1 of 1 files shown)


cpp/include/kvikio/posix_io.hpp
```diff
@@ -41,17 +41,8 @@ class StreamsByThread {
   std::map<std::pair<CUcontext, std::thread::id>, CUstream> _streams;
 
  public:
-  StreamsByThread() = default;
-  ~StreamsByThread() noexcept
-  {
-    for (auto& [_, stream] : _streams) {
-      try {
-        CUDA_DRIVER_TRY(cudaAPI::instance().StreamDestroy(stream));
-      } catch (const CUfileException& e) {
```
vuule, line 50: it's interesting that this causes segfaults even with the catch.

kingcrimsontianyu: If I didn't miss anything, this is expected? I checked kvikio and cudf, and didn't find a part where we convert the `SIGSEGV` signal into a C++ exception. 

[This reference](https://en.cppreference.com/w/cpp/utility/program/signal) states that
> If the user defined function returns when handling SIGFPE, SIGILL, SIGSEGV or any other implementation-defined signal specifying a computational exception, the behavior is undefined.

This indicates the signal handler must cause the program to exit in a certain way (abnormally with `std::abort` or normally with `std::_Exit`). So I guess it's not quite possible (?) to handle the `SIGSEGV` "exception" after all without causing any UB problem.

cpp/include/kvikio/posix_io.hpp
```diff
@@ -41,17 +41,8 @@ class StreamsByThread {
...
-        std::cerr << e.what() << std::endl;
-      }
-    }
-  }
+  StreamsByThread()  = default;
+  ~StreamsByThread() = default;
```
madsbk, line 45: It would be good with a comment and an URL to the issue in the CUDA programming guide

madsbk (changes requested):
Thanks @kingcrimsontianyu looks good. I only have one suggestion

cpp/include/kvikio/posix_io.hpp
kingcrimsontianyu: Good idea! I saw other source files in kvikio doing the same thing. I'll follow suit.

kingcrimsontianyu: Completed.

madsbk (approved):
Thanks @kingcrimsontianyu

PR #464, opened 2024-09-18
Remote IO: http support

madsbk:
Support read directly from a http server like:
```python
import kvikio
import cupy

with kvikio.RemoteFile.from_http_url("http://127.0.0.1:9000/myfile") as f:
    ary = cupy.empty(f.nbytes, dtype="uint8")
    f.read(ary)
```

This PR is the first step to support S3 using libcurl instead of [aws-s3-sdk](https://github.com/rapidsai/kvikio/pull/426), which has some pros and cons:

* Pros 
    * The [global conda pinning issue](https://github.com/rapidsai/kvikio/pull/426#discussion_r1752688778) is less of a problem.
    * We can support other protocols such as http, ftp, and Azure’s storage, without much work. 
    * We avoid the [free-after-main issue in aws-s3-sdk](https://github.com/rapidsai/kvikio/blob/000126516db430988ab9af5ee1576ca3fe6afe27/cpp/include/kvikio/remote_handle.hpp#L87-L94). This is huge since we would otherwise have to pass around a `S3Context` in libcudf and cudf to handle shutdown correctly.  This is not a problem in libcurl, see https://curl.se/libcurl/c/libcurl.html under `Global constants`.
* Cons 
    * Hard to support the AWS configuration file. We will require the user to either specify the options programmatically or through environment variables like `AWS_ACCESS_KEY_ID ` and `AWS_SECRET_ACCESS_KEY `.


dependencies.yaml
```diff
@@ -106,6 +106,7 @@ dependencies:
         packages:
           - c-compiler
           - cxx-compiler
+          - libcurl>=7.87.0  # Need <https://curl.se/libcurl/c/CURLOPT_WRITEFUNCTION.html>
```
wence-, line 109: That manpage says WRITEFUNCTION was added in curl 7.1 ?

wence- (commented):
C++, not python yet

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,381 @@
...
+
+  CurlHandle(CurlHandle const&)            = delete;
+  CurlHandle& operator=(CurlHandle const&) = delete;
+  CurlHandle(CurlHandle&& o)               = delete;
+  CurlHandle& operator=(CurlHandle&& o)    = delete;
+
+  CURL* handle() noexcept { return _handle; }
+
+  template <typename OPT, typename VAL>
+  void setopt(OPT option, VAL value)
```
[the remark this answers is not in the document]
madsbk: Fixed

dependencies.yaml
madsbk: need `CURL_WRITEFUNC_ERROR`, add: [ecd3311](https://github.com/rapidsai/kvikio/pull/464/commits/ecd3311ad58da6f18def68b10db5a3d1ec8e150c)

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,381 @@
...
+  {
+    // Need CURLOPT_NOSIGNAL to support threading, see
+    // <https://curl.se/libcurl/c/CURLOPT_NOSIGNAL.html>
+    setopt(CURLOPT_NOSIGNAL, 1L);
+
+    // We always set CURLOPT_ERRORBUFFER to get better error messages.
+    setopt(CURLOPT_ERRORBUFFER, _errbuf);
+
+    // Make curl_easy_perform() fail when receiving HTTP code errors.
+    setopt(CURLOPT_FAILONERROR, 1L);
```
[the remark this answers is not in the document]
madsbk: Yes, good point. Done.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,381 @@
...
+  PushAndPopContext c(cuda_ctx);
+  CUstream stream = detail::StreamsByThread::get();
+
+  // std::cout << "callback_device_memory() - data: " << ((void*)data)
+  //           << ", ctx->buf: " << (void*)ctx->buf << ", offset: " << ctx->offset
+  //           << ", nbytes: " << nbytes << std::endl;
+
+  CUDA_DRIVER_TRY(cudaAPI::instance().MemcpyHtoDAsync(
+    convert_void2deviceptr(ctx->buf + ctx->offset), data, nbytes, stream));
+  CUDA_DRIVER_TRY(cudaAPI::instance().StreamSynchronize(stream));
```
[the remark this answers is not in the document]
madsbk: fixed

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,381 @@
...
+      throw std::runtime_error("cannot initialize libcurl - built with thread safety disabled");
+    }
+  }
+  ~LibCurl() noexcept
+  {
+    // clean up all retained easy handles
+    while (!_free_curl_handles.empty()) {
+      curl_easy_cleanup(_free_curl_handles.top());
+      _free_curl_handles.pop();
+    }
```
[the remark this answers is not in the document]
madsbk: Good point, done in ce190a7dca2bb4979c383d915d36f5ac7da183a1

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,381 @@
...
+    curl_easy_reset(ret);
+    return ret;
+  }
+
+  void put(CURL* handle)
+  {
+    std::lock_guard const lock(_mutex);
+    _free_curl_handles.push(handle);
+  }
```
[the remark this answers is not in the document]
madsbk: done in 76931a3ca96a357ab5237a3a4efdb7430d195a17

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,425 @@
...
+
+  LibCurl(LibCurl const&)            = delete;
+  LibCurl& operator=(LibCurl const&) = delete;
+  LibCurl(LibCurl&& o)               = delete;
+  LibCurl& operator=(LibCurl&& o)    = delete;
+
+ public:
+  static LibCurl& instance()
+  {
+    static LibCurl _instance;
```
wence-, line 91: *question/thought*: Since this is an object with static lifetime, the destructor will run below main. I had a read through the libcurl documentation and did not manage to determine if this is safe. They don't explicitly call it out as unsafe, so maybe it's fine?

madsbk: I think this is what they are saying in the docs? From https://curl.se/libcurl/c/libcurl.html:

> In a C++ module, it is common to deal with the global constant situation by defining a special class that represents the global constant environment of the module. A program always has exactly one object of the class, in static storage. That way, the program automatically calls the constructor of the object as the program starts up and the destructor as it terminates. As the author of this libcurl-using module, you can make the constructor call [curl_global_init](https://curl.se/libcurl/c/curl_global_init.html) and the destructor call [curl_global_cleanup](https://curl.se/libcurl/c/curl_global_cleanup.html) and satisfy libcurl's requirements without your user having to think about it. (Caveat: If you are initializing libcurl from a Windows DLL you should not initialize it from DllMain or a static initializer because Windows holds the loader lock during that time and it could cause a deadlock.)

wence-: Thanks, I had a look at the code too and there's nothing obviously there that will bite, AFAICT (curl only registers an `atexit` callback on amigaos)

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,381 @@
...
+  curl.setopt(CURLOPT_FOLLOWLOCATION, 1L);
+  curl.perform();
+
+  curl_off_t cl;
+  curl.getinfo(CURLINFO_CONTENT_LENGTH_DOWNLOAD_T, &cl);
+  if (cl < 0) {
+    throw std::runtime_error("cannot get size of " + url +
+                             ", content-length not provided by the server");
+  }
+  return cl;
```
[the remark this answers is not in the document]
madsbk: I think I prefer `RemoteHandle::nbytes()`, which makes the unit explicit

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+  }
+};
+
+/**
+ * @brief Create a new curl handle.
+ *
+ * @returns A `kvikio::CurlHandle` instance ready to be used.
+ */
+#define create_curl_handle() \
+  kvikio::CurlHandle(kvikio::LibCurl::instance().get_handle(), __FILE__, KVIKIO_STRINGIFY(__LINE__))
```
wence-, line 258: *note*: This macro is only needed to get access to `__FILE__` and `__LINE__`? Note that once we go to C++20 we can use [`std::source_location`](https://en.cppreference.com/w/cpp/utility/source_location)

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+  char _errbuf[CURL_ERROR_SIZE];
+  LibCurl::UniqueHandlePtr _handle;
+  std::string _source_file;
+  std::string _source_line;
+
+ public:
+  /**
+   * @brief Construct a new curl handle.
+   *
+   * Typically, do not use this directly instead use the `create_curl_handle()` macro.
```
wence-, line 153: Should it ever be the case that one constructs a `CurlHandle` directly?

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+    std::lock_guard const lock(_mutex);
+    _free_curl_handles.push_back(std::move(handle));
+  }
+};
+
+/**
+ * @brief Representation of a curl easy handle pointer and its operations.
+ *
+ * An instance is given a `LibCurl::UniqueHandlePtr` on creation, which is
+ * later retailed on destruction.
```
wence-, line 140:
```suggestion
 * later retained on destruction.
```
?

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+    // <https://curl.se/libcurl/c/CURLOPT_NOSIGNAL.html>
+    setopt(CURLOPT_NOSIGNAL, 1L);
+
+    // We always set CURLOPT_ERRORBUFFER to get better error messages.
+    setopt(CURLOPT_ERRORBUFFER, _errbuf);
+
+    // Make curl_easy_perform() fail when receiving HTTP code errors.
+    setopt(CURLOPT_FAILONERROR, 1L);
+  }
+  ~CurlHandle() noexcept { LibCurl::instance().retain_handle(std::move(_handle)); }
```
wence-, line 174: *question*: One thing I wonder is whether this is ahead-of-time optimisation, or if this mechanism of keeping hold of curl handles is a recommended pattern?

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+ */
+class LibCurl {
+ public:
+  // We hold an unique pointer to the raw curl handle and sets `curl_easy_cleanup` as its Deleter.
+  using UniqueHandlePtr = std::unique_ptr<CURL, std::function<void(CURL*)>>;
+
+ private:
+  std::mutex _mutex{};
+  // Curl handles free to be used.
+  std::vector<UniqueHandlePtr> _free_curl_handles{};
```
wence-, line 66: *question*: In theory, the caller could create endless curl handles and never release them, do we care about that case?

python/kvikio/kvikio/_lib/CMakeLists.txt
```diff
@@ -17,10 +17,20 @@ set(cython_modules arr.pyx buffer.pyx defaults.pyx driver_properties.pyx file_ha
...
 )
 
+if(TARGET CURL::libcurl)
+  message(STATUS "Building remote_handle.pyx (libcurl found)")
+  list(APPEND cython_modules remote_handle.pyx)
+else()
+  message(
+    WARNING
+      "Skipping remote_handle.pyx (please set KvikIO_REMOTE_SUPPORT=ON for remote file support)"
+  )
```
wence-, line 27: As with the previous discussions/issues around noisiness, should we actually warn here? Or just emit a `STATUS` message?

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -0,0 +1,89 @@
...
+
+
+cdef extern from "<kvikio/remote_handle.hpp>" nogil:
+    cdef cppclass cpp_RemoteEndpoint "kvikio::RemoteEndpoint":
+        pass
+
+    cdef cppclass cpp_HttpEndpoint "kvikio::HttpEndpoint":
+        cpp_HttpEndpoint(string url) except +
+
+    cdef cppclass cpp_RemoteHandle "kvikio::RemoteHandle":
```
wence-, line 26:
```suggestion
cdef extern from "<kvikio/remote_handle.hpp>" nogil namespace "kvikio":
    cdef cppclass RemoteEndpoint:
        pass

    cdef cppclass HttpEndpoint:
        HttpEndpoint(string url) except +

    cdef cppclass RemoteHandle:
```

?
Since `RemoteEndPoint`/`HttpEndpoint` and `RemoteHandle` are not used by any python cdef class names.

python/kvikio/kvikio/remote_file.py
```diff
@@ -0,0 +1,123 @@
...
+            The size of the file. If None, KvikIO will ask the server
+            for the file size.
+        """
+        return RemoteFile(_get_remote_module().RemoteFile.open_http(url, nbytes))
+
+    def __enter__(self) -> RemoteFile:
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        pass
```
wence-, line 75: Question. Should there be a `close` method that releases the handle, and then it is called in `__exit__`? This would kind of match usage of the builtin `open`.

python/kvikio/kvikio/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        self.root_path = root_path
+        self.range_support = range_support
+        self.max_lifetime = max_lifetime
+
+    def __enter__(self):
+        queue = multiprocessing.Queue()
+        self.process = multiprocessing.Process(
+            target=LocalHttpServer._server,
+            args=(queue, str(self.root_path), self.range_support, self.max_lifetime),
+        )
```
wence-, line 72: *question*: Since we're starting a new process here to run the server, why do we _also_ run the server in its own thread in `_server`?

python/kvikio/tests/test_examples.py
```diff
@@ -26,3 +26,10 @@ def test_zarr_cupy_nvcomp(tmp_path, monkeypatch):
 
     monkeypatch.syspath_prepend(str(examples_path))
     import_module("zarr_cupy_nvcomp").main(tmp_path / "test-file")
+
+
+def test_http_io(tmp_path, monkeypatch):
+    """Test examples/http_io.py"""
+
+    monkeypatch.syspath_prepend(str(examples_path))
+    import_module("http_io").main(tmp_path)
```
wence-, line 35: Does this need to skip the test if remote file support is not available?

wence- (commented):
Minor comments, overall looking good to me

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+  std::string _source_file;
+  std::string _source_line;
+
+ public:
+  /**
+   * @brief Construct a new curl handle.
+   *
+   * Typically, do not use this directly instead use the `create_curl_handle()` macro.
+   *
+   * @param handle An unused curl easy handle pointer, which is retailed on destruction.
```
bdice, line 155:
```suggestion
   * @param handle An unused curl easy handle pointer, which is retained on destruction.
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,289 @@
...
+inline std::size_t callback_host_memory(char* data,
+                                        std::size_t size,
+                                        std::size_t nmemb,
+                                        void* context)
+{
+  auto ctx                 = reinterpret_cast<CallbackContext*>(context);
+  const std::size_t nbytes = size * nmemb;
+  if (ctx->size < ctx->offset + nbytes) {
+    ctx->overflow_error = true;
+    return CURL_WRITEFUNC_ERROR;
```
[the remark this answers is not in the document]
madsbk: `nbytes == CURL_WRITEFUNC_ERROR` cannot happen,  `CURL_WRITEFUNC_ERROR` is defined as `0xFFFFFFFF`, which is greater than `CURL_MAX_WRITE_SIZE`.

cpp/include/kvikio/shim/libcurl.hpp
madsbk: No :)

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+ * to think about it. (Caveat: If you are initializing libcurl from a Windows DLL you should not
+ * initialize it from DllMain or a static initializer because Windows holds the loader lock during
+ * that time and it could cause a deadlock.)
+ *
+ * Source <https://curl.se/libcurl/c/libcurl.html>.
+ */
+class LibCurl {
+ public:
+  // We hold an unique pointer to the raw curl handle and sets `curl_easy_cleanup` as its Deleter.
+  using UniqueHandlePtr = std::unique_ptr<CURL, std::function<void(CURL*)>>;
```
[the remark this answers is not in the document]
madsbk: Done

cpp/include/kvikio/shim/libcurl.hpp
madsbk: I guess if you want to specify some other source and/or line number?

cpp/include/kvikio/shim/libcurl.hpp
madsbk: I haven't benchmark it but the docs says:
> The easy handle is used to hold and control a single network transfer. It is encouraged to reuse easy handles for repeated transfers.

madsbk: https://curl.se/libcurl/c/curl_easy_init.html

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -0,0 +1,260 @@
...
+   *
+   * See <https://curl.se/libcurl/c/curl_easy_getinfo.html> for available options.
+   *
+   * @tparam OUTPUT The type of the output.
+   * @param output The output, which is used as-is: `curl_easy_getinfo(..., output)`.
+   */
+  template <typename OUTPUT>
+  void getinfo(CURLINFO info, OUTPUT output)
+  {
+    CURLcode err = curl_easy_getinfo(handle(), info, output);
```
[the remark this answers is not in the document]
madsbk: Done

python/kvikio/kvikio/_lib/CMakeLists.txt
madsbk: Good point, done

python/kvikio/kvikio/_lib/remote_handle.pyx
madsbk: I keep mixing C++ and Python objects when reading Cython, I think always using `cpp_` makes it a bit easier?

python/kvikio/kvikio/remote_file.py
madsbk: Agree, done

python/kvikio/tests/test_examples.py
madsbk: Good catch, fixed

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,289 @@
...
+namespace kvikio {
+namespace detail {
+
+/**
+ * @brief Context used by the "CURLOPT_WRITEFUNCTION" callbacks.
+ */
+struct CallbackContext {
+  char* buf;            // Output buffer to read into.
+  std::size_t size;     // Total number of bytes to read.
+  std::size_t offset;   // Offset into `buf` to start reading.
```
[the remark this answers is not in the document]
madsbk: Done

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,291 @@
...
+      std::to_string(file_offset) + "-" + std::to_string(file_offset + size - 1);
+    curl.setopt(CURLOPT_RANGE, byte_range.c_str());
+
+    if (is_host_mem) {
+      curl.setopt(CURLOPT_WRITEFUNCTION, detail::callback_host_memory);
+    } else {
+      curl.setopt(CURLOPT_WRITEFUNCTION, detail::callback_device_memory);
+    }
+    detail::CallbackContext ctx{
+      .buf = reinterpret_cast<char*>(buf), .size = size, .offset = 0, .overflow_error = false};
```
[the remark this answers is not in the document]
madsbk: done

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -0,0 +1,291 @@
...
+   * @brief Read from remote source into buffer (host or device memory) in parallel.
+   *
+   * This API is a parallel async version of `.read()` that partitions the operation
+   * into tasks of size `task_size` for execution in the default thread pool.
+   *
+   * @param buf Pointer to host or device memory.
+   * @param size Number of bytes to read.
+   * @param file_offset File offset in bytes.
+   * @param task_size Size of each task in bytes.
+   * @return Number of bytes read, which is `size` always.
```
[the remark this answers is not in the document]
madsbk: done

cpp/cmake/thirdparty/get_libcurl.cmake
```diff
@@ -0,0 +1,32 @@
...
+  include(${rapids-cmake-dir}/cpm/find.cmake)
+
+  rapids_cpm_find(
+    CURL 7.87.0
+    GLOBAL_TARGETS libcurl
+    BUILD_EXPORT_SET kvikio-exports
+    INSTALL_EXPORT_SET kvikio-exports
+    CPM_ARGS
+    GIT_REPOSITORY https://github.com/curl/curl
+    GIT_TAG curl-8_10_1 # let's get a newer release if we have to build it ourselves anyways.
```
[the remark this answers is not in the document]
madsbk: done

conda/recipes/libkvikio/meta.yaml
```diff
@@ -52,6 +52,7 @@ requirements:
     {% else %}
     - libcufile-dev  # [linux]
     {% endif %}
+    - libcurl>=7.87.0
```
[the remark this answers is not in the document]
madsbk: Thanks for the clarification, I understand `meta.yaml` a little bit better now :)

conda/recipes/libkvikio/meta.yaml
```diff
@@ -83,6 +84,7 @@ outputs:
         {% else %}
         - libcufile-dev  # [linux]
         {% endif %}
+        - libcurl>=7.87.0
```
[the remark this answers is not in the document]
madsbk: done

cpp/CMakeLists.txt
```diff
@@ -231,6 +240,12 @@ if(NOT already_set_kvikio)
       target_compile_definitions(kvikio::kvikio INTERFACE KVIKIO_CUFILE_STREAM_API_FOUND)
     endif()
   endif()
+
+  if(TARGET CURL::libcurl)
+    target_link_libraries(kvikio::kvikio INTERFACE $<BUILD_LOCAL_INTERFACE:CURL::libcurl>)
+    target_compile_definitions(kvikio::kvikio INTERFACE $<BUILD_LOCAL_INTERFACE:KVIKIO_LIBCURL_FOUND>)
```
KyleFromNVIDIA, line 246: Since this is code that gets exported out of the build, it does not need the `$<BUILD_LOCAL_INTERFACE>` genex.

```suggestion
    target_link_libraries(kvikio::kvikio INTERFACE CURL::libcurl)
    target_compile_definitions(kvikio::kvikio INTERFACE KVIKIO_LIBCURL_FOUND)
```

python/kvikio/kvikio/_lib/CMakeLists.txt
```diff
@@ -17,10 +17,20 @@ set(cython_modules arr.pyx buffer.pyx defaults.pyx driver_properties.pyx file_ha
...
+    STATUS
+      "Skipping remote_handle.pyx (please set KvikIO_REMOTE_SUPPORT=ON for remote file support)"
+  )
+endif()
+
 rapids_cython_create_modules(
   CXX
   SOURCE_FILES "${cython_modules}"
-  LINKED_LIBRARIES kvikio::kvikio nvcomp::nvcomp
+  LINKED_LIBRARIES kvikio::kvikio nvcomp::nvcomp $<TARGET_NAME_IF_EXISTS:CURL::libcurl>
```
KyleFromNVIDIA, line 33: Since `kvikio` already exports `CURL::libcurl` as an `INTERFACE` dependency, it shouldn't be needed here.

KyleFromNVIDIA (changes requested):

KyleFromNVIDIA (approved):

python/kvikio/kvikio/_lib/remote_handle.pyx
wence-: Sure (nothing to do).

cpp/include/kvikio/remote_handle.hpp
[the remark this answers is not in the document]
wence-: Ah, thanks.

cpp/include/kvikio/shim/libcurl.hpp
wence-: Thanks.

wence- (approved):
Nothing more from me, looks good.

PR #472, opened 2024-09-25
Fix release-after-main bug in `get_primary_cuda_context` 

### LINKED ISSUE #442: [BUG] `get_primary_cuda_context` static map is unsafe in three ways
1. The `CudaPrimaryContext` object has a destructor which calls into the driver (`cuDevicePrimaryCtxRelease`). But since these objects are stored in a `static` map, the destructor will run below main, and this is UB [per the cuda programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization)
2. `get_primary_cuda_context` is not thread-safe since access to the `static` map is not protected by a lock.
3. (This came out of reviewing rapidsai/rmm#1598)

Since kvikio is header-only, we rely on the linker to disambiguate `inline` functions that have (or return) static references. To do this, the relevant function must have `__attribute__((visibility("default")))`. If not, then if kvikio is used in two different DSOs, the function will appear twice, and there will be two static objects.

By default, I believe all rapids libraries are compiled with `-fvisiblity-inline-hidden`. So `kvikio::get_primary_cuda_context` (which is `inline`) will be `__attribute((visibility("hidden")))`. Thus, if kvikio is used from two separate DSOs, calls from the respective DSOs to `get_primary_cuda_context` will reference different global statics (since the functions are not the same).

To solve these:

1. We just need to leak the contexts, I think.
2. We can add a lock
3. We should explicitly mark this function as `__attribute__((visibility("default")))`

madsbk:
Closes #442

### DIFF
--- a/cpp/include/kvikio/utils.hpp
+++ b/cpp/include/kvikio/utils.hpp
@@ -153,6 +153,9 @@ class CudaPrimaryContext {
 [[nodiscard]] KVIKIO_EXPORT inline CUcontext get_primary_cuda_context(int ordinal)
 {
   static std::map<int, CudaPrimaryContext> _primary_contexts;
+  static std::mutex _mutex;
+  std::lock_guard const lock(_mutex);
+
   _primary_contexts.try_emplace(ordinal, ordinal);
   return _primary_contexts.at(ordinal).ctx;
 }
--- a/cpp/include/kvikio/utils.hpp
+++ b/cpp/include/kvikio/utils.hpp
@@ -28,6 +28,20 @@
 #include <kvikio/error.hpp>
 #include <kvikio/shim/cuda.hpp>
 
+// Macros used for defining symbol visibility, only GLIBC is supported.
+// Since KvikIO is header-only, we rely on the linker to disambiguate inline functions
+// that have (or return) static references. To do this, the relevant function must have
+// `__attribute__((visibility("default")))`. If not, then if KvikIO is used in two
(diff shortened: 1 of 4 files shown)


cpp/include/kvikio/utils.hpp
```diff
@@ -28,6 +28,20 @@
 #include <kvikio/error.hpp>
 #include <kvikio/shim/cuda.hpp>
 
+// Macros used for defining symbol visibility, only GLIBC is supported.
+// Since KvikIO is header-only, we rely on the linker to disambiguate inline functions
+// that have (or return) static references. To do this, the relevant function must have
+// `__attribute__((visibility("default")))`. If not, then if KvikIO is used in two
+// different DSOs, the function will appear twice, and there will be two static objects.
+// See <https://github.com/rapidsai/kvikio/issues/442>.
+#if (defined(__GNUC__) && !defined(__MINGW32__) && !defined(__MINGW64__))
```
wence-, line 37: *nit*: Can we also support clang (https://clang.llvm.org/docs/AttributeReference.html#visibility) given the various people compiling on macos?

wence- (approved):

cpp/include/kvikio/utils.hpp
madsbk: Actually, clang also defines `__GNUC__` by default but I guess adding `defined(__clang__)` is good since `__attribute__((visibility(...)))` is valid clang code as well.

PR #479, opened 2024-10-02
Remote IO: S3 support 

madsbk:
Implements AWS S3 read support using libcurl:
```python
import kvikio
import cupy

with kvikio.RemoteFile.from_s3_url("s://my-bucket/my-file") as f:
    ary = cupy.empty(f.nbytes, dtype="uint8")
    f.read(ary)
```

Supersedes https://github.com/rapidsai/kvikio/pull/426

### DIFF
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -146,6 +146,11 @@ class HttpEndpoint : public RemoteEndpoint {
   std::string _url;
 
  public:
+  /**
+   * @brief Create a http endpoint from a url.
+   *
+   * @param url The full http url to the remote file.
+   */
   HttpEndpoint(std::string url) : _url{std::move(url)} {}
   void setopt(CurlHandle& curl) override { curl.setopt(CURLOPT_URL, _url.c_str()); }
   std::string str() override { return _url; }
@@ -153,7 +158,7 @@ class HttpEndpoint : public RemoteEndpoint {
 };
 
 /**
- * @brief
+ * @brief A remote endpoint using AWS's S3 protocol.
  */
 class S3Endpoint : public RemoteEndpoint {
  private:
@@ -182,10 +187,7 @@ class S3Endpoint : public RemoteEndpoint {
                                                 std::optional<std::string> aws_endpoint_url)
   {
     std::string endpoint_url =
-      parse_aws_argument(std::move(aws_endpoint_url),
-                         "AWS_ENDPOINT_URL",
-                         "S3: must provide `aws_endpoint_url` if AWS_ENDPOINT_URL isn't set.",
-                         true);
+      parse_aws_argument(std::move(aws_endpoint_url), "AWS_ENDPOINT_URL", "", true);
     std::stringstream ss;
     if (endpoint_url.empty()) {
       std::string region =
@@ -229,6 +231,19 @@ class S3Endpoint : public RemoteEndpoint {
     return std::make_pair(std::move(bucket_name), std::move(object_name));
   }
 
+  /**
+   * @brief Create a S3 endpoint from a url.
+   *
+   * @param url The full http url to the S3 file. NB: this should an url starting with
+   * "http://" or "https://". If you have an S3 url of the form "s3://<bucket>/<object>",
+   * please use `S3Endpoint::parse_s3_url()` to convert it.
+   * @param aws_region The AWS region, such as "us-east-1", to use. If nullopt, the value of the
+   * `AWS_DEFAULT_REGION` environment variable is used.
+   * @param aws_access_key The AWS access key to use. If nullopt, the value of the
+   * `AWS_ACCESS_KEY_ID` environment variable is used.
+   * @param aws_secret_access_key The AWS secret access key to use. If nullopt, the value of the
+   * `AWS_SECRET_ACCESS_KEY` environment variable is used.
+   */
   S3Endpoint(std::string url,
              std::optional<std::string> aws_region            = std::nullopt,
              std::optional<std::string> aws_access_key        = std::nullopt,
@@ -263,6 +278,22 @@ class S3Endpoint : public RemoteEndpoint {
       _aws_userpwd = ss.str();
     }
   }
+
+  /**
+   * @brief Create a S3 endpoint from a bucket and object name.
+   *
+   * @param bucket_name The name of the S3 bucket.
+   * @param object_name The name of the S3 object.
+   * @param aws_region The AWS region, such as "us-east-1", to use. If nullopt, the value of the
+   * `AWS_DEFAULT_REGION` environment variable is used.
+   * @param aws_access_key The AWS access key to use. If nullopt, the value of the
+   * `AWS_ACCESS_KEY_ID` environment variable is used.
+   * @param aws_secret_access_key The AWS secret access key to use. If nullopt, the value of the
+   * `AWS_SECRET_ACCESS_KEY` environment variable is used.
+   * @param aws_endpoint_url Overwrite the endpoint url to use. If nullopt, the value of
+   * the `AWS_ENDPOINT_URL` environment variable is used. If this is also not set, the regular
+   * AWS url scheme is used: "https://<bucket>.s3.<region>.amazonaws.com/<object>"
+   */
   S3Endpoint(const std::string& bucket_name,
              const std::string& object_name,
              std::optional<std::string> aws_region            = std::nullopt,
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -194,6 +194,7 @@ class S3Endpoint : public RemoteEndpoint {
     return std::string(env);
   }
 
+ public:
   /**
    * @brief Get url from a AWS S3 bucket and object name.
    *
@@ -229,7 +230,6 @@ class S3Endpoint : public RemoteEndpoint {
     return ss.str();
   }
 
- public:
   /**
    * @brief Given an url like "s3://<bucket>/<object>", return the name of the bucket and object.
    *
@@ -251,8 +251,8 @@ class S3Endpoint : public RemoteEndpoint {
    * @brief Create a S3 endpoint from a url.
    *
    * @param url The full http url to the S3 file. NB: this should be an url starting with
-   * "http://" or "https://". If you have an S3 url of the form "s3://<bucket>/<object>",
-   * please use `S3Endpoint::parse_s3_url()` to convert it.
+   * "http://" or "https://". If you have an S3 url of the form "s3://<bucket>/<object>", please
+   * use `S3Endpoint::parse_s3_url()` and `S3Endpoint::url_from_bucket_and_object() to convert it.
    * @param aws_region The AWS region, such as "us-east-1", to use. If nullopt, the value of the
    * `AWS_DEFAULT_REGION` environment variable is used.
    * @param aws_access_key The AWS access key to use. If nullopt, the value of the
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -18,6 +18,7 @@
 #include <cstddef>
 #include <cstring>
 #include <memory>
+#include <optional>
 #include <sstream>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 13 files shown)


bdice (approved):
Very nice work. I don't have any suggestions from a quick look. It'd be good to have a second pair of eyes, possibly from @wence-?

madsbk:
> Very nice work. I don't have any suggestions from a quick look. It'd be good to have a second pair of eyes, possibly from @wence-?

Thanks @bdice. 
Agree, @wence- if you can find time, it would be good with a review.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+    size_t pos              = p.find_first_of('/');
+    std::string bucket_name = p.substr(0, pos);
+    if (bucket_name.empty()) {
+      throw std::invalid_argument("The S3 url does not contain a bucket name.");
+    }
+    std::string object_name = (pos == std::string::npos) ? "" : p.substr(pos + 1);
+    if (object_name.empty()) {
+      throw std::invalid_argument("The S3 url does not contain an object name.");
+    }
+    return std::make_pair(std::move(bucket_name), std::move(object_name));
```
vyasr, line 231: This feels like a lot of work to avoid a regex :joy: 
```suggestion
    // Regular expression to match s3://<bucket>/<object>
    std::regex pattern{R"(s3://([^/]+)/(.+))"};
    std::smatch matches;
    if (std::regex_match(s3_url, matches, pattern)) {
      return {matches[1].str(), matches[2].str()};
    }
    throw std::invalid_argument("Input string does not match the expected S3 URL format.");
```
You'll have to `#include <regex>` as well.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+    }
+    return std::string(env);
+  }
+
+  static std::string url_from_bucket_and_object(std::string const& bucket_name,
+                                                std::string const& object_name,
+                                                std::optional<std::string> const& aws_region,
+                                                std::optional<std::string> aws_endpoint_url)
+  {
+    std::string endpoint_url =
```
vyasr, line 189: May as well deduce all these types and make them const where appropriate.
```suggestion
    auto const endpoint_url =
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+      parse_aws_argument(std::move(aws_endpoint_url), "AWS_ENDPOINT_URL", "", true);
+    std::stringstream ss;
+    if (endpoint_url.empty()) {
+      std::string region =
```
vyasr, line 193:
```suggestion
      auto const region =
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+   * @param aws_secret_access_key The AWS secret access key to use. If nullopt, the value of the
+   * `AWS_SECRET_ACCESS_KEY` environment variable is used.
+   */
+  S3Endpoint(std::string url,
+             std::optional<std::string> aws_region            = std::nullopt,
+             std::optional<std::string> aws_access_key        = std::nullopt,
+             std::optional<std::string> aws_secret_access_key = std::nullopt)
+    : _url{std::move(url)}
+  {
+    std::string region =
```
vyasr, line 253:
```suggestion
    auto const region =
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+      parse_aws_argument(std::move(aws_region),
+                         "AWS_DEFAULT_REGION",
+                         "S3: must provide `aws_region` if AWS_DEFAULT_REGION isn't set.");
+
+    std::string access_key =
```
vyasr, line 258:
```suggestion
    auto const access_key =
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+      parse_aws_argument(std::move(aws_access_key),
+                         "AWS_ACCESS_KEY_ID",
+                         "S3: must provide `aws_access_key` if AWS_ACCESS_KEY_ID isn't set.");
+
+    std::string secret_access_key = parse_aws_argument(
```
vyasr, line 263:
```suggestion
    auto const secret_access_key = parse_aws_argument(
```

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -23,6 +23,15 @@ cdef extern from "<kvikio/remote_handle.hpp>" nogil:
     cdef cppclass cpp_HttpEndpoint "kvikio::HttpEndpoint":
         cpp_HttpEndpoint(string url) except +
 
+    cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint":
```
vyasr, line 26: It's not strictly in scope for this PR, but why not add a namespace clause to the `cdef extern` on line 19 so that you don't have to manually alias each of these types? You can also drop the `cpp_` prefixes then.

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -67,6 +76,59 @@ cdef class RemoteFile:
         ret._handle = make_unique[cpp_RemoteHandle](move(ep), n)
         return ret
 
+    @classmethod
```
vyasr, line 79: There is a lot of logic repeated in these three functions and it would benefit from adding a single helper with most of the logic that you could just pass the endpoint created since the creation of `ep` is really the only difference.

python/kvikio/kvikio/benchmarks/s3_io.py
vyasr, line 1: Glancing through the other Python benchmarks it looks like there is some copy-pasting going on that we might be able to include. Could we try and consolidate some of the logic? For instance, we could add some common helper functions for specific arguments (like the dtype), or [argument groups](https://docs.python.org/3/library/argparse.html#argument-groups) for related arguments. We could use [`rich`](https://rich.readthedocs.io/en/stable/introduction.html) to simplify and prettify some of the table output.

I'm OK not going too far in this PR since I don't want to hold things up, but I would encourage at least trying to get `main` a little consolidated between the different benchmark scripts.

vyasr (changes requested):
Sorry I didn't get around to reviewing #464, I took this opportunity to familiarize myself with that code and then reviewed here.

cpp/include/kvikio/remote_handle.hpp
madsbk: Heh, thanks :)

python/kvikio/kvikio/_lib/remote_handle.pyx
madsbk: Heh, @wence- asked the same thing. I just like the `cpp_` so I don't mix up what is C++ and what is Cython/Python. This also makes it easier to search the codebase.

python/kvikio/kvikio/benchmarks/s3_io.py
madsbk: Agree, the benchmarks could use some clean up and consolidation :)
But I think we should leave this benchmark as-it and do a follow-up PR that looks at all of the benchmarks in unison? Added an issue to track this: https://github.com/rapidsai/kvikio/issues/496

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+/**
+ * @brief A remote endpoint using AWS's S3 protocol.
+ */
+class S3Endpoint : public RemoteEndpoint {
+ private:
+  std::string _url;
+  std::string _aws_sigv4;
+  std::string _aws_userpwd;
+
+  static std::string parse_aws_argument(std::optional<std::string> aws_arg,
```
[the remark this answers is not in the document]
madsbk: done

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +146,178 @@ class HttpEndpoint : public RemoteEndpoint {
...
+                                        std::string const& env_var,
+                                        std::string const& err_msg,
+                                        bool allow_empty = false)
```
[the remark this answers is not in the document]
madsbk: Good point, using the empty string as `nullopt` to avoid an optional const string.
Done in 84007c1b797919c6a6dd806793541246fcdea47b

madsbk:
Thanks @vyasr, I think I addressed all of your suggestions

python/kvikio/kvikio/_lib/remote_handle.pyx
wence-: Perhaps like this. I am not sure it is much cleaner:
```cython

def _set_handle(self, unique_ptr[cpp_RemoteEndpoint] ep, nbytes):
    if nbytes is None:
        self._handle = make_unique[cpp_RemoteHandle](move(ep))
    else:
        self._handle = make_unique[cpp_RemoteHandle](move(ep), <size_t>nbytes)

def open_http(...):
    cdef RemoteFile ret = RemoteFile()
    cdef unique_ptr[cpp_HttpEndpoint] ep = make_unique[cpp_HttpEndpoint](...)
    ret._set_handle(<unique_ptr[cpp_RemoteEndpoint]>move(ep));
    return ret;
```

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -67,6 +76,59 @@ cdef class RemoteFile:
...
+    def open_s3(
+        cls,
+        bucket_name: str,
+        object_name: str,
+        nbytes: Optional[int],
+    ):
+        cdef RemoteFile ret = RemoteFile()
```
wence-, line 86: *nit* throughout (and in the existing usages);
```suggestion
        cdef RemoteFile ret = cls()
```

Or else you should use a `staticmethod`.

python/kvikio/kvikio/benchmarks/s3_io.py
```diff
@@ -0,0 +1,224 @@
...
+    os.environ["KVIKIO_NTHREADS"] = str(args.nthreads)
+    kvikio.defaults.num_threads_reset(args.nthreads)
+
+    print("Remote S3 benchmark")
+    print("--------------------------------------")
+    print(f"nelem       | {args.nelem} ({format_bytes(args.nbytes)})")
+    print(f"dtype       | {args.dtype}")
+    print(f"nthreads    | {args.nthreads}")
+    print(f"nruns       | {args.nruns}")
+    print(f"server      | {os.getenv('AWS_ENDPOINT_URL', 'http://*.amazonaws.com')}")
```
wence-, line 124: This default seems odd.

wence- (changes requested):
Mostly looks good, and I am bikeshedding on the names. Though I would like it to be harder for the user to use an unencrypted URL.

Some suggestions for cleanup in the cython wrapper.

python/kvikio/kvikio/_lib/remote_handle.pyx
vyasr: Fair enough, whatever you prefer is fine.

python/kvikio/kvikio/_lib/remote_handle.pyx
vyasr: The fundamental problem is that Cython does not natively understand smart pointer polymorphism in the same way that it understands raw pointer polymorphism, so you cannot pass a unique_ptr to a child class where it expects you to pass a unique_ptr to a base class or vice versa. It's a clear sign of the C rather than C++ roots of the project. Here's a patch that builds for me locally and is a bit cleaner IMHO than what Lawrence proposed, with the caveat that you temporarily have a raw pointer that you construct a unique_ptr from rather than using make_unique. Up to you two if you like it or not.
```
diff --git a/python/kvikio/kvikio/_lib/remote_handle.pyx b/python/kvikio/kvikio/_lib/remote_handle.pyx
index 1156300..de48bb9 100644
--- a/python/kvikio/kvikio/_lib/remote_handle.pyx
+++ b/python/kvikio/kvikio/_lib/remote_handle.pyx
@@ -20,13 +20,11 @@ cdef extern from "<kvikio/remote_handle.hpp>" nogil:
     cdef cppclass cpp_RemoteEndpoint "kvikio::RemoteEndpoint":
         pass
 
-    cdef cppclass cpp_HttpEndpoint "kvikio::HttpEndpoint":
+    cdef cppclass cpp_HttpEndpoint "kvikio::HttpEndpoint"(cpp_RemoteEndpoint):
         cpp_HttpEndpoint(string url) except +
 
-    cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint":
+    cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint"(cpp_RemoteEndpoint):
         cpp_S3Endpoint(string url) except +
-
-    cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint":
         cpp_S3Endpoint(string bucket_name, string object_name) except +
 
     pair[string, string] cpp_parse_s3_url \
@@ -56,6 +54,19 @@ cdef string _to_string(str_or_none):
     return str.encode(str(str_or_none))
 
 
+cdef RemoteFile make_remotefile(
+    cpp_RemoteEndpoint* ep,
+    nbytes: Optional[int],
+):
+    cdef RemoteFile ret = RemoteFile()
+    if nbytes is None:
+        ret._handle = make_unique[cpp_RemoteHandle](unique_ptr[cpp_RemoteEndpoint](ep))
+        return ret
+    cdef size_t n = nbytes
+    ret._handle = make_unique[cpp_RemoteHandle](unique_ptr[cpp_RemoteEndpoint](ep), n)
+    return ret
+
+
 cdef class RemoteFile:
     cdef unique_ptr[cpp_RemoteHandle] _handle
 
@@ -65,16 +76,10 @@ cdef class RemoteFile:
         url: str,
         nbytes: Optional[int],
     ):
-        cdef RemoteFile ret = RemoteFile()
-        cdef unique_ptr[cpp_HttpEndpoint] ep = make_unique[cpp_HttpEndpoint](
-            _to_string(url)
+        return make_remotefile(
+            new cpp_HttpEndpoint(_to_string(url)),
+            nbytes,
         )
-        if nbytes is None:
-            ret._handle = make_unique[cpp_RemoteHandle](move(ep))
-            return ret
-        cdef size_t n = nbytes
-        ret._handle = make_unique[cpp_RemoteHandle](move(ep), n)
-        return ret
 
     @classmethod
     def open_s3(
@@ -83,16 +88,10 @@ cdef class RemoteFile:
         object_name: str,
         nbytes: Optional[int],
     ):
-        cdef RemoteFile ret = RemoteFile()
-        cdef unique_ptr[cpp_S3Endpoint] ep = make_unique[cpp_S3Endpoint](
-            _to_string(bucket_name), _to_string(object_name)
+        return make_remotefile(
+            new cpp_S3Endpoint(_to_string(bucket_name), _to_string(object_name)),
+            nbytes,
         )
-        if nbytes is None:
-            ret._handle = make_unique[cpp_RemoteHandle](move(ep))
-            return ret
-        cdef size_t n = nbytes
-        ret._handle = make_unique[cpp_RemoteHandle](move(ep), n)
-        return ret
 
     @classmethod
     def open_s3_from_http_url(
@@ -100,16 +99,10 @@ cdef class RemoteFile:
         url: str,
         nbytes: Optional[int],
     ):
-        cdef RemoteFile ret = RemoteFile()
-        cdef unique_ptr[cpp_S3Endpoint] ep = make_unique[cpp_S3Endpoint](
-            _to_string(url)
+        return make_remotefile(
+            new cpp_S3Endpoint(_to_string(url)),
+            nbytes,
         )
-        if nbytes is None:
-            ret._handle = make_unique[cpp_RemoteHandle](move(ep))
-            return ret
-        cdef size_t n = nbytes
-        ret._handle = make_unique[cpp_RemoteHandle](move(ep), n)
-        return ret
 
     @classmethod
     def open_s3_from_s3_url(
@@ -118,16 +111,10 @@ cdef class RemoteFile:
         nbytes: Optional[int],
     ):
         cdef pair[string, string] bucket_and_object = cpp_parse_s3_url(_to_string(url))
-        cdef RemoteFile ret = RemoteFile()
-        cdef unique_ptr[cpp_S3Endpoint] ep = make_unique[cpp_S3Endpoint](
-            bucket_and_object.first, bucket_and_object.second
+        return make_remotefile(
+            new cpp_S3Endpoint(bucket_and_object.first, bucket_and_object.second),
+            nbytes,
         )
-        if nbytes is None:
-            ret._handle = make_unique[cpp_RemoteHandle](move(ep))
-            return ret
-        cdef size_t n = nbytes
-        ret._handle = make_unique[cpp_RemoteHandle](move(ep), n)
-        return ret
 
     def nbytes(self) -> int:
         return deref(self._handle).nbytes()
```

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -23,6 +23,15 @@ cdef extern from "<kvikio/remote_handle.hpp>" nogil:
     cdef cppclass cpp_HttpEndpoint "kvikio::HttpEndpoint":
         cpp_HttpEndpoint(string url) except +
 
+    cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint":
+        cpp_S3Endpoint(string url) except +
+
+    cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint":
+        cpp_S3Endpoint(string bucket_name, string object_name) except +
```
vyasr, line 30: This is already done in my patch below, but if you don't accept it then this should be done at least.
```suggestion
    cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint":
        cpp_S3Endpoint(string url) except +
        cpp_S3Endpoint(string bucket_name, string object_name) except +
```

I would also recommend reflecting the C++ inheritance scheme into Cython like in my snippet below so that you have access to parent members if needed.

python/kvikio/kvikio/benchmarks/s3_io.py
vyasr: I agree, a follow-up is good. Thanks for opening an issue.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +147,189 @@ class HttpEndpoint : public RemoteEndpoint {
...
+   * @brief Parse a AWS argument such as `aws_region` or `aws_access_key`.
+   *
+   * If not nullopt, the optional's value is returned otherwise the environment
+   * variable `env_var` is used. If that also doesn't have a value:
+   *   - if `err_msg` is empty, the empty string is returned.
+   *   - if `err_msg` is not empty, `std::invalid_argument(`err_msg`)` is thrown.
+   *
+   * @param aws_arg The AWS argument to parse.
+   * @param env_var The name of the environment variable to check if `aws_arg` isn't set.
+   * @param err_msg The error message to throw on error or the empty string.
```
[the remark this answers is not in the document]
vyasr: I had a similar thought about this not being the cleanest pattern, and similarly struggled to come up with something that didn't make every call site more verbose. I could go either way here.

python/kvikio/kvikio/_lib/remote_handle.pyx
jakirkham: Could this just be a templated C++ factory function that we `extern` to Cython and call here?

vyasr: You could do part of it in C++ but RemoteFile is a Python class so you'd still need a wrapper around the C++ function to handle instantiation and assignment to attributes of that object.

jakirkham: Think that is fine. Basically what we would be saying is here is a factory function that takes some arguments and returns a `std::unique_ptr<cpp_RemoteHandle>`, which we just `move` to `ret._handle`

vyasr: Yup that would be fine. Probably just use inline C++ to define that in this file itself.

madsbk: Using a C++ cast function as @jakirkham suggest: 6104106ec4ef725640c32775f1a5a1637ad29faa
It is a bit more verbose than @vyasr's raw pointer approach, but it enforces the pointer uniqueness.

python/kvikio/kvikio/benchmarks/s3_io.py
madsbk: fixed

cpp/include/kvikio/remote_handle.hpp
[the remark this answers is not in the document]
madsbk: renamed to `unwrap_or_default`

python/kvikio/kvikio/_lib/remote_handle.pyx
madsbk: fixed

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +147,189 @@ class HttpEndpoint : public RemoteEndpoint {
...
+   * @param object_name The name of the S3 object.
+   * @param aws_region The AWS region, such as "us-east-1", to use. If nullopt, the value of the
+   * `AWS_DEFAULT_REGION` environment variable is used.
+   * @param aws_access_key The AWS access key to use. If nullopt, the value of the
+   * `AWS_ACCESS_KEY_ID` environment variable is used.
+   * @param aws_secret_access_key The AWS secret access key to use. If nullopt, the value of the
+   * `AWS_SECRET_ACCESS_KEY` environment variable is used.
+   * @param aws_endpoint_url Overwrite the endpoint url to use. If nullopt, the value of
+   * the `AWS_ENDPOINT_URL` environment variable is used. If this is also not set, the regular
+   * AWS url scheme is used: "https://<bucket>.s3.<region>.amazonaws.com/<object>"
```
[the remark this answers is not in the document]
madsbk: done

wence- (approved):
Thanks Mads

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -145,12 +147,189 @@ class HttpEndpoint : public RemoteEndpoint {
...
+    if (std::regex_match(s3_url, matches, pattern)) { return {matches[1].str(), matches[2].str()}; }
+    throw std::invalid_argument("Input string does not match the expected S3 URL format.");
+  }
+
+  /**
+   * @brief Create a S3 endpoint from a url.
+   *
+   * @param url The full http url to the S3 file. NB: this should be an url starting with
+   * "http://" or "https://". If you have an S3 url of the form "s3://<bucket>/<object>",
+   * please use `S3Endpoint::parse_s3_url()` to convert it.
```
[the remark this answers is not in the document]
madsbk: > Can we only use https please and reject http? Or do you want that for testing?

We also wants http for high performance access to public data.

> It looks like yes. I would like this interface to be "safe" by default, and so I would like the user to have to explicitly opt in to using an unencrypted link, given that we send secrets over the wire.

NB: only a time specific signature are send over the wire, curl uses `aws_secret_access_key` to generate the [AWS authentication signature V4](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html). Of cause, the payload is send unencrypted. 

I think it is reasonable to use https by default and accept http if the user overwrite the endpoint url explicitly?

madsbk:
@vyasr do you have anything else ?

vyasr (approved):
Some cosmetic suggestions but generally LGTM now.

PR #484, opened 2024-10-03
Fix no-cuda build

### LINKED ISSUE #482: `nvtx3/nvtx3.hpp` file not found for CPU-only build.
```sh
path/to/kvikio/cpp/include/kvikio/utils.hpp:28:10: fatal error: 'nvtx3/nvtx3.hpp' file not found
   28 | #include <nvtx3/nvtx3.hpp>
      |          ^~~~~~~~~~~~~~~~~
```

madsbk:
Closes #482

### DIFF
--- a/cpp/include/kvikio/utils.hpp
+++ b/cpp/include/kvikio/utils.hpp
@@ -25,7 +25,9 @@
 #include <tuple>
 #include <type_traits>
 
+#ifdef KVIKIO_CUDA_FOUND
 #include <nvtx3/nvtx3.hpp>
+#endif
 
 #include <kvikio/error.hpp>
 #include <kvikio/shim/cuda.hpp>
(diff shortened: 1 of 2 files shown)


cpp/include/kvikio/utils.hpp
```diff
@@ -308,6 +311,11 @@ struct libkvikio_domain {
       msg, nvtx3::payload { convert_to_64bit(val) }           \
     }                                                         \
   }
+#else
+#define KVIKIO_NVTX_FUNC_RANGE_2(msg, val) \
+  do {                                     \
+  } while (0)
+#endif
```
wence-, line 318: Could shovel this into the shim, but because it only appears here that seems like overkill.

wence- (approved):

cpp/include/kvikio/utils.hpp
```diff
@@ -300,6 +302,7 @@ struct libkvikio_domain {
 
 // Macro overloads of KVIKIO_NVTX_FUNC_RANGE
 #define KVIKIO_NVTX_FUNC_RANGE_1() NVTX3_FUNC_RANGE_IN(libkvikio_domain)
+#ifdef KVIKIO_CUDA_FOUND
```
bdice, line 305: Why is this guard only needed for `KVIKIO_NVTX_FUNC_RANGE_2` but not `KVIKIO_NVTX_FUNC_RANGE_1`?

madsbk: Good catch, it worked because `BASIC_NO_CUDA_EXAMPLE` never materialize `KVIKIO_NVTX_FUNC_RANGE_1`. 
I have included all of them in the ifdef guard now: 168dd0c3f91449a16aadbf2cd84c1fcfbedfadc7

bdice (approved):

PR #492, opened 2024-10-07
Visibility of static class methods

madsbk:
We need to export static class methods that returns a reference to a static variable like we do with inline functions: https://github.com/rapidsai/kvikio/issues/442


cpp/include/kvikio/shim/utils.hpp
```diff
@@ -23,6 +23,21 @@
 
 namespace kvikio {
 
+// Macros used for defining symbol visibility, only GLIBC is supported.
```
wence-, line 26: and clang?

madsbk: Right, removed the `only GLIBC is supported` bit

bdice (approved):

PR #493, opened 2024-10-09
Small improvements

kingcrimsontianyu:
This PR makes a couple of small, general improvements.
- Improve the use of `std::tolower` to avoid UB. References [cppreference](https://en.cppreference.com/w/cpp/string/byte/tolower).
- Improve the readability of POSIX read/write interface.
- Improve the use of `std::map` to avoid a repeated search in the self-balancing BST, which takes `O(n)`.

### DIFF
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -59,9 +59,10 @@ inline bool getenv_or(std::string_view env_var_name, bool default_val)
   std::string str{env_val};
   // Special considerations regarding the case conversion:
   // - std::tolower() is not an addressable function. Passing it to std::transform() as
-  //   a function pointer causes the program behavior "unspecified (possibly ill-formed)",
-  //   hence the lambda.
-  // - To avoid UB in std::tolower(), the character must be cast to unsigned char.
+  //   a function pointer, if the compile turns out successful, causes the program behavior
+  //   "unspecified (possibly ill-formed)", hence the lambda. ::tolower() is addressable
+  //   and does not have this problem, but the following item still applies.
+  // - To avoid UB in std::tolower() or ::tolower(), the character must be cast to unsigned char.
   std::transform(
     str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::tolower(c); });
   // Trim whitespaces
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -57,7 +57,11 @@ inline bool getenv_or(std::string_view env_var_name, bool default_val)
   }
   // Convert to lowercase
   std::string str{env_val};
-  // To avoid UB in std::tolower(), the character must be cast to unsigned char
+  // Special considerations regarding the case conversion:
+  // - std::tolower() is not an addressable function. Passing it to std::transform() as
+  //   a function pointer causes the program behavior "unspecified (possibly ill-formed)",
+  //   hence the lambda.
+  // - To avoid UB in std::tolower(), the character must be cast to unsigned char.
   std::transform(
     str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::tolower(c); });
   // Trim whitespaces
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -57,7 +57,9 @@ inline bool getenv_or(std::string_view env_var_name, bool default_val)
   }
   // Convert to lowercase
   std::string str{env_val};
-  std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+  // To avoid UB in std::tolower(), the character must be cast to unsigned char
+  std::transform(
+    str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::tolower(c); });
   // Trim whitespaces
   std::stringstream trimmer;
   trimmer << str;
--- a/cpp/include/kvikio/posix_io.hpp
+++ b/cpp/include/kvikio/posix_io.hpp
@@ -196,8 +196,6 @@ std::size_t posix_device_io(int fd,
   return size;
 }
 
-}  // namespace detail
-
 /**
  * @brief Read from disk to host memory using POSIX
  *
@@ -289,5 +287,7 @@ inline std::size_t posix_device_write(int fd,
   return detail::posix_device_io<IOOperationType::WRITE>(
     fd, devPtr_base, size, file_offset, devPtr_offset);
 }
+
+}  // namespace detail
 
 }  // namespace kvikio
--- a/cpp/include/kvikio/posix_io.hpp
+++ b/cpp/include/kvikio/posix_io.hpp
@@ -28,6 +28,24 @@
 
 namespace kvikio {
 
+/**
+ * @brief Type of the IO operation.
+ *
+ */
+enum class IOType : uint8_t {
+  READ,   ///< POSIX read.
+  WRITE,  ///< POSIX write.
+};
+
+/**
+ * @brief The degree to which the requested size of data is processed.
+ *
+ */
+enum class IODataCompletionLevel : uint8_t {
+  PARTIAL,  ///< POSIX read/write is called only once, which may not process all data requested.
+  FULL,     ///< POSIX read/write is called repeatedly until all requested data are processed.
+};
+
 namespace detail {
 
 /**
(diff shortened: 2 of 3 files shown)


cpp/include/kvikio/posix_io.hpp
```diff
@@ -183,18 +206,20 @@ std::size_t posix_device_io(int fd,
...
+ * @tparam ioDataCompletionLevel Whether all requested data are processed or not. If `FULL`, all of
+ * `count` bytes are read.
  * @param fd File descriptor
  * @param buf Base address of buffer in host memory.
  * @param size Size in bytes to read.
  * @param file_offset Offset in the file to read from.
- * @param partial If false, all of `size` bytes are read.
  * @return Size of bytes that were successfully read.
  */
-inline std::size_t posix_host_read(
```
kingcrimsontianyu, line 193: This change and the one to write are breaking. Is this acceptable? @madsbk

cpp/include/kvikio/defaults.hpp
```diff
@@ -57,7 +57,9 @@ inline bool getenv_or(std::string_view env_var_name, bool default_val)
   }
   // Convert to lowercase
   std::string str{env_val};
-  std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+  // To avoid UB in std::tolower(), the character must be cast to unsigned char
```
mythrocks, line 60: 😮

cpp/include/kvikio/posix_io.hpp
```diff
@@ -84,43 +104,44 @@ class StreamsByThread {
...
  * @param fd File descriptor
  * @param buf Buffer to write
  * @param count Number of bytes to write
  * @param offset File offset
- * @param partial If false, all of `count` bytes are read or written.
  * @return The number of bytes read or written (always gather than zero)
  */
-template <bool IsReadOperation>
-ssize_t posix_host_io(int fd, const void* buf, size_t count, off_t offset, bool partial)
+template <IOType ioType, IODataCompletionLevel ioDataCompletionLevel>
```
wence-, line 116: Is there really a benefit to templating on `IODataCompletionLevel`? It adds a single break to the loop which should be predicted correctly every time by the branch predictor.

cpp/include/kvikio/posix_io.hpp
```diff
@@ -28,6 +28,24 @@
...
+
+/**
+ * @brief The degree to which the requested size of data is completed.
+ *
+ */
+enum class IODataCompletionLevel : uint8_t {
+  PARTIAL,  ///< POSIX read/write is called only once, which may not process all data requested.
+  FULL,     ///< POSIX read/write is called repeatedly until all requested data are processed.
+};
+
```
wence-, line 48: I wonder if, rather than having two enums, it makes more sense to spell out the Cartesian product in the `IOType` enum:

```
enum class IOType : uint8_t {
   READ_PARTIAL,
   READ_FULL,
   WRITE_PARTIAL,
   WRITE_FULL,
};
```

If not, I'm not sure I love `IODataCompletionLevel` as a name. How about:

- `PartialIO::YES/NO`
- `IsPartial::YES/NO`

?

cpp/include/kvikio/posix_io.hpp
```diff
@@ -131,15 +152,15 @@ ssize_t posix_host_io(int fd, const void* buf, size_t count, off_t offset, bool
...
+ * @tparam ioType Whether the operation is a read or a write.
  * @param fd File descriptor
  * @param devPtr_base Device pointer to read or write to.
  * @param size Number of bytes to read or write.
  * @param file_offset Byte offset to the start of the file.
  * @param devPtr_offset Byte offset to the start of the device pointer.
  * @return Number of bytes read or written.
  */
-template <bool IsReadOperation>
+template <IOType ioType>
```
wence-, line 163: *nit*: I think one prefers template parameters to start with an uppercase letter. I think since every usage of the value is matched up with comparison with an `IOType::XXX` then we can get away with not writing `io` in the name as well.
```suggestion
template <IOType Type>
```

wence- (commented):
A few stylistic suggestions.

cpp/include/kvikio/posix_io.hpp
mythrocks: Watching this, to see how breaking changes are best handled.  (One option would have been to deprecate in one release, and then remove in the other.)

cpp/include/kvikio/posix_io.hpp
kingcrimsontianyu: Thanks for the suggestion! The first 4-choice enum looks very succinct. I'll make the changes.

cpp/include/kvikio/posix_io.hpp
mythrocks: > which should be predicted correctly every time by the branch predictor.

@wence-: ELI5 how we know that that would be predicted correctly?  (Asking out of ignorance.)

Would that not also apply to `IsReadOperation` in the prior/original version?

cpp/include/kvikio/posix_io.hpp
kingcrimsontianyu: On a second thought, I probably should keep the two enums and rename the 2nd one. There's some case like [this](https://github.com/rapidsai/kvikio/pull/493/files#diff-ce16f202f49a0591c9c773df9a234a0f2198885e23ffb29d49e1e466f4d9bcedR267) where callers only need to specify read/write, and not partial/full.

cpp/include/kvikio/posix_io.hpp
kingcrimsontianyu: My intention was to make the interface more consistent. Either `IsReadOperation` and  `partial` are both template parameters or both function parameters.

cpp/include/kvikio/posix_io.hpp
```diff
@@ -28,6 +28,24 @@
 
 namespace kvikio {
 
+/**
+ * @brief Type of the IO operation.
+ *
```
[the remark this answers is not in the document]
kingcrimsontianyu: Completed.

cpp/include/kvikio/posix_io.hpp
```diff
@@ -28,6 +28,24 @@
...
+ */
+enum class IOType : uint8_t {
+  READ,   ///< POSIX read.
+  WRITE,  ///< POSIX write.
+};
+
+/**
+ * @brief The degree to which the requested size of data is completed.
+ *
```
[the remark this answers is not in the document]
kingcrimsontianyu: Completed.

cpp/include/kvikio/posix_io.hpp
kingcrimsontianyu: Completed.

cpp/include/kvikio/posix_io.hpp
kingcrimsontianyu: Completed.

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: I've expanded the initial comment, since there's actually another issue on top of UB: `std::tolower` is not an addressable function, and passing it as a function pointer to the STL algorithm functions would cause the program behavior to be "[unspecified (possibly ill-formed)](https://eel.is/c++draft/namespace.constraints#namespace.std-6)". C++ is wild.

mythrocks: > `std::tolower` is not an addressable function, and passing it as a function pointer to the STL algorithm functions would cause the program behavior to be "[unspecified (possibly ill-formed)](https://eel.is/c++draft/namespace.constraints#namespace.std-6)". C++ is wild.

What in the (hello) world...!

cpp/include/kvikio/defaults.hpp
```diff
@@ -57,7 +57,13 @@ inline bool getenv_or(std::string_view env_var_name, bool default_val)
...
   // Convert to lowercase
   std::string str{env_val};
-  std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+  // Special considerations regarding the case conversion:
+  // - std::tolower() is not an addressable function. Passing it to std::transform() as
+  //   a function pointer causes the program behavior "unspecified (possibly ill-formed)",
+  //   hence the lambda.
+  // - To avoid UB in std::tolower(), the character must be cast to unsigned char.
+  std::transform(
+    str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::tolower(c); });
```
madsbk, line 66: wow, good catch!

cpp/include/kvikio/posix_io.hpp
madsbk: I agree, a deprecation cycle is a good thing however, it is unlikely that anyone is calling this function directly. At least, a search on github [doesn't show anything](https://github.com/search?q=posix_host_read&type=code).

In this case, I think we should allow a breaking chance without deprecation. In fact, I think we should put all of the functions in `posix_io.hpp` under `kvikio::detail`, they are not meant to be used directly.

cpp/include/kvikio/posix_io.hpp
```diff
@@ -28,6 +28,22 @@
...
+};
+
+/**
+ * @brief Specifies whether all requested bytes are to be processed or not.
+ */
+enum class PartialIO : uint8_t {
+  YES,  ///< POSIX read/write is called only once, which may not process all bytes requested.
+  NO,   ///< POSIX read/write is called repeatedly until all requested bytes are processed.
+};
+
```
madsbk, line 46: Move this inside `detail` as well?

madsbk: i.e. just use 
```c++
namespace kvikio::detail {
```

kingcrimsontianyu: Good idea! Completed.

madsbk (approved):
Great work @kingcrimsontianyu

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: Is there an advantage of using `::tolower` over `std::tolower`? From the references, they seem functionally the same. I guess perhaps the reason is passing `::tolower` would make the initial code compile, while `std::tolower` doesn't (instead the template version `std::tolower(int, const std::locale&)` is picked, which doesn't compile due to two parameters).

cpp/include/kvikio/defaults.hpp
```diff
@@ -57,7 +57,13 @@ inline bool getenv_or(std::string_view env_var_name, bool default_val)
   }
   // Convert to lowercase
   std::string str{env_val};
-  std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+  // Special considerations regarding the case conversion:
+  // - std::tolower() is not an addressable function. Passing it to std::transform() as
+  //   a function pointer causes the program behavior "unspecified (possibly ill-formed)",
```
vuule, line 62: does this point need to be removed? based on https://github.com/rapidsai/kvikio/pull/493/files#r1795596566

wence-: We are using std::tolower in this PR, so I think this comment is correct.

kingcrimsontianyu: I've expanded the comment a bit more to clarify.

mythrocks (approved):
Good work, this.

PR #498, opened 2024-10-16
Simplify `_to_string` encoding of Python `str`s

jakirkham:
Makes a few improvements to the Cython logic of `_to_string`:

* Types the input as `str`: This allows Cython to use CPython API relevant for the specific type (it will check type on entry)
* Do `is not None` check first: This way when Cython generates additional `None` checks inside the branch, the compiler can optimize them out
* Use `str`'s `encode` method: Cython makes a direct CPython call to generate `bytes` and then extracts the pointer and size for C++ to copy


jakirkham:
Thanks Lawrence! 🙏

PR #500, opened 2024-10-16
Type `open_http`'s `nbytes` as `Py_ssize_t`

jakirkham:
Python typically defines size values as `Py_ssize_t`. Leverage this with `nbytes`. Also as `Py_ssize_t` is `signed` it can take negative values. So pick make the `nbytes` default value `-1`. Update internal checks for `nbytes` being passed similarly. This eliminates the overhead associated with a Python object.


jakirkham:
Superseded with PR: https://github.com/rapidsai/kvikio/pull/501

PR #502, opened 2024-10-17
Mark all of `Array`'s `nogil` `cdef` functions as `noexcept`

jakirkham:
These are all pure C functions implemented in Cython. They do not raise. [Cython 3 adds checks for exceptions]( https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept ) in these functions, which is unnecessary given none of these set exceptions. So add `noexcept` to turn off these Cython checks.


madsbk (approved):
Looks good, thanks @jakirkham

jakirkham:
Thanks Mads! 🙏

PR #503, opened 2024-10-21
Disabling curl tests doesn't now disable kvikio cpp tests

robertmaynard:
Works around https://github.com/curl/curl/issues/15351 . We cache the `BUILD_TESTING` cache variable when it exists so that our tests are enabled when building curl from source.


kingcrimsontianyu (approved):
Lgtm!

bdice (approved):

PR #504, opened 2024-10-22
Use Cython's `array` to back `Py_ssize_t[::1]`

jakirkham:
In `Array`, `Py_ssize_t[::1]` objects are currently backed by [CPython `array`'s]( https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#cpython-array-module ) with some internal bits expressed in Cython. However these are not compatible with [Python's Limited API and Stable ABI]( https://docs.python.org/3/c-api/stable.html#c-api-stability ). To address that, switch to [Cython's own `array` type]( https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#cython-arrays ). As this is baked into Cython and doesn't use anything special, it is compatible with Python's Limited API and Stable ABI.

xref: https://github.com/rapidsai/build-planning/issues/42

### DIFF
--- a/python/kvikio/kvikio/_lib/arr.pyx
+++ b/python/kvikio/kvikio/_lib/arr.pyx
@@ -4,13 +4,13 @@
 # cython: language_level=3
 
 
-from cpython.array cimport array, newarrayobject
 from cpython.buffer cimport PyBuffer_IsContiguous
+from cpython.mem cimport PyMem_Free, PyMem_Malloc
 from cpython.memoryview cimport PyMemoryView_FromObject, PyMemoryView_GET_BUFFER
-from cpython.object cimport PyObject
 from cpython.ref cimport Py_INCREF
 from cpython.tuple cimport PyTuple_New, PyTuple_SET_ITEM
 from cython cimport auto_pickle, boundscheck, initializedcheck, nonecheck, wraparound
+from cython.view cimport array
 from libc.stdint cimport uintptr_t
 from libc.string cimport memcpy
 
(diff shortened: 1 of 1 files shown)


jakirkham:
The performance of the two approaches can be compared with the following Cython test code

<details>

```cython
# filename: new_arr.pyx

from cpython.array cimport array, newarrayobject
from cpython.mem cimport PyMem_Malloc, PyMem_Free
from cpython.object cimport PyObject

from cython.view cimport array as cvarray


cdef array array_Py_ssize_t = array("q")


cpdef Py_ssize_t[::1] new_Py_ssize_t_array_0(Py_ssize_t n):
    return newarrayobject(
        (<PyObject*>array_Py_ssize_t).ob_type, n, array_Py_ssize_t.ob_descr
    )


cdef sizeof_Py_ssize_t = sizeof(Py_ssize_t)


cpdef Py_ssize_t[::1] new_Py_ssize_t_array_1(Py_ssize_t n):
    cdef cvarray a = cvarray((n,), sizeof_Py_ssize_t, b"q", "c", False)
    a.data = <char*>PyMem_Malloc(n * sizeof(Py_ssize_t))
    a.callback_free_data = PyMem_Free
    return a
```

</details>

Running it in IPython shows results like this

```python
In [1]: from new_arr import new_Py_ssize_t_array_0, new_Py_ssize_t_array_1

In [2]: %timeit new_Py_ssize_t_array_0(1)
179 ns ± 0.952 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [3]: %timeit new_Py_ssize_t_array_1(1)
232 ns ± 1.11 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit new_Py_ssize_t_array_0(5)
178 ns ± 0.74 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit new_Py_ssize_t_array_1(5)
235 ns ± 1.95 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [6]: %timeit new_Py_ssize_t_array_0(1_000)
213 ns ± 2.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [7]: %timeit new_Py_ssize_t_array_1(1_000)
284 ns ± 5.85 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

This is ~1/3 slower across cases. That said, the runtime difference is ~60ns. So roughly the time of 1-2 attribute access calls on an object

madsbk (approved):
Looks good, thanks @jakirkham

jakirkham:
Thanks Mads! 🙏

PR #505, opened 2024-10-22
Include <utility> since we use std::exchange

robertmaynard:


jakirkham:
Relevant usage on these lines below

https://github.com/rapidsai/kvikio/blob/43bbca6f8a26bfc9101940979510bfa43fa13a16/cpp/include/kvikio/stream.hpp#L83-L96

Also here's [CPP reference's page one `std::exchange` mentioning it comes from `<utility>`]( https://en.cppreference.com/w/cpp/utility/exchange )

jakirkham (approved):
Thanks Rob! 🙏

Issue #510, opened 2024-10-23
Redefinition of `CUstream` with different type


Jacobfaib:
```
/Users/jfaibussowit/soft/nv/legate.core.internal/arch-cu-driver/cmake_build/_deps/kvikio-src/cpp/include/kvikio/shim/cuda_h_wrapper.hpp:36:7: error: type alias redefinition with different types ('void *' vs 'struct CUstream_st *')
   36 | using CUstream    = void*;
      |       ^
/Users/jfaibussowit/soft/nv/legate.core.internal/src/cpp/legate/cuda/cuda.h:35:7: note: previous definition is here
   35 | using CUstream = struct CUstream_st*;
      |       ^
```
This occurs when:
1. Kvikio is built without CUDA.
2. Kvikio headers are included in a library that also shims various CUDA driver types.

IMO, Kvikio should use the right typedefs for these to match the CUDA driver API. In our case, our forward decls _do_ need to be ABI compatible with CUDA driver, since we `dlsym()` the driver at runtime, but do not want to introduce a build-time dependency on the driver API.

Specifically:
```c++
#if defined(_WIN64) || defined(__LP64__)
// Don't use uint64_t, we want to match the driver headers exactly
using CUdeviceptr = unsigned long long;
#else
using CUdeviceptr = unsigned int;
#endif
static_assert(sizeof(CUdeviceptr) == sizeof(void*));

using CUresult   = int;
using CUdevice   = int;
using CUcontext  = struct CUctx_st*;
using CUstream   = struct CUstream_st*;
using CUevent    = struct CUevent_st*;
using CUfunction = struct CUfunc_st*;
using CUlibrary  = struct CUlib_st*;
using CUkernel   = struct CUkern_st*;
```

PR #511, opened 2024-10-23
Fix CUDA driver type stub definitions

Jacobfaib:
closes https://github.com/rapidsai/kvikio/issues/510


madsbk (approved):
Looks good, thanks @Jacobfaib

Issue #497, opened 2024-10-15
kvikio still segfaults on program termination


EricKern:
Hi everyone,

I'm getting a segfault when my python script terminates. This only happens when kvikio is used.

# Reproducer
`mamba env create -f img2tensor_kvikio.yaml && mamba clean -afy`
```yaml
// img2tensor_kvikio.yaml
name: img2tensor
channels:
  - pytorch
  - nvidia
  - rapidsai
  - conda-forge
dependencies:
  - notebook
  - tifffile
  - python=3.11
  - pytorch
  - pytorch-cuda=12.4
  - kvikio
  ````

bug.py
```python
import kvikio

file_name = 'file0.txt'

fd = kvikio.CuFile(file_name, "w")
fd.close()
```
I'm running in a kubernetes environment. We use the open kernel driver 535.183.01

I assumed this #462  has fixed the issue but it seems there is more to it.

You can find the concretized environment here:
[exported_img2tensor_kvikio.txt](https://github.com/user-attachments/files/17374447/exported_img2tensor_kvikio.txt)

It uses kvikio 24.10 which should include the previously mentioned PR.

jakirkham:
Could you please slim the environment further like so and retry?

```yaml
# filename: kvikio2410_cuda122.yaml
name: kvikio2410_cuda122
channels:
  - rapidsai
  - conda-forge
dependencies:
  - cuda-version=12.2
  - python=3.11
  - kvikio=24.10
```

Asking as there are mismatching CUDA versions in the reproducing environment. Plus some extra bits that appear unused in the example. So would like to simplify further to avoid other potential issues

EricKern:
Unfortunately it still segfaults.
I again attached the concretized dependency list [kvikio2410_cuda122.txt](https://github.com/user-attachments/files/17377315/kvikio2410_cuda122.txt).

The cuda version mismatch seems resolved.
Also the [cufile.log](https://github.com/user-attachments/files/17377369/cufile.log) seems fine to me.
I'm using a MIG slice from an A100 and writing to a weka fs works fine. It only segfaults on program termination

EricKern:
```
(gdb) run
Starting program: /opt/conda/envs/kvikio2410_cuda122/bin/python bug.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff47eb700 (LWP 2675)]
[...]

PR #514, opened 2024-10-24
Python bindings to `cuFileDriverOpen()` and `cuFileDriverClose()`

madsbk:
Changes:

- Adding Python bindings to `cuFileDriverOpen()` and `cuFileDriverClose()`. 
- We now [only open the cufile driver explicitly](https://github.com/rapidsai/kvikio/pull/160) in CUDA versions older than v12.2.
- Introducing `kvikio.cufile_driver.initialize()`, which open the cuFile driver and close it again at module exit.
- Let CI fail if KvikIO wasn't built with cuFile support.
  * Except on cuda11.8+arm64; cuFile didn't support arm until cuda v12.4.
- Some refactor and clean up!


madsbk:
@EricKern calling `kvikio.cufile_driver.initialize()` should fix your segfault. Are you able to build KvikIO in your setup? If so, could you test this PR?

EricKern:
Thank you @madsbk for your efforts!
I'll try to build it next week. Maybe I can even find time tomorrow.

EricKern:
I've built and reran my small segfault reproducer script without explicitly opening and closing the driver.
This still causes the segfault [when I set `profile.cufile_stats` in cufile.json to anything above 0](https://github.com/rapidsai/kvikio/issues/497#issuecomment-2439716688).
Also when I explicitly open and close the driver it still happens.

If `profile.cufile_stats=0` everything works fine.

I guess my segfault (#497) is unrelated to the driver initialization and destruction.

I have tested this on my local machine where I currently don't have a GDS-supported file system. So no actual writing happened.
Only initialization and then cufile's switch to its own compatibility mode.
But even then, the segfault was reproducible on another machine.

madsbk:
Thanks @EricKern, good to get this confirmed. Let's continue the discussion in https://github.com/rapidsai/kvikio/issues/497

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Raises
+    ------
+    RuntimeError
+        If cuFile isn't available.
+    """
+    return cufile_driver.driver_close()
+
+
+def initialize() -> None:
```
wence-, line 45: *question*: Whose job is it to call `initialize`?

madsbk: The user for now. If we find a Python example that segfaults because of cuFile's termination issues, we should consider calling it in `__init__.py`.

wence- (approved):
Tiny suggestions, none are blocking, I think

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -137,6 +140,33 @@ class cuFileAPI {
...
     return _instance;
   }
+
+  /**
+   * @brief Open the cuFile driver
+   *
+   * cuFile accept multiple calls to `cufileDriverOpen()`, only the first call opens
+   * the driver, but every call should have a matching call to `cufileDriverClose()`.
+   */
+  void driver_open()
```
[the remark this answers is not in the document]
madsbk: It is used by `kvikio::DriverInitializer`, which can be used to pay for init overhead up front

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -77,8 +81,8 @@ class cuFileAPI {
     get_symbol(Write, lib, KVIKIO_STRINGIFY(cuFileWrite));
     get_symbol(BufRegister, lib, KVIKIO_STRINGIFY(cuFileBufRegister));
     get_symbol(BufDeregister, lib, KVIKIO_STRINGIFY(cuFileBufDeregister));
-    get_symbol(DriverOpen, lib, KVIKIO_STRINGIFY(cuFileDriverOpen));
-    get_symbol(DriverClose, lib, KVIKIO_STRINGIFY(cuFileDriverClose));
+    get_symbol(_DriverOpen, lib, KVIKIO_STRINGIFY(cuFileDriverOpen));
+    get_symbol(_DriverClose, lib, KVIKIO_STRINGIFY(cuFileDriverClose));
```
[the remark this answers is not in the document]
madsbk: done

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -107,23 +111,22 @@ class cuFileAPI {
 
     // cuFile is supposed to open and close the driver automatically but because of a bug in
     // CUDA 11.8, it sometimes segfault. See <https://github.com/rapidsai/kvikio/issues/159>.
-    CUfileError_t const error = DriverOpen();
-    if (error.err != CU_FILE_SUCCESS) {
-      throw std::runtime_error(std::string{"cuFile error at: "} + __FILE__ + ":" +
-                               KVIKIO_STRINGIFY(__LINE__) + ": " +
-                               cufileop_status_error(error.err));
+    if (!stream_available) {  // The stream API was introduced in CUDA 12.2.
+      driver_open();
```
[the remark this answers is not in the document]
madsbk: Actually, the version of CTK and cuFile might not match, updated the comment:
```
    // cuFile is supposed to open and close the driver automatically but
    // because of a bug in cuFile v1.4 (CUDA v11.8), it sometimes segfault:
    // <https://github.com/rapidsai/kvikio/issues/159>.
    // We use the stream API as an version indicator of cuFile, it was introduced
    // in cuFile v1.7 (CUDA v12.2).
```

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
+
+import atexit
+
+from kvikio._lib import cufile_driver  # type: ignore
+
+# TODO: Wrap nicely, maybe as a dataclass?
+DriverProperties = cufile_driver.DriverProperties
```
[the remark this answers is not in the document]
madsbk: defer to https://github.com/rapidsai/kvikio/issues/526

cpp/include/kvikio/shim/cufile.hpp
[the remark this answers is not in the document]
wence-: Ah, ok.

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -105,25 +109,25 @@ class cuFileAPI {
...
-    if (error.err != CU_FILE_SUCCESS) {
-      throw std::runtime_error(std::string{"cuFile error at: "} + __FILE__ + ":" +
-                               KVIKIO_STRINGIFY(__LINE__) + ": " +
-                               cufileop_status_error(error.err));
-    }
+    // cuFile is supposed to open and close the driver automatically but
+    // because of a bug in cuFile v1.4 (CUDA v11.8), it sometimes segfault:
+    // <https://github.com/rapidsai/kvikio/issues/159>.
+    // We use the stream API as an version indicator of cuFile, it was introduced
+    // in cuFile v1.7 (CUDA v12.2).
```
vyasr, line 116:
```suggestion
    // We use the stream API as a version indicator of cuFile since it was introduced
    // in cuFile v1.7 (CUDA v12.2).
```

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -105,25 +109,25 @@ class cuFileAPI {
...
-    // cuFile is supposed to open and close the driver automatically but because of a bug in
-    // CUDA 11.8, it sometimes segfault. See <https://github.com/rapidsai/kvikio/issues/159>.
-    CUfileError_t const error = DriverOpen();
-    if (error.err != CU_FILE_SUCCESS) {
-      throw std::runtime_error(std::string{"cuFile error at: "} + __FILE__ + ":" +
-                               KVIKIO_STRINGIFY(__LINE__) + ": " +
-                               cufileop_status_error(error.err));
-    }
+    // cuFile is supposed to open and close the driver automatically but
+    // because of a bug in cuFile v1.4 (CUDA v11.8), it sometimes segfault:
```
vyasr, line 113:
```suggestion
    // because of a bug in cuFile v1.4 (CUDA v11.8) it sometimes segfaults:
```

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -105,25 +109,25 @@ class cuFileAPI {
...
+    // We use the stream API as an version indicator of cuFile, it was introduced
+    // in cuFile v1.7 (CUDA v12.2).
+    if (!stream_available) { driver_open(); }
   }
+
+  // Notice, we have to close the driver at program exit (if we opened it) even though we are
+  // not allowed to call CUDA after main[1]. This is because, cuFile will segfault if the
+  // driver isn't closed on program exit i.e. we are doomed if we do, doomed if we don't, but
+  // this seems to be the lesser of two evils.
+  // [1] <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization>
```
vyasr, line 124: Could we explicitly register a `std::atexit` handler? Assuming that the CUDA runtime is doing the same, atexit has ordering guarantees so we'll be pushing our exit handler onto the stack to run before CUDA does its cleanup.

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -0,0 +1,62 @@
...
+
+# TODO: Wrap nicely, maybe as a dataclass?
+# <https://github.com/rapidsai/kvikio/issues/526>
+DriverProperties = cufile_driver.DriverProperties
+
+
+def driver_open() -> None:
+    """Open the cuFile driver
+
+    cuFile accept multiple calls to `driver_open()`, only the first call
```
vyasr, line 16:
```suggestion
    cuFile accepts multiple calls to `driver_open()`. Only the first call
```

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -0,0 +1,62 @@
...
+    RuntimeError
+        If cuFile isn't available.
+    """
+    return cufile_driver.driver_open()
+
+
+def driver_close() -> None:
+    """Close the cuFile driver
+
+    cuFile accept multiple calls to `driver_open()`, only the first call
```
vyasr, line 34:
```suggestion
    cuFile accepts multiple calls to `driver_open()`. Only the first call
```

vyasr (approved):
Mostly cosmetic suggestions. I had one thought about an atexit handler but it seems like that doesn't order the way we need w.r.t. CUDA teardown?

cpp/include/kvikio/shim/cufile.hpp
vyasr: Do we know how CUDA does its teardown, if not via atexit handlers that have predictable ordering?

madsbk: I don't think `std::atexit` has  ordering guarantees between multiple shared libraries :/

madsbk:
Thanks @wence- and @vyasr

PR #516, opened 2024-10-24
Don't use macros for cuda driver functions

Jacobfaib:
Don't use macros for CUDA driver functions, since that does not play nice with `decltype()`. Revert to using type-erased stubs.


madsbk (approved):

madsbk:
Thanks @Jacobfaib

PR #519, opened 2024-10-28
RemoteIO: use a pinned bounce buffer

madsbk:
We use a bounce buffer to avoid many small memory copies to device. Libcurl has a maximum chunk size of 16kb (`CURL_MAX_WRITE_SIZE`) but chunks are often much smaller.

For now, the bounce buffer isn't double buffered, let's defer that to later: https://github.com/rapidsai/kvikio/issues/520


cpp/include/kvikio/remote_handle.hpp
```diff
@@ -450,6 +538,10 @@ class RemoteHandle {
         curl.perform();
       } else {
         PushAndPopContext c(get_context_from_pointer(buf));
+        // We use a bounce buffer to avoid many small memory copies to device. Libcurl has a
+        // maximum chunk size of 16kb (`CURL_MAX_WRITE_SIZE`) but chunks are often much smaller.
+        detail::BounceBufferH2D bounce_buffer(detail::StreamsByThread::get(), buf);
```
vuule, line 543: how big is the bounce buffer?

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -34,6 +36,96 @@
 namespace kvikio {
 namespace detail {
 
+/**
+ * @brief Bounce buffer in pinned host memory.
```
wence-, line 40: *nit*: Maybe note that this object is not thread-safe.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -34,6 +36,96 @@
...
+  {
+    if (size > 0) {
+      CUDA_DRIVER_TRY(cudaAPI::instance().MemcpyHtoDAsync(_dev + _dev_offset, src, size, _stream));
+      CUDA_DRIVER_TRY(cudaAPI::instance().StreamSynchronize(_stream));
+      _dev_offset += size;
+    }
+  }
+
+  /**
+   * @brief FLush the bounce buffer by writing everything to the output device buffer.
```
wence-, line 94:
```suggestion
   * @brief Flush the bounce buffer by writing everything to the output device buffer.
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -34,6 +36,96 @@
...
+   * @brief Write host memory to the bounce buffer (also host memory).
+   *
+   * Only when the bounce buffer has been filled up is data copied to the output device buffer.
+   *
+   * @param data The host memory source.
+   * @param size Number of bytes to write.
+   */
+  void write(char const* data, std::size_t size)
+  {
+    if (_host_buffer.size() - _host_offset < size) {  // Not enough space left on the bounce buffer
```
wence-, line 113:
```suggestion
    if (_host_buffer.size() - _host_offset < size) {  // Not enough space left in the bounce buffer
```

wence- (approved):

PR #522, opened 2024-10-28
CurlHandle: fix error message handling 

madsbk:
When building with conda, the `__FILE__` macro will contain the (long) conda build path. At install time, this is replaced by the actual path, which may be shorter. Any extra characters are replaced by `\0`.

The extra `\0` is problematic if `CurlHandle` later throws an exception to Cython since while converting the exception to Python, Cython might truncate the error message.

### DIFF
--- a/cpp/include/kvikio/shim/libcurl.hpp
+++ b/cpp/include/kvikio/shim/libcurl.hpp
@@ -150,7 +150,7 @@ class CurlHandle {
   /**
    * @brief Construct a new curl handle.
    *
-   * Typically, do not use this directly instead use the `create_curl_handle()` macro.
+   * Typically, do not call this directly instead use the `create_curl_handle()` macro.
    *
    * @param handle An unused curl easy handle pointer, which is retained on destruction.
    * @param source_file Path of source file of the caller (for error messages).
@@ -166,6 +166,7 @@ class CurlHandle {
     setopt(CURLOPT_NOSIGNAL, 1L);
 
     // We always set CURLOPT_ERRORBUFFER to get better error messages.
+    _errbuf[0] = 0;  // Set the error buffer as empty.
     setopt(CURLOPT_ERRORBUFFER, _errbuf);
 
     // Make curl_easy_perform() fail when receiving HTTP code errors.
@@ -216,7 +217,7 @@ class CurlHandle {
     // Perform the curl operation and check for errors.
     CURLcode err = curl_easy_perform(handle());
     if (err != CURLE_OK) {
-      std::string msg(_errbuf);
+      std::string msg(_errbuf);  // We can do this because we always initialize `_errbuf` as empty.
       std::stringstream ss;
       ss << "curl_easy_perform() error near " << _source_file << ":" << _source_line;
       if (msg.empty()) {
--- a/cpp/include/kvikio/shim/libcurl.hpp
+++ b/cpp/include/kvikio/shim/libcurl.hpp
@@ -265,7 +265,7 @@ namespace detail {
  *
  * [1] <https://docs.conda.io/projects/conda-build/en/latest/resources/make-relocatable.html>
  */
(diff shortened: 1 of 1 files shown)


wence-:
Can you add a description of what you observed is the problematic thing this is fixing please?

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -249,12 +250,31 @@ class CurlHandle {
   }
 };
 
+namespace detail {
+/**
+ * @brief Fix Conda's manipulation of __FILE__.
+ *
+ * Conda manipulate the path information in its shared libraries with the results that the C macro
```
wence-, line 257:
```suggestion
 * Conda manipulates the path information in its shared libraries with the results that the C macro
```

Add link to the relevant conda docs? https://docs.conda.io/projects/conda-build/en/latest/resources/make-relocatable.html

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -249,12 +250,31 @@ class CurlHandle {
...
+ * `__FILE__` might contain trailing `\0` chars. Normally, this isn't a problem because `__FILE__`
+ * is a `const char*` that are terminated by the first encounter of `\0`. However, when creating a
+ * `std::string` from a `char*`, the compiler might optimize the code such that the `std::string`
+ * is created from the full size of `__FILE__` including the trailing `\0` chars.
+ */
+__attribute__((optnone)) inline std::string fix_conda_file_path_hack(std::string filename)
```
wence-, line 263: *issue*: I think we should explicitly request `noinline` rather than hoping that `optnone` will ensure the function is not inlined.


```suggestion
__attribute__((noinline)) inline std::string fix_conda_file_path_hack(std::string filename)
```

wence- (changes requested):

cpp/include/kvikio/shim/libcurl.hpp
pentschev: I think we can do both, no?

```
__attribute__((optnone)) __attribute__((noinline)) std::string fix_conda_file_path_hack(std::string filename)
```

pentschev: Interesting, somehow it's not in GCC's but it's in Clang's saying it's available in GNU (I assume GCC): https://clang.llvm.org/docs/AttributeReference.html#optnone

pentschev: Maybe it should be `__attribute__((optimize(0)))` instead?

pentschev (approved):
If https://github.com/rapidsai/kvikio/pull/522/commits/a68b9aa3b209d1f3839e2dcd1771488c75522c5b works then this is probably good as is. However, as [@wence- pointed out `optnone` is not documented in GCC](https://github.com/rapidsai/kvikio/pull/522#discussion_r1820694814) and perhaps might require something different [such as `optimize(0)`](https://github.com/rapidsai/kvikio/pull/522#discussion_r1820694814).

cpp/include/kvikio/shim/libcurl.hpp
madsbk: > Maybe it should be `__attribute__((optimize(0)))` instead?

Does clang support `optimize(0)` ?

pentschev: `optimize` is undocumented in Clang. 🤣 

I guess you'll either have to risk leaving `optnone` or do some compiler checking.

madsbk: Heh, lovely :)

madsbk: Let's see if `__attribute__((noinline))` is enough

wence- (approved):

madsbk:
@wence- @pentschev, thanks for the help debugging this!

Issue #512, opened 2024-10-23
Adjust wheel build tags appropriately for new built component


vyasr:
With the addition of curl to the build, the libkvikio wheels now have a compiled component. They are not being run through auditwheel, though, and I'm not sure auditwheel would even detect this as a valid wheel to audit because libcurl is built as a static library so there are no DSOs in the wheel. The problem is that the libkvikio wheels are now being shipped with `linux` rather than `manylinux` tags, so they won't be as portable as we need them to be. A bigger problem is that the libcurl archive may not be portable anyway if it embeds any noncompliant symbols, but we have no easy way to check that without running auditwheel. I think what we really want is for curl to only be built as part of the C++ interface (i.e. when the Python build or some other consumer like libcudf finds the C++), not as part of the C++ build.

vyasr:
This was resolved in #527 by compiling and shipping a shared library for libkvikio and statically linking libcurl into that while hiding its symbols so that we would stop shipping libcurl.

PR #527, opened 2024-10-30
Build KvikIO as a shared library

madsbk:
Moving libcurl dependent functions to `remote_handle.cpp` (let's move the rest in a follow up PR) and setup building `libkvikio.so`.

### Background 

Now that KvikIO has evolved into a standalone IO library that works without cuFile and CUDA, I think it makes sense to move to a shared library. Originally, KvikIO was a very thin wrapper around cuFile, which is why we decided to keep it header-only (not because of templates). 

**Pros**:
  - Enables us to statically compile libcurl into `libkvikio.so`. 
    * Required by our wheels. If we don’t do this, we will have to find another solution to https://github.com/rapidsai/kvikio/issues/512 
  - Avoid having to compile libcurl in downstream projects. Currently, cudf must compile libcurl in every build. 
  - Reduce compile time in CI, both for KvikIO and downstream projects. 
  - Ease development by not having to rebuild downstream projects like cudf every time KvikIO is modified. 

**Cons**: 
 - Projects cannot vendoring KvikIO as a header-only project. As far as we know, nobody does this.


cpp/include/kvikio/shim/utils.hpp
```diff
@@ -24,11 +24,10 @@
 namespace kvikio {
 
 // Macros used for defining symbol visibility.
-// Since KvikIO is header-only, we rely on the linker to disambiguate inline functions
-// and static methods that have (or return) static references. To do this, the relevant
-// function/method must have `__attribute__((visibility("default")))`. If not, then if
-// KvikIO is used in two different DSOs, the function will appear twice, and there will
-// be two static objects.
+// Since KvikIO declare global default values in headers, we rely on the linker to disambiguate
```
bdice, line 27:
```suggestion
// Since KvikIO declares global default values in headers, we rely on the linker to disambiguate
```

cpp/CMakeLists.txt
```diff
@@ -35,14 +35,23 @@ rapids_cmake_write_version_file(include/kvikio/version_config.hpp)
 # Set a default build type if none was specified
 rapids_cmake_build_type(Release)
 
-# build options
-option(KvikIO_REMOTE_SUPPORT "Configure CMake to build with remote IO support" ON)
+# ##################################################################################################
```
bdice, line 38: These comment blocks are super long and the lengths of the `---` aren't the same length as the `###`. Can we keep this to ~72 chars instead of ~100?

bdice (commented):
This seems mostly right to me. I have a couple pieces of feedback and I noticed CI is not running. It gave an error on the last commit.

```

Invalid workflow file: .github/workflows/pr.yaml#L78
The workflow is not valid. .github/workflows/pr.yaml (Line: 78, Col: 11): Input shared_actions_repo is required, but not provided while calling. .github/workflows/pr.yaml (Line: 78, Col: 11): Input shared_actions_ref is required, but not provided while calling.
```
https://github.com/rapidsai/kvikio/actions/runs/11667562812

bdice:
> I noticed CI is not running. It gave an error on the last commit.

This was due to an unexpected merge of https://github.com/rapidsai/shared-workflows/pull/257, which should be corrected by https://github.com/rapidsai/shared-workflows/pull/258. cc: @msarahan

AyodeAwe:
> This was due to an unexpected merge of [rapidsai/shared-workflows#257](https://github.com/rapidsai/shared-workflows/pull/257), which should be corrected by [rapidsai/shared-workflows#258](https://github.com/rapidsai/shared-workflows/pull/258). cc: @msarahan

@bdice perhaps a typo but [rapidsai/shared-workflows#257](https://github.com/rapidsai/shared-workflows/pull/257) at this time hasn't been merged.

bdice:
> > This was due to an unexpected merge of [rapidsai/shared-workflows#257](https://github.com/rapidsai/shared-workflows/pull/257), which should be corrected by [rapidsai/shared-workflows#258](https://github.com/rapidsai/shared-workflows/pull/258). cc: @msarahan
> 
> @bdice perhaps a typo but [rapidsai/shared-workflows#257](https://github.com/rapidsai/shared-workflows/pull/257) at this time hasn't been merged.

Sorry, I meant https://github.com/rapidsai/shared-workflows/pull/247.

cpp/CMakeLists.txt
madsbk: The 100 width of the comment blocks are forced by the cmake style, which adds `=` until the width is 100. I have fixed the `----` length to match. 
I think we need a style update  in a separate PR.

cpp/CMakeLists.txt
```diff
@@ -55,99 +64,95 @@ if(KvikIO_REMOTE_SUPPORT)
...
-    target_compile_definitions(
-      kvikio INTERFACE $<BUILD_LOCAL_INTERFACE:KVIKIO_CUFILE_STREAM_API_FOUND>
-    )
-  endif()
-endif()
+# To avoid symbol conflicts when statically linking to libcurl.a (see get_libcurl.cmake) and its
+# dependency OpenSSL, we exclude them when building libkvikio.so. This way, libkvikio.so will not
+# expose any OpenSSL symbols that could conflict with downstream users like CPython that also links
+# to (another version of) OpenSSL.
+target_link_options(kvikio PRIVATE "-Wl,--exclude-libs,ALL")
```
robertmaynard, line 117:
```suggestion
target_link_options(kvikio PRIVATE "LINKER:--exclude-libs,ALL")
```
Allow CMake to convert `LINKER:` to `-Wl,` or `-Xlinker` depending on the compiler toolchain used

robertmaynard (changes requested):

madsbk:
@robertmaynard, do you have anything else?

robertmaynard (approved):

vyasr (approved):
Looks good. I checked everything, but not exactly with a fine-toothed comb since we'll be testing this very quickly in the downstream PRs anyway.

madsbk:
@bdice do you have anything?

bdice (approved):
Approving. I previously looked at everything, and now approve based on the diff since my last look.

madsbk:
Thanks all. With this merged, we need to update cudf asap: https://github.com/rapidsai/cudf/pull/17239

PR #537, opened 2024-11-06
Disallow cuda-python 12.6.1 and 11.8.4

bdice:
Due to a bug in cuda-python we must disallow cuda-python 12.6.1 and 11.8.4. This PR disallows those versions.

See https://github.com/rapidsai/build-planning/issues/116 for more information.


jakirkham (approved):
Thanks Bradley! 🙏

PR #540, opened 2024-11-06
Fix exporting of include directories

KyleFromNVIDIA:
The CUDA and cufile include directories should not be exported as absolute paths when installing, only in the build tree. Add a `$<BUILD_INTERFACE>` generator expression.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -125,7 +125,8 @@ add_library(kvikio::kvikio ALIAS kvikio)
 
 target_include_directories(
   kvikio
-  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>" "$<BUILD_INTERFACE:${CUDAToolkit_INCLUDE_DIRS}>"
+  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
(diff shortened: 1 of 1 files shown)


vyasr (approved):
I think that this is a reasonable minimum fix to unblock downstream builds that are currently receiving invalid includes when linking to the kvikio target. I'll open an issue for the longer-term fixes that I'd like to see after this.

vyasr:
OK so now the C++ wheel isn't exporting sufficient information for the Python wheel to know it needs to find CUDA. I thought that we could fix that by addressing point 2 from https://github.com/rapidsai/kvikio/issues/541 and linking to the cuda target directly, but then I remembered that kvikio's cuda support [is optional](https://github.com/rapidsai/kvikio/blob/branch-24.12/cpp/CMakeLists.txt#L45) and we handle cuda [via a dlopen shim](https://github.com/rapidsai/kvikio/blob/branch-24.12/cpp/include/kvikio/shim/cuda.hpp). I think it should still be OK to `target_link_libraries(cuda*` though, right? The linker will prune any libraries that aren't actually used so extra libs shouldn't be a problem, and we'll get the correct includes added.

bdice (approved):

ttnghia:
Hi there! Can you check if this commit causes this error please: https://github.com/NVIDIA/spark-rapids-jni/issues/2580

PR #542, opened 2024-11-06
Replace FindcuFile with upstream FindCUDAToolkit support

KyleFromNVIDIA:
CMake's `FindCUDAToolkit` has supported cuFile since 3.25. Use this support and remove the custom `FindcuFile` module.


cpp/CMakeLists.txt
```diff
@@ -64,41 +64,56 @@ if(KvikIO_REMOTE_SUPPORT)
...
-if(NOT cuFile_FOUND)
+if(KvikIO_CUDA_SUPPORT)
+  if(NOT CUDAToolkit_FOUND)
+    message(FATAL_ERROR "CUDA support requested but CUDA toolkit not found")
+  endif()
+
+  include(cmake/thirdparty/get_nvtx.cmake)
+endif()
+
+if(NOT TARGET CUDA::cuFile)
```
vyasr, line 81: I think this whole block can go inside the `if(KvikIO_CUDA_SUPPORT)` block, right? You never check any of this if you aren't building for CUDA support.

cpp/CMakeLists.txt
```diff
@@ -200,26 +211,40 @@ install(
...
+set(KvikIO_CUFILE_SUPPORT [=[${cuFile_FOUND}]=])
+"
+)
+string(
+  APPEND
+  final_code_string
+  [=[
+if(KvikIO_CUDA_SUPPORT OR KvikIO_CUFILE_SUPPORT)
+  find_package(CUDAToolkit REQUIRED QUIET)
+  target_include_directories(kvikio::kvikio INTERFACE ${CUDAToolkit_INCLUDE_DIRS})
```
vyasr, line 232: Would it be equivalent to add a `$<IF:KvikIO_CUDA_SUPPORT OR KvikIO_CUFILE_SUPPORT,"$<BUILD_INTERFACE:${CUDAToolkit_INCLUDE_DIRS}>",>` to the original `target_include_directories` call?

vyasr (approved):
This looks good. I have a couple of suggestions, but I trust you to apply them or not appropriately. I don't need to review again.

cpp/CMakeLists.txt
KyleFromNVIDIA: Is the CUDA support different from the cuFile support?

bdice: cuFile support requires CUDA support to be enabled.

PR #544, opened 2024-11-07
Remove unused headers

kingcrimsontianyu:
This PR removes unused headers.

### DIFF
--- a/cpp/include/kvikio/error.hpp
+++ b/cpp/include/kvikio/error.hpp
@@ -16,8 +16,6 @@
 #pragma once
 
 #include <cstring>
-#include <exception>
-#include <system_error>
 
 #include <kvikio/shim/cuda.hpp>
 #include <kvikio/shim/cufile_h_wrapper.hpp>
--- a/cpp/include/kvikio/error.hpp
+++ b/cpp/include/kvikio/error.hpp
@@ -16,6 +16,7 @@
 #pragma once
 
 #include <cstring>
+#include <stdexcept>
 
 #include <kvikio/shim/cuda.hpp>
 #include <kvikio/shim/cufile_h_wrapper.hpp>
--- a/cpp/src/remote_handle.cpp
+++ b/cpp/src/remote_handle.cpp
@@ -17,11 +17,6 @@
 #include <cassert>
 #include <cstddef>
 #include <cstring>
-#include <iostream>
-#include <memory>
-#include <sstream>
-#include <stdexcept>
-#include <string>
 
 #include <kvikio/defaults.hpp>
 #include <kvikio/error.hpp>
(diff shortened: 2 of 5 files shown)


cpp/include/kvikio/error.hpp
```diff
@@ -16,8 +16,6 @@
 #pragma once
 
 #include <cstring>
```
bdice, line 18: Do we need `<stdexcept>` for `std::runtime_error`?

cpp/src/remote_handle.cpp
```diff
@@ -17,11 +17,6 @@
 #include <cassert>
 #include <cstddef>
 #include <cstring>
-#include <iostream>
-#include <memory>
```
bdice, line 21: `<memory>` is needed for `std::unique_ptr`.

kingcrimsontianyu: The header file `remote_handle.hpp` has already included `<memory>` for the unique pointers used in the header.

vuule: IIUC, we should include all used headers, regardless of redundancy with what the included headers include themselves.

kingcrimsontianyu: The headers I removed were reported by the `clangd` plug-in that comes with the dev container. The plug-in uses the "[include cleaner](https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html)" to identify the unused include. Given that transitive include is discouraged, that we do not have `misc-include-cleaner` in the `.clang-tidy` file, and that, more importantly, the reported warnings are mysteriously gone now (vscode bug?!), I'll revert the changes.

cpp/include/kvikio/error.hpp
kingcrimsontianyu: Thanks. Done.

kingcrimsontianyu:
Running `clang-tidy` with `misc-include-cleaner` on, it turns out quite a number of places rely on transitive include. Further changes would be necessary to fully follow the best practices.

cpp/src/remote_handle.cpp
bdice: Yes, we aim to avoid relying transitive header inclusions. If an identifier is used in a file, the corresponding header should be included.

bdice: That said, I am happy to accept partial fixes that improve the state of things. For cleanups like this, it is easy to let work drag on. I prefer to accept improvements as-is even if they don’t fix every instance of a problem. Then future diffs can be smaller and we avoid generating conflicts along the way.

madsbk:
@kingcrimsontianyu, please re-open if you are still working on this

PR #478, opened 2024-09-30
Use nvcomp wheel instead of bundling nvcomp

KyleFromNVIDIA:
Contributes to https://github.com/rapdisai/build-planning/issues/74

PR #545, opened 2024-11-07
remove WheelHelpers.cmake

jameslamb:
## Description

Related to https://github.com/rapidsai/build-planning/issues/33 and https://github.com/rapidsai/build-planning/issues/74

The last use of CMake function `install_aliased_imported_targets()` here was removed in #478. This proposes removing the file holding its definition.


KyleFromNVIDIA (approved):

jameslamb:
Thanks @KyleFromNVIDIA !

PR #549, opened 2024-11-14
Pin numcodecs to version that support python 3.10

robertmaynard:


dependencies.yaml
```diff
@@ -356,7 +356,7 @@ dependencies:
           - numpy>=1.23,<3.0a0
           - zarr
           # See https://github.com/zarr-developers/numcodecs/pull/475
-          - numcodecs !=0.12.0
+          - numcodecs>=0.13.0,<14
```
bdice, line 359:
```suggestion
          - numcodecs>=0.13.0,<0.14.0a0
```

bdice (changes requested):
Can we allow newer numcodecs versions on newer Python versions? Why is the upper bound necessary? Are there significant incompatibilities between 0.13 and 0.14?

Also needs an update here: https://github.com/rapidsai/kvikio/blob/d40b7e4541a77842d0af911858b4f0708fe19225/conda/recipes/kvikio/meta.yaml#L74

conda/environments/all_cuda-118_arch-aarch64.yaml
```diff
@@ -21,7 +21,7 @@ dependencies:
 - libcurl>=7.87.0
 - moto>=4.0.8
 - ninja
-- numcodecs !=0.12.0
+- numcodecs>=0.13.0,<14
```
jakirkham, line 24:
```suggestion
- numcodecs>=0.13.0,<0.14.0a0
```

conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -23,7 +23,7 @@ dependencies:
 - libcurl>=7.87.0
 - moto>=4.0.8
 - ninja
-- numcodecs !=0.12.0
+- numcodecs>=0.13.0,<14
```
jakirkham, line 26:
```suggestion
- numcodecs>=0.13.0,<0.14.0a0
```

conda/environments/all_cuda-125_arch-aarch64.yaml
```diff
@@ -22,7 +22,7 @@ dependencies:
 - libcurl>=7.87.0
 - moto>=4.0.8
 - ninja
-- numcodecs !=0.12.0
+- numcodecs>=0.13.0,<14
```
jakirkham, line 25:
```suggestion
- numcodecs>=0.13.0,<0.14.0a0
```

conda/environments/all_cuda-125_arch-x86_64.yaml
```diff
@@ -22,7 +22,7 @@ dependencies:
 - libcurl>=7.87.0
 - moto>=4.0.8
 - ninja
-- numcodecs !=0.12.0
+- numcodecs>=0.13.0,<14
```
jakirkham, line 25:
```suggestion
- numcodecs>=0.13.0,<0.14.0a0
```

conda/recipes/kvikio/meta.yaml
```diff
@@ -71,7 +71,7 @@ requirements:
     - cupy >=12.0.0
     - zarr
     # See https://github.com/zarr-developers/numcodecs/pull/475
-    - numcodecs !=0.12.0
+    - numcodecs>=0.13.0,<14
```
jakirkham, line 74:
```suggestion
- numcodecs>=0.13.0,<0.14.0
```

python/kvikio/pyproject.toml
```diff
@@ -21,7 +21,7 @@ requires-python = ">=3.10"
 dependencies = [
     "cupy-cuda11x>=12.0.0",
     "libkvikio==24.12.*,>=0.0.0a0",
-    "numcodecs !=0.12.0",
+    "numcodecs>=0.13.0,<14",
```
jakirkham, line 24:
```suggestion
    "numcodecs>=0.13.0,<0.14.0a0",
```

jakirkham (commented):
Thanks Rob! 🙏

Added suggestions below for the pinnings. Think we need them a bit tighter

conda/recipes/kvikio/meta.yaml
```diff
@@ -71,7 +71,7 @@ requirements:
     - cupy >=12.0.0
     - zarr
     # See https://github.com/zarr-developers/numcodecs/pull/475
-    - numcodecs !=0.12.0
+- numcodecs>=0.13.0,<0.14.0
```
jakirkham, line 74: Sorry lost some blank spaces in my last review by accident

```suggestion
    - numcodecs >=0.13.0,<0.14.0a0
```

jakirkham:
> Can we allow newer numcodecs versions on newer Python versions? Why is the upper bound necessary? Are there significant incompatibilities between 0.13 and 0.14?

This is a good question. Perhaps first it is helpful to provide some context around the existing pin?

We added this pin due to a bug in Numcodecs' entrypoint handling logic that occurred in `0.12.0` and was fixed in `0.12.1` with Vyas' PR ( https://github.com/zarr-developers/numcodecs/pull/475 ). So we just excluded the problematic version here. AIUI Rob is keeping with that and simply bumping the minimum version past the problematic version giving us `0.13`

As to newer versions of Numcodecs, periodically it does add new optional codecs. Sometimes there are small fixes or improvements to existing ones. The API has been pretty stable for a while

Looking at [the diff between 0.12.1 and 0.14]( https://github.com/zarr-developers/numcodecs/compare/v0.12.1...v0.14.0 ), it appears Python 3.10 was dropped (as discussed above), some dependencies were bumped or new versions supported (like NumPy 2), Zarr v3 (unreleased) support was added, a new cyclic redundancy check (CRC) with optional dependency was added, existing CRCs got a new optional flag, a performance fix to the `Delta` codec was made, a `multiprocessing` bug fix, a breaking change with `Zstd` attributes (don't think we use these), etc. In terms of API breaks am not seeing much to worry about here. Think that will likely be true in the future (though we should keep an eye on how Zarr v3 support in Numcodecs evolves)

<hr>

Think the real question is, are we ok with having some drift between what our oldest supported Python version (currently 3.10) and what our newest supported Python version (currently 3.12) can be installed with?

And if so, are there any changes we may need to make elsewhere to adapt to this reality? If not, should we start constraining all our dependencies based on the oldest Python version we support (and when do we refresh them)?

Admittedly we might want to have this discussion in somewhere more general than this PR for greater visibility and easier referencing. Though wanted to follow up here since this is where the questions came up

bdice:
> Think the real question is, are we ok with having some drift between what our oldest supported Python version (currently 3.10) and what our newest supported Python version (currently 3.12) can be installed with?


Yes, we are usually happy with allowing drift here. If we pin all our dependencies to only those version ranges that work with all our supported Python versions, we would be too tightly pinned for broad compatibility with the Python ecosystem. If the numcodecs APIs we use are compatible with 0.13 and 0.14, we should not exclude 0.14.

robertmaynard:
I have expanded the comment in the `dependencies.yaml` with more context on the pinning.

KyleFromNVIDIA (approved):

dependencies.yaml
```diff
@@ -356,7 +356,7 @@ dependencies:
           - numpy>=1.23,<3.0a0
           - zarr
           # See https://github.com/zarr-developers/numcodecs/pull/475
```
bdice, line 358: Is the PR linked above still relevant? I think it may be safe to remove this comment now.

conda/recipes/kvikio/meta.yaml
```diff
@@ -71,7 +71,7 @@ requirements:
     - cupy >=12.0.0
     - zarr
     # See https://github.com/zarr-developers/numcodecs/pull/475
```
bdice, line 73: (Same as dependencies.yaml) Is the PR linked above still relevant? I think it may be safe to remove this comment now.

robertmaynard: I presume it is still relevant as it explains why we have the minimum value clamp as well. If kvikio is combined with another package that requests 12.X we need to report as unsolvable.

robertmaynard:
> > I have expanded the comment in the `dependencies.yaml` with more context on the pinning.
> 
> @robertmaynard Did you push? I don't see the context / comments. Blocking merge until we can figure out why this needs the upper bound.

Somehow didn't push. Updated now

dependencies.yaml
```diff
@@ -355,8 +355,10 @@ dependencies:
         packages:
           - numpy>=1.23,<3.0a0
           - zarr
-          # See https://github.com/zarr-developers/numcodecs/pull/475
-          - numcodecs !=0.12.0
+          # Pinned to the 0.13 for the following reasons:
+          #   https://github.com/zarr-developers/numcodecs/pull/475
+          #   0.14.0 doesn't offer python 3.10 support
+          - numcodecs>=0.13.0,<0.14.0a0
```
bdice, line 361: Not every version in this range needs to support every Python version that we build for. We only need _one_ compatible version of `numcodecs` in this range for each Python version we support.

Can we relax this?
```suggestion
          - numcodecs>=0.13.0
```

bdice (changes requested):

robertmaynard:
> Can we allow newer numcodecs versions on newer Python versions? Why is the upper bound necessary? Are there significant incompatibilities between 0.13 and 0.14?
> 
> Also needs an update here:
> 
> https://github.com/rapidsai/kvikio/blob/d40b7e4541a77842d0af911858b4f0708fe19225/conda/recipes/kvikio/meta.yaml#L74

Updated to allow all version >= 0.13.0

bdice (changes requested):
Why is this PR needed at all? The existing pinnings should be resolvable with Python 3.10, which is the PR title's stated goal. Do we need `>=0.13` for some feature used in the kvikio codebase? If so, what? numcodecs version 0.12.1 should still suffice for Python 3.10 users, as far as I can tell. Also, the PR still lacks a description.

From above,
> AIUI Rob is keeping with that and simply bumping the minimum version past the problematic version giving us 0.13

I would like a specific reason why 0.12.1 is not viable before bumping the lower bound.

PR #550, opened 2024-11-14
Backport `FindCUDAToolkit` from CMake 3.31

KyleFromNVIDIA:
Backport to get
https://gitlab.kitware.com/cmake/cmake/-/commit/b38a8e77cb3c8401b3022a68f07a4fd77b290524, so that `find_library()` finds `libcufile` correctly.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,6 +62,10 @@ repos:
           - cmakelang==0.6.13
         verbose: true
         require_serial: true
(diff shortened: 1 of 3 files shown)


bdice (approved):

vyasr (changes requested):
Do we need to copy in a license file before we can vendor CMake code directly like this?

vyasr (approved):

PR #551, opened 2024-11-14
prefer wheel-provided libkvikio.so, use RTLD_LOCAL

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/118

Modifies `libkvikio.load_library()` in the following ways:

* prefer wheel-provided `libkvikio.so` to system installation
* expose environment variable `RAPIDS_LIBKVIKIO_PREFER_SYSTEM_LIBRARY` for switching that preference
* load `libkvikio.so` with `RTLD_LOCAL`, to prevent adding symbols to the global namespace ([dlopen docs](https://linux.die.net/man/3/dlopen))

### How I tested this

Tested the general approach on https://github.com/rapidsai/cudf/pull/17316 and https://github.com/rapidsai/cuspatial/pull/1483.

### DIFF
--- a/python/libkvikio/libkvikio/load.py
+++ b/python/libkvikio/libkvikio/load.py
@@ -69,5 +69,5 @@ def load_library():
 
     # The caller almost never needs to do anything with this library, but no
     # harm in offering the option since this object at least provides a handle
-    # to inspect where the libkvikio was loaded from.
+    # to inspect where libkvikio was loaded from.
     return libkvikio_lib
--- a/python/libkvikio/libkvikio/load.py
+++ b/python/libkvikio/libkvikio/load.py
@@ -16,30 +16,58 @@
 import ctypes
 import os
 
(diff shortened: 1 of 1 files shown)


python/libkvikio/libkvikio/load.py
```diff
@@ -16,30 +16,58 @@
...
+        # (which might be the case in builds where the library was prebuilt
+        # before packaging the wheel), look for a system installation.
+        libkvikio_lib = _load_wheel_installation(soname)
+        if libkvikio_lib is None:
+            libkvikio_lib = _load_system_installation(soname)
 
     # The caller almost never needs to do anything with this library, but no
     # harm in offering the option since this object at least provides a handle
-    # to inspect where libkvikio was loaded from.
+    # to inspect where the libkvikio was loaded from.
```
jameslamb, line 72:
```suggestion
    # to inspect where libkvikio was loaded from.
```

PR #553, opened 2024-11-15
fix library-loading issues in editable installs

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/118

The pattern introduced in #551 breaks editable installs in devcontainers. In that type of build, `libkvikio.so` is built outside of the wheel but **not installed**, so it can't be found by `ld`. Extension modules in `kvikio` are able to find it via RPATHs instead.

This proposes:

* try-catching the entire library-loading attempt, to silently do nothing in cases like that
* adding an import of the `kvikio` Python library in the `devcontainers` CI job, as a smoke test to catch issues like this in the future

### How I tested this

Reproduced that with the CUDA 12.5 pip devcontainers today:

```shell
build-all
python -c "import kvikio"
# OSError: libkvikio.so: cannot open shared object file: No such file or directory
```

Confirmed that the changes in this PR fix that.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -104,6 +104,7 @@ jobs:
       build_command: |
         sccache -z;
         build-all --verbose;
+        python -c "import kvikio; print(kvikio.__version__)";
         sccache -s;
   wheel-cpp-build:
     secrets: inherit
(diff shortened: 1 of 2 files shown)


vyasr (approved):
About as good as we can do for now. Let's revisit in 25.02.

wence- (approved):
Thanks James

PR #559, opened 2024-11-20
Fix the pytest error for async io

### LINKED ISSUE #558: Python test_async_io fails on GDS-enabled system
```
# <my_gds_supported_path> below represents a GDS supported file system.
# Note that in the dev container, the temp directory must not be the default /tmp.
# The default /tmp is an OverlayFS, which would cause cuFile I/O to have "internal errors" at runtime
export TMPDIR=<my_gds_supported_path>
pytest ~/kvikio/python/kvikio/tests/test_async_io.py -x
```

On a GDS-enabled system, the above test yields the following error:

```
python/kvikio/tests/test_async_io.py:36: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   RuntimeError: cuFile error at: /home/coder/kvikio/cpp/include/kvikio/stream.hpp:136: unsupported file open flags

kvikio/_lib/future.pyx:18: RuntimeError

During handling of the above exception, another exception occurred:

tmp_path = PosixPath('/mnt/profile/pytest-of-coder/pytest-37/test_read_write_1_0'), size = 1

    @pytest.mark.parametrize("size", [1, 10, 100, 1000, 1024, 4096, 4096 * 10])
    def test_read_write(tmp_path, size):
        """Test basic read/write"""
        filename = tmp_path / "test-file"
    
        stream = cupy.cuda.Stream()
    
        # Write file
        a = cupy.arange(size)
        f = kvikio.CuFile(filename, "w")
        assert not f.closed
        assert check_bit_flags(f.open_flags(), os.O_WRONLY)
        assert f.raw_write_async(a, stream.ptr).check_bytes_done() == a.nbytes
    
        # Try to read file opened in write-only mode
>       with pytest.raises(RuntimeError, match="Operation not permitted"):
E       AssertionError: Regex pattern did not match.
E        Regex: 'Operation not permitted'
E        Input: 'cuFile error at: /home/coder/kvikio/cpp/include/kvikio/stream.hpp:136: unsupported file open flags'

python/kvikio/tests/test_async_io.py:33: AssertionError
======================================================== short test summary info =========================================================
FAILED python/kvikio/tests/test_async_io.py::test_read_write[1] - AssertionError: Regex pattern did not match.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=========================================================== 1 failed in 2.82s ============================================================
cuFile error at: /home/coder/kvikio/cpp/include/kvikio/stream.hpp:136: unsupported file open flags
```

kingcrimsontianyu:
This PR fixes #558 

The root cause is that POSIX and cuFile report different error messages for the same invalid operation which attempts to read from a write-only file. The unit test needs to take account of these different messages.

### DIFF
--- a/python/kvikio/tests/test_async_io.py
+++ b/python/kvikio/tests/test_async_io.py
@@ -32,7 +32,9 @@ def test_read_write(tmp_path, size):
     # Try to read file opened in write-only mode
     # POSIX read would yield the error "Operation not permitted"
     # cuFile read would yield the error "unsupported file open flags"
-    with pytest.raises(RuntimeError, match="Operation not permitted|unsupported file open flags"):
+    with pytest.raises(
+        RuntimeError, match="Operation not permitted|unsupported file open flags"
+    ):
         # The exception is raised when we call the raw_read_async API.
         future_stream = f.raw_read_async(a, stream.ptr)
         future_stream.check_bytes_done()
--- a/python/kvikio/tests/test_async_io.py
+++ b/python/kvikio/tests/test_async_io.py
@@ -30,7 +30,9 @@ def test_read_write(tmp_path, size):
     assert f.raw_write_async(a, stream.ptr).check_bytes_done() == a.nbytes
 
     # Try to read file opened in write-only mode
-    with pytest.raises(RuntimeError, match="Operation not permitted"):
+    # POSIX read would yield the error "Operation not permitted"
+    # cuFile read would yield the error "unsupported file open flags"
+    with pytest.raises(RuntimeError, match="Operation not permitted|unsupported file open flags"):
         # The exception is raised when we call the raw_read_async API.
         future_stream = f.raw_read_async(a, stream.ptr)
         future_stream.check_bytes_done()


python/kvikio/tests/test_async_io.py
```diff
@@ -29,8 +29,13 @@ def test_read_write(tmp_path, size):
...
     assert f.raw_write_async(a, stream.ptr).check_bytes_done() == a.nbytes
 
+    if f.is_compat_mode_preferred_for_async():
+        expected_except_msg = "Operation not permitted"
+    else:
+        expected_except_msg = "unsupported file open flags"
+
     # Try to read file opened in write-only mode
-    with pytest.raises(RuntimeError, match="Operation not permitted"):
+    with pytest.raises(RuntimeError, match=expected_except_msg):
```
vuule, line 38: not a big fan of tests that check the error message content, but that's not on this PR.

vuule (approved):

python/kvikio/tests/test_async_io.py
bdice: It see ms like we're adding a lot of public APIs just for this error. Can we instead just verify that the error matches _one_ of these strings, without worrying about which one? That would let us remove all the `is_compat_mode_preferred` logic.

kingcrimsontianyu: That makes sense. I'm wondering if ideally the public function `is_compat_mode_preferred()` and `is_compat_mode_preferred_for_async()` should be included in the Python API, or are they considered too low-level?

kingcrimsontianyu: The Python API addition has been reverted for now.

vuule (approved):

madsbk (approved):
Looks good, thanks @kingcrimsontianyu

PR #565, opened 2024-12-04
cufile version 

madsbk:
Fixes https://github.com/rapidsai/kvikio/issues/566

To clean up and make cufile capability detection more robust,  we now use `cuFileGetVersion()`. 
Should fix the nightly failure: https://github.com/rapidsai/kvikio/actions/runs/12133477083/job/33877440755

NB:  `cuFileGetVersion()` first became available in cufile v1.8 (CTK v12.3) thus the stream and batch API detection will return false for  versions older than v1.8. I think this is acceptable for robustness.

### DIFF
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -62,23 +62,20 @@ class FileHandle {
    */
   bool is_compat_mode_preferred_for_async(CompatMode requested_compat_mode)
   {
-    if (!defaults::is_compat_mode_preferred(requested_compat_mode)) {
-      if (!is_batch_and_stream_available()) {
-        if (requested_compat_mode == CompatMode::AUTO) { return true; }
-        throw std::runtime_error("Missing cuFile batch or stream library symbol.");
-      }
+    if (defaults::is_compat_mode_preferred(requested_compat_mode)) { return true; }
 
-      // When checking for availability, we also check if cuFile's config file exist. This is
-      // because even when the stream API is available, it doesn't work if no config file exist.
-      if (config_path().empty()) {
-        if (requested_compat_mode == CompatMode::AUTO) { return true; }
-        throw std::runtime_error("Missing cuFile configuration file.");
-      }
-
-      return false;
+    if (!is_stream_api_available()) {
+      if (requested_compat_mode == CompatMode::AUTO) { return true; }
+      throw std::runtime_error("Missing the cuFile stream api.");
     }
 
-    return true;
+    // When checking for availability, we also check if cuFile's config file exist. This is
+    // because even when the stream API is available, it doesn't work if no config file exist.
+    if (config_path().empty()) {
+      if (requested_compat_mode == CompatMode::AUTO) { return true; }
+      throw std::runtime_error("Missing cuFile configuration file.");
+    }
+    return false;
   }
 
  public:
@@ -670,7 +667,7 @@ class FileHandle {
    */
   [[nodiscard]] bool is_compat_mode_preferred_for_async() const noexcept
   {
-    static bool is_extra_symbol_available = is_batch_and_stream_available();
+    static bool is_extra_symbol_available = is_stream_api_available();
     static bool is_config_path_empty      = config_path().empty();
     return is_compat_mode_preferred() || !is_extra_symbol_available || is_config_path_empty;
   }
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -69,8 +69,8 @@ class FileHandle {
       throw std::runtime_error("Missing the cuFile stream api.");
     }
 
-    // When checking for availability, we also check if cuFile's config file exist. This is
-    // because even when the stream API is available, it doesn't work if no config file exist.
+    // When checking for availability, we also check if cuFile's config file exists. This is
+    // because even when the stream API is available, it doesn't work if no config file exists.
     if (config_path().empty()) {
       if (requested_compat_mode == CompatMode::AUTO) { return true; }
       throw std::runtime_error("Missing cuFile configuration file.");
--- a/cpp/include/kvikio/shim/cufile.hpp
+++ b/cpp/include/kvikio/shim/cufile.hpp
@@ -242,7 +242,7 @@ constexpr bool cufile_version() { return 0; }
  *
  * @return The boolean answer
  */
-inline bool is_batch_api_available() noexcept { return cufile_version() > 1060; }
+inline bool is_batch_api_available() noexcept { return cufile_version() >= 1060; }
 
 /**
  * @brief Check if cuFile's stream (async) API is available.
@@ -253,6 +253,6 @@ inline bool is_batch_api_available() noexcept { return cufile_version() > 1060; 
  *
  * @return The boolean answer
  */
-inline bool is_stream_api_available() noexcept { return cufile_version() > 1070; }
+inline bool is_stream_api_available() noexcept { return cufile_version() >= 1070; }
 
 }  // namespace kvikio
--- a/cpp/include/kvikio/shim/cufile.hpp
+++ b/cpp/include/kvikio/shim/cufile.hpp
@@ -212,13 +212,13 @@ inline bool is_cufile_available()
 /**
  * @brief Get cufile version (or zero if older than v1.8).
  *
- * The version is returned as (1000 major + 10 minor). E.g., cufile v1.8.0 would
+ * The version is returned as (1000*major + 10*minor). E.g., cufile v1.8.0 would
  * be represented by 1080.
  *
  * Notice, this is not the version of the CUDA toolkit. cufile is part of the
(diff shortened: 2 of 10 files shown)


python/kvikio/tests/test_cufile_driver.py
```diff
@@ -6,6 +6,10 @@
 import kvikio.cufile_driver
 
 
+def test_version():
+    print(kvikio.cufile_driver.libcufile_version())
```
bdice, line 10: Please assert some property here, like that it is an tuple of two elements with each value >=0. Avoid printing in tests.

bdice (commented):
I think I understand the goals here, but can you make the PR description clearer? Something like “we observed a failure on CTK 12.0 because …”. Or file an issue for this PR.

python/kvikio/tests/test_cufile_driver.py
madsbk: done

madsbk:
> I think I understand the goals here, but can you make the PR description clearer? Something like “we observed a failure on CTK 12.0 because …”. Or file an issue for this PR.

Added https://github.com/rapidsai/kvikio/issues/566

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -88,33 +91,35 @@ class cuFileAPI {
...
+      get_symbol(BatchIODestroy, lib, KVIKIO_STRINGIFY(cuFileBatchIODestroy));
+    }
+
+    if (version >= 1070) {
+      get_symbol(ReadAsync, lib, KVIKIO_STRINGIFY(cuFileReadAsync));
+      get_symbol(WriteAsync, lib, KVIKIO_STRINGIFY(cuFileWriteAsync));
+      get_symbol(StreamRegister, lib, KVIKIO_STRINGIFY(cuFileStreamRegister));
+      get_symbol(StreamDeregister, lib, KVIKIO_STRINGIFY(cuFileStreamDeregister));
+    }
+
```
wence-, line 118: So these symbols existed in, respectively 1.6 and 1.7, but we won't enable them unless we're using v1.8 (because `cuFileGetVersion` didn't exist, and so `version` is zero)?

cpp/include/kvikio/shim/cufile_h_wrapper.hpp
```diff
@@ -75,7 +75,7 @@ CUfileError_t cuFileDriverSetMaxPinnedMemSize(...);
 
 #endif
 
-// If the Batch API isn't defined, we define some of the data types here.
+// If some cufile APIs isn't defined, we define some of the data types here.
```
wence-, line 78:
```suggestion
// If some cufile APIs aren't defined, we define some of the data types here.
```

wence- (approved):
Tiny typo suggestion

cpp/include/kvikio/shim/cufile.hpp
madsbk: Yes, added:
```cpp
    // Some symbols were introduced in later versions, so version guards are required.
    // Note: `version` is 0 for cuFile versions prior to v1.8 because `cuFileGetVersion`
    // did not exist. As a result, the batch and stream APIs are not loaded in versions
    // 1.6 and 1.7, respectively, even though they are available. This trade-off is made
    // for improved robustness.
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -62,23 +62,20 @@ class FileHandle {
...
-      return false;
+    if (defaults::is_compat_mode_preferred(requested_compat_mode)) { return true; }
+
+    if (!is_stream_api_available()) {
+      if (requested_compat_mode == CompatMode::AUTO) { return true; }
+      throw std::runtime_error("Missing the cuFile stream api.");
     }
 
-    return true;
+    // When checking for availability, we also check if cuFile's config file exist. This is
```
vyasr, line 72:
```suggestion
    // When checking for availability, we also check if cuFile's config file exists. This is
```

cpp/include/kvikio/file_handle.hpp
```diff
@@ -62,23 +62,20 @@ class FileHandle {
...
+    // because even when the stream API is available, it doesn't work if no config file exist.
```
vyasr, line 73:
```suggestion
    // because even when the stream API is available, it doesn't work if no config file exists.
```

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -10,6 +11,24 @@
 DriverProperties = cufile_driver.DriverProperties
 
 
+def libcufile_version() -> Tuple[int, int]:
+    """Get the libcufile version (or zero if older than v1.8).
```
vyasr, line 15: Minor note that this returns (0, 0) and not 0 for old versions. I would suggest removing the parenthetical here and instead specifying that behavior in the Notes below.

vyasr (approved):
Some minor suggestions but otherwise LGTM.

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -205,25 +214,49 @@ inline bool is_cufile_available()
...
-  } catch (const std::runtime_error&) {
-    return false;
+    return cuFileAPI::instance().version;
+  } catch (std::runtime_error const&) {
+    return 0;
   }
 }
 #else
-constexpr bool is_batch_and_stream_available() { return false; }
+constexpr bool cufile_version() { return 0; }
```
bdice, line 237: The return type should match the above.
```suggestion
constexpr int cufile_version() { return 0; }
```

cpp/include/kvikio/shim/cufile_h_wrapper.hpp
```diff
@@ -75,7 +75,7 @@ CUfileError_t cuFileDriverSetMaxPinnedMemSize(...);
 
 #endif
```
bdice, line 76: It seems like this would be safer if we nested this macro so that we got
```
#ifdef KVIKIO_CUFILE_FOUND
...
  #ifndef KVIKIO_CUFILE_BATCH_API_FOUND
  ...
  #endif
#endif
```

Currently it seems like we are defining these other batch/stream types even if `KVIKIO_CUFILE_FOUND` is false? Or I'm misreading the macros... basically I am proposing that we move the `#endif` on line 76 to the bottom of the file.

python/kvikio/kvikio/benchmarks/utils.py
```diff
@@ -41,6 +42,10 @@ def pprint_sys_info() -> None:
         gpu_name = f"{pynvml.nvmlDeviceGetName(dev)} (dev #0)"
         mem_total = format_bytes(pynvml.nvmlDeviceGetMemoryInfo(dev).total)
         bar1_total = format_bytes(pynvml.nvmlDeviceGetBAR1MemoryInfo(dev).bar1Total)
+    if version == (0, 0):
+        libcufile_version = "unknown"
```
bdice, line 46: Maybe we can add this info here or perhaps just in a code comment.
```suggestion
        libcufile_version = "unknown (earlier than cuFile 1.8)"
```

python/kvikio/kvikio/benchmarks/utils.py
```diff
@@ -41,6 +42,10 @@ def pprint_sys_info() -> None:
...
+    else:
+        libcufile_version = f"{version[0]}.{version[1]}"
```
bdice, line 48:
```suggestion
        libcufile_version = ".".join(version)
```

python/kvikio/kvikio/cufile_driver.py
madsbk: Changed to:
```python
def libcufile_version() -> Tuple[int, int]:
    """Get the libcufile version.

    Returns (0, 0) for cuFile versions prior to v1.8.

    Notes
    -----
    This is not the version of the CUDA toolkit. cufile is part of the
    toolkit but follows its own version scheme.

    Returns
    -------
    The version as a tuple (MAJOR, MINOR).
    """
```

python/kvikio/kvikio/benchmarks/utils.py
madsbk: Reverted back to `f"{version[0]}.{version[1]}"`, mypy complains because `version` is integers and not strings.

bdice (approved):

madsbk:
Thanks everyone

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -10,6 +11,26 @@
...
+    toolkit but follows its own version scheme.
+
+    Returns
+    -------
+    The version as a tuple (MAJOR, MINOR).
+    """
+    v = cufile_driver.libcufile_version()
+    major = v // 1000
+    minor = (v % 1000) // 10
+    return (major, minor)
```
jakirkham, line 31: Maybe we can use `divmod` to consolidate these a bit?

```suggestion
    major, minor = divmod(v, 1000)
    minor //= 10
    return (major, minor)
```

Issue #556, opened 2024-11-15
Add C++ and Python mock tests for compatibility mode-dependent code paths


kingcrimsontianyu:
Some key components in KvikIO such as the methods of `FileHandle` or `BatchHandle` branch on different compatibility modes (`ON`, `OFF`, `AUTO`). Commonly, the CI and local environment do not satisfy the system requirement for GDS I/O, leaving the code on the `OFF` path (or `AUTO` but inferred at runtime to be `OFF`) untested.

We may improve our testing methodology (and code coverage) by introducing mock tests for the C++ and Python code.

PR #567, opened 2024-12-04
Improve compatibility mode consistency

kingcrimsontianyu:
This PR adjusts the behavior of the file handle's compatibility mode to make it more consistent with its naming, i.e. `AUTO` actively falls back to the compatibility mode when any GDS step goes wrong; `OFF` sticks to the GDS path and errors out when things go wrong.


vuule (approved):

madsbk (commented):
It would be good with a test. Is it possible to make a file path in testing that cannot be opened with the `O_DIRECT` flag?

kingcrimsontianyu:
@madsbk Hmm. I'm not sure how to do that. OpenBSD does not support direct I/O, so open any file with this flag would fail, but I'm not sure how to cause the failure in a more generic, OS-agnostic way.
Imo,  mock test (https://github.com/rapidsai/kvikio/issues/556) is probably best suited for the compatibility mode-related test.

madsbk (approved):
Sounds good

kingcrimsontianyu:
\merge

PR #568, opened 2024-12-05
Upgrade nvcomp to 4.1.1.1

bdice:
This updates kvikio to use nvcomp 4.1.1.1. This includes some bug fixes needed for Spark-RAPIDS.

The version is updated in rapids-cmake in https://github.com/rapidsai/rapids-cmake/pull/727.


jakirkham:
Thanks Bradley! 🙏

Are we able to use a looser pinning for this dependency? Think that would make it easier to update (and for users to install)

bdice:
> Are we able to use a looser pinning for this dependency?

No, unfortunately. We require this to match rapids-cmake's pinnings exactly, and to ensure that we get bug fixes in the latest release. Eventually it would be nice to expand our range of nvcomp versions we allow, but we have been pinned at "latest" for a long time due to the requirement of matching rapids-cmake, in addition to many breaking changes and required bug fixes in each new release of nvcomp. Things are getting more stable over time, so this may be revisited in the medium/long term.

jakirkham:
Ok happy to punt

Though let's see if we can find a simpler update path: https://github.com/rapidsai/build-planning/issues/125

jakirkham:
Looks like the wheels are not on PyPI yet. Hence those CI jobs fail

jakirkham (approved):
Thanks Bradley! 🙏

Approving as these look like the changes we want

We can restart CI when the remaining packages are uploaded

bdice:
Retargeting to 25.02.

bdice:
After discussion with @sameerz, I think we can hold off on taking this update. We may take a different nvcomp release in 25.02.

PR #574, opened 2024-12-09
Use curl >=8.5.0 to align with conda-forge and avoid CVEs.

bdice:
This PR uses `libcurl` 8.5.0 at build time, and should permit `>=8.5.0,<9.0a0` at runtime. This is needed to align with conda-forge which uses `libcurl` 8, and also gets a new enough minor version to avoid some known CVEs.

### DIFF
--- a/cpp/cmake/thirdparty/get_libcurl.cmake
+++ b/cpp/cmake/thirdparty/get_libcurl.cmake
@@ -27,7 +27,7 @@ function(find_and_configure_libcurl)
     BUILD_EXPORT_SET kvikio-exports
     INSTALL_EXPORT_SET kvikio-exports
     CPM_ARGS
-   GIT_REPOSITORY https://github.com/curl/curl
+    GIT_REPOSITORY https://github.com/curl/curl
     GIT_TAG curl-8_5_0
     OPTIONS "BUILD_CURL_EXE OFF" "BUILD_SHARED_LIBS OFF" "BUILD_TESTING OFF" "CURL_USE_LIBPSL OFF"
             "CURL_DISABLE_LDAP ON" "CMAKE_POSITION_INDEPENDENT_CODE ON"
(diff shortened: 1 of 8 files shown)


raydouglass (approved):

cpp/cmake/thirdparty/get_libcurl.cmake
```diff
@@ -22,13 +22,13 @@ function(find_and_configure_libcurl)
...
-    CURL 7.87.0
+    CURL 8.5.0
     GLOBAL_TARGETS libcurl
     BUILD_EXPORT_SET kvikio-exports
     INSTALL_EXPORT_SET kvikio-exports
     CPM_ARGS
-    GIT_REPOSITORY https://github.com/curl/curl
-    GIT_TAG curl-7_87_0
+   GIT_REPOSITORY https://github.com/curl/curl
+    GIT_TAG curl-8_5_0
```
jakirkham, line 31: Looks like [the style check failed]( https://github.com/rapidsai/kvikio/actions/runs/12244907972/job/34157647803?pr=574#step:7:1596 ) due to the indent change here

```suggestion
    GIT_REPOSITORY https://github.com/curl/curl
    GIT_TAG curl-8_5_0
```

vyasr (approved):
Looks fine to me.

Leaving some notes for posterity:
- The conda run exports are overly strict because libcurl has been ABI-stable since 7.16.0, so building against 7.* and running with 8.* is fine. 
- For conda, we shouldn't really need to bump the build version for CVE reasons since we could simply relax the run_exported constraint above and be fine. It's not really worthwhile though since [the conda-forge pinning is already at 8](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/4f4040e6ab06cbbc47c4c93f14cb50ad32605c3c/recipe/conda_build_config.yaml#L478-L479).
- For wheels, for CVE reasons we definitely need to bump the build version in addition to runtime constraints because we statically link libcurl into wheels so the build version entirely controls runtime behavior.

jakirkham:
> * The conda run exports are overly strict because libcurl has been ABI-stable since 7.16.0, so building against 7.* and running with 8.* is fine.

Please feel free to [raise a feedstock issue]( https://github.com/conda-forge/curl-feedstock/issues/new )

madsbk (approved):

PR #583, opened 2025-01-13
Zarr: pin to version 2.*

madsbk:
Now that Zarr v3 has been released, we need to pin to version 2 for now.


conda/recipes/kvikio/meta.yaml
```diff
@@ -75,7 +75,7 @@ requirements:
     - python
     - numpy >=1.23,<3.0a0
     - cupy >=12.0.0
-    - zarr
+    - zarr <3.0.0,>=2.0.0
```
bdice, line 78: Typically we use the order lower bound/upper bound, and require an alpha pinning on the upper bound (otherwise alphas/prereleases of 3.x are compatible).
```suggestion
    - zarr >=2.0.0,<3.0.0a0
```

dependencies.yaml
```diff
@@ -365,7 +365,7 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - numpy>=1.23,<3.0a0
-          - zarr
+          - zarr<3.0.0,>=2.0.0
```
bdice, line 368:
```suggestion
          - zarr>=2.0.0,<3.0.0a0
```

bdice (commented):
Two suggestions to apply and then rerun rapids-dependency-file-generator.

madsbk:
Thanks @bdice

PR #587, opened 2025-01-14
Fix the misplaced libcurl in CMake

kingcrimsontianyu:
This PR fixes a CMake bug introduced in https://github.com/rapidsai/kvikio/pull/581 where `libcurl` is unconditionally compiled.


KyleFromNVIDIA (approved):

bdice (approved):

abellina (approved):
I built spark-rapids-jni with this change, and we are unblocked. Thank you.

PR #588, opened 2025-01-20
Separate more interface and definition. Add comments on std::future. Mark noexcept to compat mode-related functions

kingcrimsontianyu:
This PR performs makes the following three improvements:
- Separates interface and definition for `file_handle.hpp` that was missed in the previous PR https://github.com/rapidsai/kvikio/pull/581.
- To help avoid UB (e.g. program crash) for downstream applications, adds the following qualifying remark to the returned future object of `pread/pwrite`:
  >The `std::future` object's `wait()` or `get()` should not be called after the lifetime of the FileHandle object ends. Otherwise, the behavior is undefined.
- Add `noexcept` specifier to compatibility mode-related functions.

### DIFF
--- a/cpp/src/file_handle.cpp
+++ b/cpp/src/file_handle.cpp
@@ -190,15 +190,18 @@ bool FileHandle::closed() const noexcept { return !_initialized; }
 
 void FileHandle::close() noexcept
 {
-  if (closed()) { return; }
+  try {
+    if (closed()) { return; }
 
-  if (!is_compat_mode_preferred()) { cuFileAPI::instance().HandleDeregister(_handle); }
-  _compat_mode = CompatMode::AUTO;
-  ::close(_fd_direct_off);
-  if (_fd_direct_on != -1) { ::close(_fd_direct_on); }
-  _fd_direct_on  = -1;
-  _fd_direct_off = -1;
-  _initialized   = false;
+    if (!is_compat_mode_preferred()) { cuFileAPI::instance().HandleDeregister(_handle); }
+    _compat_mode = CompatMode::AUTO;
+    ::close(_fd_direct_off);
+    if (_fd_direct_on != -1) { ::close(_fd_direct_on); }
+    _fd_direct_on  = -1;
+    _fd_direct_off = -1;
+    _initialized   = false;
+  } catch (...) {
+  }
 }
 
 CUfileHandle_t FileHandle::handle()
--- a/cpp/src/file_handle.cpp
+++ b/cpp/src/file_handle.cpp
@@ -163,6 +163,55 @@ FileHandle::FileHandle(const std::string& file_path,
   }
 }
 
+FileHandle::FileHandle(FileHandle&& o) noexcept
+  : _fd_direct_on{std::exchange(o._fd_direct_on, -1)},
+    _fd_direct_off{std::exchange(o._fd_direct_off, -1)},
+    _initialized{std::exchange(o._initialized, false)},
+    _compat_mode{std::exchange(o._compat_mode, CompatMode::AUTO)},
+    _nbytes{std::exchange(o._nbytes, 0)},
+    _handle{std::exchange(o._handle, CUfileHandle_t{})}
+{
+}
+
+FileHandle& FileHandle::operator=(FileHandle&& o) noexcept
+{
+  _fd_direct_on  = std::exchange(o._fd_direct_on, -1);
+  _fd_direct_off = std::exchange(o._fd_direct_off, -1);
+  _initialized   = std::exchange(o._initialized, false);
+  _compat_mode   = std::exchange(o._compat_mode, CompatMode::AUTO);
+  _nbytes        = std::exchange(o._nbytes, 0);
+  _handle        = std::exchange(o._handle, CUfileHandle_t{});
+  return *this;
+}
+
+FileHandle::~FileHandle() noexcept { close(); }
+
+bool FileHandle::closed() const noexcept { return !_initialized; }
+
+void FileHandle::close() noexcept
+{
+  if (closed()) { return; }
+
+  if (!is_compat_mode_preferred()) { cuFileAPI::instance().HandleDeregister(_handle); }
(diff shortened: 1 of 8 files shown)


cpp/src/file_handle.cpp
```diff
@@ -163,6 +163,55 @@ FileHandle::FileHandle(const std::string& file_path,
...
+
+FileHandle::~FileHandle() noexcept { close(); }
+
+bool FileHandle::closed() const noexcept { return !_initialized; }
+
+void FileHandle::close() noexcept
+{
+  if (closed()) { return; }
+
+  if (!is_compat_mode_preferred()) { cuFileAPI::instance().HandleDeregister(_handle); }
```
madsbk, line 195: Unrelated to this refactoring, but might make sense to fix in this PR:
We need an exception catch or make `is_compat_mode_preferred()` and `compat_mode()` `noexcept` as well.

madsbk (approved):

cpp/src/shim/cufile.cpp
```diff
@@ -123,28 +123,28 @@ void cuFileAPI::driver_close()
 }
 
 #ifdef KVIKIO_CUFILE_FOUND
-bool is_cufile_library_available()
+bool is_cufile_library_available() noexcept
 {
   try {
     cuFileAPI::instance();
-  } catch (const std::runtime_error&) {
+  } catch (...) {
```
kingcrimsontianyu, line 130: Imho `...` is better since it simplifies the effort to cover all possible exception types that the try block may generate (the exact types can be documented in `cuFileAPI::instance()`'s `@throw` section) while satisfying the `noexcept` promise. But I'm ready to revert if there's any other consideration I'm not aware of. @madsbk @vuule

cpp/src/file_handle.cpp
```diff
@@ -163,6 +163,58 @@ FileHandle::FileHandle(const std::string& file_path,
...
+    if (closed()) { return; }
+
+    if (!is_compat_mode_preferred()) { cuFileAPI::instance().HandleDeregister(_handle); }
+    _compat_mode = CompatMode::AUTO;
+    ::close(_fd_direct_off);
+    if (_fd_direct_on != -1) { ::close(_fd_direct_on); }
+    _fd_direct_on  = -1;
+    _fd_direct_off = -1;
+    _initialized   = false;
+  } catch (...) {
```
kingcrimsontianyu, line 203: `cuFileAPI` constructor (hence `cuFileAPI::instance()`) is potentially throwing. So to satisfy the `noexcept` promise, the try-catch block has to be here.

cpp/src/file_handle.cpp
kingcrimsontianyu: This is interesting: `defaults::compat_mode()` can't be `noexcept` because it calls `defaults::instance()` and `defaults`'s constructor is potentially throwing. One workaround is:
```cpp
CompatMode defaults::compat_mode() noexcept { 
    try {
        return instance()->_compat_mode; 
    } catch(...) {}
    return CompatMode::ON; // Or we may add an CompatMode::Invalid?
}
```
But is it worth the hassle to make it `noexcept` at all?
@wence- @madsbk

madsbk: > But is it worth the hassle to make it noexcept at all?

Heh, maybe not :)

madsbk: Side note, we can use `noexcept` on functions that might throw exceptions, it just makes them fatal.

cpp/src/shim/cufile.cpp
madsbk: Agree, `...` is better!

wence-: +2

cpp/src/shim/utils.cpp
```diff
@@ -45,23 +45,27 @@ void* load_library(const std::vector<const char*>& names, int mode)
...
-    return name.find("icrosoft") != std::string::npos;
+  try {
+    struct utsname buf {};
+    int err = ::uname(&buf);
+    if (err == 0) {
+      const std::string name(static_cast<char*>(buf.release));
+      // 'Microsoft' for WSL1 and 'microsoft' for WSL2
+      return name.find("icrosoft") != std::string::npos;
+    }
+    return false;
```
wence-, line 58: *question*: what might throw here?

wence- (approved):
One query about the catch in `is_running_in_wsl`, otherwise looks good!

cpp/src/file_handle.cpp
madsbk: > It's UB, rather than just making things fatal

It is UB in the sense that calling `std::terminate()` is UB (it is implementation-defined whether any stack unwinding). 
But `std::terminate()` is guaranteed to be called when violating `noexcept`: https://en.cppreference.com/w/cpp/error/terminate

wence-: Ah sorry, I was wrong, it is not UB, so all good.

cpp/src/shim/utils.cpp
kingcrimsontianyu: I was a bit paranoid about STL function calls. Here the string's constructor `basic_string(const charT* s, const Allocator& a = Allocator());` should be potentially throwing, because it does not have a "Throws" section in its [specs](https://eel.is/c++draft/string.cons#16). According to the [specs](https://eel.is/c++draft/res.on.exception.handling#4):

> Functions defined in the C++ standard library that do not have a Throws: paragraph but do have a potentially-throwing exception specification may throw implementation-defined exceptions.

> In particular, they can report a failure to allocate storage by throwing an exception of type bad_alloc, or a class derived from bad_alloc

I would imagine the rare but not impossible case in practice, where the stack may overflow (when short string optimization is in effect), or the heap may run out of memory.

The string find overload `size_type F(const charT* s, size_type pos) const;` is on the same boat.

wence-: Ah thanks!

PR #589, opened 2025-01-21
allow deselecting nvcomp wheels

jameslamb:
## Description

Follow-up to #478

Adds a matrix filter `use_cuda_wheels` in the `dependencies.yaml` list used for `libkvikio`'s `nvcomp` dependency. Some types of builds (like RAPIDS devcontainers) prefer to use the system-installed nvCOMP to one provided by wheels.

This ensures that preference is respected, because those builds pass matrix selector `use_cuda_wheels=false` through `rapids-dependency-file-generator` (https://github.com/rapidsai/devcontainers/pull/382).

## Notes for Reviewers

Similar changes in cuDF: https://github.com/rapidsai/cudf/pull/17774

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -295,16 +295,25 @@ dependencies:
         packages:
           - nvcomp==4.1.0.6
     specific:
-      - output_types: [requirements, pyproject]
+      - output_types: [pyproject, requirements]
         matrices:
           - matrix:
               cuda: "12.*"
+              use_cuda_wheels: "true"
             packages:
               - nvidia-nvcomp-cu12==4.1.0.6
           - matrix:
               cuda: "11.*"
+              use_cuda_wheels: "true"
             packages:
               - nvidia-nvcomp-cu11==4.1.0.6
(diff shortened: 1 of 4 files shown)


dependencies.yaml
```diff
@@ -295,16 +295,25 @@ dependencies:
         packages:
           - nvcomp==4.1.0.6
     specific:
-      - output_types: [requirements, pyproject]
+      - output_types: [pyproject, requirements]
```
jameslamb, line 298:
```suggestion
      - output_types: pyproject
```

dependencies.yaml
```diff
@@ -295,16 +295,25 @@ dependencies:
         packages:
           - nvcomp==4.1.0.6
     specific:
-      - output_types: [requirements, pyproject]
+      - output_types: pyproject
```
jameslamb, line 298:
```suggestion
      - output_types: [requirements, pyproject]
```

bdice (approved):

Issue #593, opened 2025-01-22
Segmentation fault when using RemoteFile.pread on multiple remote files


TomAugspurger:
The following script, which uses `RemoteFile.open_s3` followed by a `pread` on multiple s3 objects, segfaults.

```python
import cupy as cp
import kvikio

BUCKET = "kvikiobench-56481"

def main():

    paths = [
        f"data/parquet-large/timeseries.parquet/part.{i}.parquet" for i in range(4)
    ]

    def pread_one(path):
        rf = kvikio.RemoteFile.open_s3(BUCKET, path)
        buf = cp.empty(rf.nbytes() // 8)
        return rf.pread(buf)

    not_done = {pread_one(path) for path in paths}

    # spin lock, till we're done
    while not_done:
        not_done = {
            x for x in not_done if not x.done()
        }

    print("Parallel done")


if __name__ == "__main__":
    main()
```

I'm not a gdb expert, but here's some output indicating the failure happens in `kvikio::RemoteHandle::read`:

```
$ gdb python
[...]
Reading symbols from python...
(gdb) run bug.py
Starting program: /home/ubuntu/miniforge3/envs/kvikio-env/bin/python bug.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff4400640 (LWP 8711)]
[New Thread 0x7ffff3a00640 (LWP 8712)]
[New Thread 0x7fffe3000640 (LWP 8713)]
[Detaching after vfork from child process 8714]
[New Thread 0x7fffd3c00640 (LWP 8715)]
[Thread 0x7fffd3c00640 (LWP 8715) exited]
[New Thread 0x7fffd3c00640 (LWP 8716)]
[New Thread 0x7fffc5e00640 (LWP 8720)]
[New Thread 0x7fffc1e00640 (LWP 8721)]
[New Thread 0x7fffc1400640 (LWP 8722)]

Thread 8 "python" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffc1e00640 (LWP 8721)]
0x00007fffd428bf97 in kvikio::RemoteHandle::read(void*, unsigned long, unsigned long) () from /home/ubuntu/miniforge3/envs/kvikio-env/lib/python3.12/site-packages/kvikio/_lib/../../../../libkvikio.so
```

As an aside, would this be the recommended way to read multiple objects in parallel, without having to use Python threads?

[...]

PR #596, opened 2025-01-26
Fix std::future status query

kingcrimsontianyu:
This PR fixes the `std::future` status query according to the discussion in #593 .

- The future objects of parallel I/O operations are no longer created using `std::async` with the deferred launch policy. Doing so would cause the future status to be falsely reported as done. Instead, as suggested, they are created by the underlying thread pool (more specifically by the promise objects). This enables non-blocking future status query.
- The trivial I/O cases that do not involve task splitting are now performed in a synchronous, blocking way as suggested, instead of being evaluated lazily via `std::async` with the deferred launch policy.

### DIFF
--- a/cpp/include/kvikio/parallel_operation.hpp
+++ b/cpp/include/kvikio/parallel_operation.hpp
@@ -110,11 +110,17 @@ std::future<std::size_t> submit_task(F op,
  * @return A future to be used later to check if the operation has finished its execution.
  */
 template <typename F>
-std::future<std::size_t> submit_move_only_task(F op_move_only)
+std::future<std::size_t> submit_move_only_task(
+  F op_move_only,
+  std::uint64_t nvtx_payload = 0ull,
+  nvtx_color_type nvtx_color = nvtx_manager::default_color())
 {
   static_assert(std::is_invocable_r_v<std::size_t, F>);
   auto op_copyable = make_copyable_lambda(std::move(op_move_only));
-  return defaults::thread_pool().submit_task(op_copyable);
+  return defaults::thread_pool().submit_task([=] {
+    KVIKIO_NVTX_SCOPED_RANGE("task", nvtx_payload, nvtx_color);
+    return op_copyable();
+  });
 }
 
 }  // namespace detail
@@ -154,34 +160,34 @@ std::future<std::size_t> parallel_io(F op,
     return detail::submit_task(op, buf, size, file_offset, devPtr_offset, call_idx, nvtx_color);
   }
 
-  // We know an upper bound of the total number of tasks
-  std::vector<std::future<std::size_t>> tasks;
-  tasks.reserve(size / task_size + 2);
+  std::vector<std::future<std::size_t>> tasks_before_last;
+  auto const num_full_tasks  = size / task_size;
+  auto const remaining_bytes = size - num_full_tasks * task_size;
+  auto num_tasks{num_full_tasks};
+  if (remaining_bytes != 0) { ++num_tasks; }
+  tasks_before_last.reserve(num_tasks - 1);
 
-  // 1) Submit `task_size` sized tasks
-  while (size >= task_size) {
-    tasks.push_back(
+  // 1) Submit all tasks before the last one. These are all `task_size` sized tasks.
+  for (std::size_t i = 0; i < num_tasks - 1; ++i) {
+    tasks_before_last.push_back(
       detail::submit_task(op, buf, task_size, file_offset, devPtr_offset, call_idx, nvtx_color));
     file_offset += task_size;
     devPtr_offset += task_size;
     size -= task_size;
   }
 
-  // 2) Submit a task for the remainder
-  if (size > 0) {
-    tasks.push_back(
-      detail::submit_task(op, buf, size, file_offset, devPtr_offset, call_idx, nvtx_color));
-  }
+  // 2) Submit the last task
+  auto last_task_size = (remaining_bytes == 0) ? task_size : remaining_bytes;
 
-  // Finally, we sum the result of all tasks.
-  auto gather_tasks = [tasks = std::move(tasks)]() mutable -> std::size_t {
-    std::size_t ret = 0;
-    for (auto& task : tasks) {
+  auto last_task = [=, tasks_before_last = std::move(tasks_before_last)]() mutable -> std::size_t {
+    std::size_t ret = op(buf, last_task_size, file_offset, devPtr_offset);
+    for (auto& task : tasks_before_last) {
       ret += task.get();
     }
     return ret;
   };
-  return detail::submit_move_only_task(std::move(gather_tasks));
+
+  return detail::submit_move_only_task(std::move(last_task), call_idx, nvtx_color);
 }
 
 }  // namespace kvikio
--- a/cpp/include/kvikio/parallel_operation.hpp
+++ b/cpp/include/kvikio/parallel_operation.hpp
@@ -160,16 +160,12 @@ std::future<std::size_t> parallel_io(F op,
     return detail::submit_task(op, buf, size, file_offset, devPtr_offset, call_idx, nvtx_color);
   }
 
-  std::vector<std::future<std::size_t>> tasks_before_last;
-  auto const num_full_tasks  = size / task_size;
-  auto const remaining_bytes = size - num_full_tasks * task_size;
-  auto num_tasks{num_full_tasks};
-  if (remaining_bytes != 0) { ++num_tasks; }
-  tasks_before_last.reserve(num_tasks - 1);
+  std::vector<std::future<std::size_t>> tasks;
+  tasks.reserve(size / task_size);
 
   // 1) Submit all tasks but the last one. These are all `task_size` sized tasks.
-  for (std::size_t i = 0; i < num_tasks - 1; ++i) {
-    tasks_before_last.push_back(
+  while (size > task_size) {
+    tasks.push_back(
       detail::submit_task(op, buf, task_size, file_offset, devPtr_offset, call_idx, nvtx_color));
     file_offset += task_size;
     devPtr_offset += task_size;
@@ -178,16 +174,13 @@ std::future<std::size_t> parallel_io(F op,
 
   // 2) Submit the last task, which consists of performing the last I/O and waiting the previous
   // tasks.
-  auto last_task_size = (remaining_bytes == 0) ? task_size : remaining_bytes;
-
-  auto last_task = [=, tasks_before_last = std::move(tasks_before_last)]() mutable -> std::size_t {
-    std::size_t ret = op(buf, last_task_size, file_offset, devPtr_offset);
-    for (auto& task : tasks_before_last) {
+  auto last_task = [=, tasks = std::move(tasks)]() mutable -> std::size_t {
+    auto ret = op(buf, size, file_offset, devPtr_offset);
+    for (auto& task : tasks) {
       ret += task.get();
     }
     return ret;
   };
-
   return detail::submit_move_only_task(std::move(last_task), call_idx, nvtx_color);
 }
 
--- a/cpp/include/kvikio/parallel_operation.hpp
+++ b/cpp/include/kvikio/parallel_operation.hpp
@@ -49,10 +49,8 @@ namespace detail {
 template <typename F>
 auto make_copyable_lambda(F op)
 {
-  static_assert(std::is_move_constructible_v<F>);
-
-  // Create the callable on the heap by moving from f. Use a shared pointer to manage its lifetime.
-  auto sp = std::make_shared<F>(std::forward<F>(op));
+  // Create the callable on the heap by moving from op. Use a shared pointer to manage its lifetime.
+  auto sp = std::make_shared<F>(std::move(op));
 
   // Use the copyable closure as the proxy of the move-only callable.
   return
--- a/cpp/include/kvikio/parallel_operation.hpp
+++ b/cpp/include/kvikio/parallel_operation.hpp
@@ -113,7 +113,7 @@ template <typename F>
 std::future<std::size_t> submit_move_only_task(
   F op_move_only,
   std::uint64_t nvtx_payload = 0ull,
-  nvtx_color_type nvtx_color = nvtx_manager::default_color())
+  nvtx_color_type nvtx_color = NvtxManager::default_color())
 {
   static_assert(std::is_invocable_r_v<std::size_t, F>);
   auto op_copyable = make_copyable_lambda(std::move(op_move_only));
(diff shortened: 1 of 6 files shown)


vyasr:
Discussing further in #593 rather than here.

madsbk (commented):
Looks good @kingcrimsontianyu. The only question is if the overhead of creating the extra _gather_ task is an issue. I suspect not, but it would be good to get it confirmed. 

cc. @GregoryKimball, @vuule.

cpp/include/kvikio/parallel_operation.hpp
```diff
@@ -30,12 +32,60 @@ namespace kvikio {
...
-  return defaults::thread_pool().submit_task(
-    [=] { return op(buf, size, file_offset, devPtr_offset); });
+  static_assert(std::is_invocable_r_v<std::size_t, std::decay_t<F>, Args...>);
+  return defaults::thread_pool().submit_task([=] { return op(args...); });
+}
+
+/**
+ * @brief Submit the move-only task callable to the underlying thread pool.
+ *
+ * @tparam F Callable type. F shall be move-only and have no argument.
```
kingcrimsontianyu, line 79: It is tempting to add the "can't copy" part to the code to improve type safety:
```cpp
template <typename F, std::enable_if_t<!std::is_copy_constructible_v<F>, bool> = true>
std::future<std::size_t> submit_move_only_task(F op_move_only)
```
But this does not work.
For the non-copyable `gather_tasks` below, trying to copy construct results in compile error:
```cpp
auto gather_tasks_copy = gather_tasks; // Compile error
```
But `std::is_copy_constructible_v<decltype(gather_tasks)> == true` still holds.
A simpler example:
```cpp
std::vector<std::future<int>> move_only;
static_assert(std::is_copy_constructible_v<decltype(move_only)>); // No compile error. Surprise.
```

vyasr:
I think this looks good overall to me now. Is it ready for a thorough review? I'm not sure if it is intentionally still in draft or not.

kingcrimsontianyu:
Thanks for the reminder. Other than pending performance assessment, this PR should be ready for review.

kingcrimsontianyu:
I'll use cuDF bench for performance assessment, and post the results here when ready.

cpp/include/kvikio/utils.hpp
```diff
@@ -152,6 +153,7 @@ std::tuple<void*, std::size_t, std::size_t> get_alloc_info(void const* devPtr,
 template <typename T>
 bool is_future_done(T const& future)
 {
+  assert(future.valid());
```
vyasr, line 156: Since this is a public API, do we want to throw if this condition is violated or are we happy with a debug assertion?

cpp/src/file_handle.cpp
```diff
@@ -293,11 +293,12 @@ std::future<std::size_t> FileHandle::pread(void* buf,
 
   // Shortcut that circumvent the threadpool and use the POSIX backend directly.
   if (size < gds_threshold) {
-    auto task = [this, ctx, buf, size, file_offset]() -> std::size_t {
-      PushAndPopContext c(ctx);
-      return detail::posix_device_read(_fd_direct_off, buf, size, file_offset, 0);
-    };
-    return std::async(std::launch::deferred, task);
+    PushAndPopContext c(ctx);
```
vyasr, line 296: Maybe add a note that this execution model provides API compatibility with the rest of the future-based APIs in the library while also ensuring that the future is actually immediately available since we don't want this call to be asynchronous.

vuule (approved):

cpp/include/kvikio/parallel_operation.hpp
vyasr: This is a good writeup: https://quuxplusone.github.io/blog/2020/02/05/vector-is-copyable-except-when-its-not/

vyasr (approved):
Thanks! I'll leave it to you to verify performance before merging.

cpp/src/file_handle.cpp
kingcrimsontianyu: Thanks. Done.
I realized that the C++ concurrency TS has a utility function meant to do the exact same thing, so I put a simplistic implementation there in `utils.hpp`:

https://github.com/rapidsai/kvikio/blob/b519cc2196e8507d1caa430a953b2e789733023c/cpp/include/kvikio/utils.hpp#L165

cpp/include/kvikio/utils.hpp
kingcrimsontianyu: Thanks! I think we do need to check the pre-condition to avoid UB. Done.

madsbk (approved):

kingcrimsontianyu:
Increase the thread number N by 1, and compare `(N+1) this PR` vs `(N) branch-25.04`:

|            |                         | 5 threads      |       |                                  | 9 threads      |       |                                  | 17 threads     |       |                                   |
| ---------- | ----------------------- | -------------- | ----- | -------------------------------- | -------------- | ----- | -------------------------------- | -------------- | ----- | --------------------------------- |
|            | orc_read_io_compression | CPU Time (sec) | Noise | Time increase vs 4 threads 25.04 | CPU Time (sec) | Noise | Time increase vs 8 threads 25.04 | CPU Time (sec) |       | Time increase vs 16 threads 25.04 |
| cold cache | This PR                 | 0.131          | 0.034 | 19.43%                           | 0.140          | 0.052 | 21.32%                           | 0.138          | 0.017 | 11.96%                            |
|            |                         |                |       |                                  |                |       |                                  |                |       |

madsbk:
@kingcrimsontianyu, so the benchmark are saying that for large IO the new PR is good and for small IO the new PR is bad? 

Could this be an ordering issue? 
Consider a mix of small and large IOs in an arbitrary order. In `branch-25.04` all of the large IOs will be submitted to the thread pool **before** the small IOs are being executed. 
Whereas, in this PR, the small IOs are executed immediately and thus delaying the submission of the following large IOs.

kingcrimsontianyu:
@madsbk We had a sync last week, and yes! this is what we suspected was happening to the ORC benchmark. I'll do some profiling this week to hopefully better understand its I/O pattern.

cpp/include/kvikio/parallel_operation.hpp
```diff
@@ -60,37 +110,44 @@ std::future<std::size_t> parallel_io(F op,
...
-  std::vector<std::future<std::size_t>> tasks;
-  tasks.reserve(size / task_size + 2);
+  std::vector<std::future<std::size_t>> tasks_before_last;
+  auto const num_full_tasks  = size / task_size;
+  auto const remaining_bytes = size - num_full_tasks * task_size;
+  auto num_tasks{num_full_tasks};
+  if (remaining_bytes != 0) { ++num_tasks; }
+  tasks_before_last.reserve(num_tasks - 1);
 
-  // 1) Submit `task_size` sized tasks
```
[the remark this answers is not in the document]
kingcrimsontianyu: This PR initially made that implementation as you described: we submit the sequence of N tasks, and instead of using `std::async(std::launch::deferred)` to launch the gather task as the dev branch does, we submit the gather task to the task queue of the thread pool. However, the performance regression was observed, somehow only on the ORC benchmark in libcudf. But we have noticed that if we let the last task do the gather operation (for the N-1 tasks), the regression would disappear.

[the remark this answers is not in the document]
kingcrimsontianyu: PS: The suggested code is much more concise. Replaced as suggested. Thanks!

kingcrimsontianyu:
The profiling confirms that the regression occurring to ORC cold cache only is attributed to the special I/O pattern where small I/O and parallel I/O are mingled together, and a bunch of small I/O precede parallel I/O, causing delayed submission on the main thread and therefore delayed execution on the worker threads.

## This PR
image
- The first row is the small I/O on the main thread. The colorful first sub-row is the trace of the task. The white second sub-row is the trace of the actual POSIX read.
- The second row is the parallel I/O on the worker thread (for demonstration purpose this profile was done with 1 worker thread).

## Branch 25.04
image
- The first row is the small I/O on the main thread. The small colorful regions are the trace of the submission of the small I/O. The actual POSIX reads occur lazily at a later time on the same row (not shown in this figure).
- The second row is the parallel I/O on the worker thread. This figure is more zoomed in than the one above. For branch 25.04 parallel I/O starts ~75 microseconds later than small I/O, whereas for this PR this time difference increases to ~ 3.3 millisecond.

The parquet, json, and csv benchmark have rare occurrences of small I/O, and are therefore not affected.

kingcrimsontianyu:
For the cases with mixed small I/O and parallel I/O, one effective tuning approach for this PR is to reduce `KVIKIO_GDS_THRESHOLD` (so that more instances of small I/O are treated as parallel I/O and are run by the thread pool) while keeping a fair amount of threads running. 

For example, when `KVIKIO_NTHREADS=8`, if we set `KVIKIO_GDS_THRESHOLD=0`, the ORC bench time for this PR will reduce from ~0.14 to ~0.115, which is as good as branch 25.04.

kingcrimsontianyu:
The adverse performance impact of this PR on ORC cold cache can be mitigated by reducing the default GDS threshold parameter from 1 MiB to 16 KiB. 

Regardless of user-specified compatibility mode, I/O request with size lower than GDS threshold is forced to take the POSIX I/O path using `std::async`-based lazy evaluation (until this PR) / on the calling thread with `make_ready_future` (since this PR).

Previously, for 2, 4, 8, 16 threads, the time increase in ORC cold cache benchmark is 6.78%, 25.36%, 29.06% and 15.63%. Now these numbers are 13.56%, 10.45%, -2.88% and -9.84%. I consider this result acceptable.

Below is the complete re-benchmark.

## parquet_read_io_compression
|            |                             | 2 threads      |       |               | 4 threads      |       |               | 8 threads      |       |               | 16 threads     |       |               |
| ---------- | --------------------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- |
|            | parquet_read_io_compression | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase |
| cold cache | This PR                     | 0.135          | 0.014 | \-1.82%       | 0.100          | 0.015 | \-1.10%       | 0.096          | 0.012 | \-0.84%       | 0.098          | 0.010 | \-1.28%       |
|            | branch-25.04                | 0.138          | 0.028 |               | 0.102          | 0.028 |               | 0.097          | 0.022 |               | 0.099          | 0.021 |               |
| hot cache  | This PR                     | 0.054          | 0.009 | \-0.15%       | 0.045          | 0.008 | \-4.07%       | 0.043          | 0.004 | \-3.26%       | 0.044          | 0.013 | \-0.86%       |
|            | branch-25.04                | 0.054          | 0.010 |               | 0.047          | 0.041 |               | 0.045          | 0.047 |               | 0.045          | 0.035 |               |

## orc_read_io_compression
|            |                             | 2 threads      |       |               | 4 threads      |       |               | 8 threads      |       |               | 16 threads     |       |               |
| ---------- | --------------------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- |
|            | orc_read_io_compression     | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) |       | Time increase |
| cold cache | This PR                     | 0.157          | 0.030 | 13.56%        | 0.121          | 0.008 | 10.45%        | 0.111          | 0.013 | \-2.88%       | 0.112          | 0.009 | \-9.84%       |
|            | branch-25.04                | 0.138          | 0.013 |               | 0.110          | 0.023 |               | 0.115          | 0.023 |               | 0.124          | 0.037 |               |
| hot cache  | This PR                     | 0.067          | 0.005 | 5.90%         | 0.061          | 0.020 | \-0.54%       | 0.059          | 0.007 | \-3.95%       | 0.057          | 0.053 | \-9.50%       |
|            | branch-25.04                | 0.063          | 0.010 |               | 0.061          | 0.051 |               | 0.062          | 0.032 |               | 0.063          | 0.034 |               |

## json_read_io
|            |                             | 2 threads      |       |               | 4 threads      |       |               | 8 threads      |       |               | 16 threads     |       |               |
| ---------- | --------------------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- |
|            | json_read_io                | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) |       | Time increase |
| cold cache | This PR                     | 0.557          | 0.025 | 3.34%         | 0.464          | 0.036 | \-0.47%       | 0.451          | 0.045 | 0.52%         | 0.477          | 0.047 | \-3.42%       |
|            | branch-25.04                | 0.539          | 0.008 |               | 0.466          | 0.037 |               | 0.449          | 0.034 |               | 0.494          | 0.013 |               |
| hot cache  | This PR                     | 0.340          | 0.018 | 0.31%         | 0.327          | 0.054 | \-0.29%       | 0.309          | 0.004 | \-1.62%       | 0.317          | 0.039 | \-1.67%       |
|            | branch-25.04                | 0.339          | 0.006 |               | 0.328          | 0.054 |               | 0.314          | 0.024 |               | 0.322          | 0.041 |               |

## csv_read_io
|            |                             | 2 threads      |       |               | 4 threads      |       |               | 8 threads      |       |               | 16 threads     |       |               |
| ---------- | --------------------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- | -------------- | ----- | ------------- |
|            | csv_read_io                 | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) | Noise | Time increase | CPU Time (sec) |       | Time increase |
| cold cache | This PR                     | 0.364          | 0.041 | 5.44%         | 0.302          | 0.043 | \-0.87%       | 0.302          | 0.041 | 1.91%         | 0.323          | 0.047 | 1.89%         |
|            | branch-25.04                | 0.345          | 0.016 |               | 0.305          | 0.043 |               | 0.296          | 0.011 |               | 0.317          | 0.039 |               |
| hot cache  | This PR                     | 0.236          | 0.062 | 5.40%         | 0.221          | 0.057 | 3.96%         | 0.210          | 0.025 | \-1.47%       | 0.216          | 0.023 | \-2.02%       |
|            | branch-25.04                | 0.224          | 0.017 |               | 0.212          | 0.028 |               | 0.213          | 0.045 |               | 0.221          | 0.051 |               |

madsbk (approved):
Thanks @kingcrimsontianyu, looks good

vyasr (approved):
I'm happy with the final performance numbers here.

PR #600, opened 2025-01-28
Build and test with CUDA 12.8.0

bdice:
This PR uses CUDA 12.8.0 to build and test.

xref: https://github.com/rapidsai/build-planning/issues/139

### DIFF
--- a/cpp/doxygen/main_page.md
+++ b/cpp/doxygen/main_page.md
@@ -36,7 +36,7 @@ mamba install -c rapidsai -c conda-forge libkvikio
 # Create new environment (CUDA 11.8)
 mamba create -n libkvikio-env -c rapidsai -c conda-forge cuda-version=11.8 libkvikio
 # Create new environment (CUDA 12.5)
-mamba create -n libkvikio-env -c rapidsai -c conda-forge cuda-version=12.5 libkvikio
+mamba create -n libkvikio-env -c rapidsai -c conda-forge cuda-version=12.8 libkvikio
 ```
 
 Install the **nightly release** from the ``rapidsai-nightly`` channel with the following:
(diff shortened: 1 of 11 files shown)


cpp/doxygen/main_page.md
```diff
@@ -47,7 +47,7 @@ mamba install -c rapidsai-nightly -c conda-forge libkvikio
 # Create new environment (CUDA 11.8)
 mamba create -n libkvikio-env -c rapidsai-nightly -c conda-forge python=3.12 cuda-version=11.8 libkvikio
 # Create new environment (CUDA 12.5)
-mamba create -n libkvikio-env -c rapidsai-nightly -c conda-forge python=3.12 cuda-version=12.5 libkvikio
+mamba create -n libkvikio-env -c rapidsai-nightly -c conda-forge python=3.12 cuda-version=12.8 libkvikio
```
jameslamb, line 50: The comment above this line should also be changed (currently reads "CUDA 12.5"). And similar changes a few lines up.

To avoid mistakes like that in the future, I think it'd probably be better to just change those parentheticals to `(CUDA 11)` and `(CUDA 12)`.

This needs to be updated too, now that the file it references is being removed from the repo:

https://github.com/rapidsai/kvikio/blob/716a99f212ce2e0f9862bad1be81a12c56d03dc6/docs/source/install.rst#L48-L49

jameslamb (approved):
Found a few small things that should be changed in docs, but otherwise looks fine. Approving so you don't need a re-review.

cpp/doxygen/main_page.md
bdice: Improved in e83f99c.

vyasr (approved):

Issue #607, opened 2025-01-30
FileHandle file resource may leak if the constructor throws an exception


kingcrimsontianyu:
The constructor of `FileHandle` may create the following resources:
 
- File resource created by the system call `open` without the `O_DIRECT` flag.
- File resource created by the system call `open` with the `O_DIRECT` flag.

There are two possibilities where the constructor may throw exceptions and cause trouble:

- The compatibility mode is set to `OFF` but the file cannot be opened with the `O_DIRECT` flag.
- The compatibility mode is set to `OFF` but the `cuFileHandleRegister` fails, for instance, due to unsupported or misconfigured system for cuFile I/O.

When the constructor throws, the destructor will not be called, resulting in resource leak. This issue can be fixed by using an RAII file resource manager.

PR #608, opened 2025-01-31
Adjust the way of determining FileHandle's compatibility mode for sync and async I/O to improve code readability

kingcrimsontianyu:
This PR improves the readability of compatibility mode handling.

The current way of determining FileHandle's compatibility mode is somewhat complicated and unintuitive. The data member `_compat_mode` accompanied by some utility functions more or less combines 3 different things into one:

- The initially requested compat mode (`ON`/`OFF`/`AUTO`)
- The capability of performing synchronous cuFile I/O (bool)
- The capability of performing asynchronous cufile I/O (bool)

The disadvantages include:
- `FileHandle::is_compat_mode_preferred()` always derives the preferred compat mode on the fly as opposed to getting an already determined value.
- `FileHandle::is_compat_mode_preferred_for_async(CompatMode)` is potentially throwing, which is asymmetric to `is_compat_mode_preferred()`. Also when the compat mode is `OFF`, it has to invoke `is_stream_api_available()` and `config_path()` on each pass instead of getting an already determined value.
- There is no way to retrieve what the original requested compat mode is.

These add to cognitive burden when rereading the source to introduce new features to FileHandle. This PR attempts to improve the logic by making it concise and crystal clear.

This PR also fixes a line number bug in error handling.

This PR is breaking in that the rarely used public functions to query the compat mode data in the `FileHandle` are removed. These data are instead queryable via the new `CompatModeManager` class.

### DIFF
--- a/cpp/include/kvikio/compat_mode.hpp
+++ b/cpp/include/kvikio/compat_mode.hpp
@@ -49,9 +49,11 @@ CompatMode parse_compat_mode_str(std::string_view compat_mode_str);
 
 }  // namespace detail
 
+// Forward declaration.
+class FileHandle;
+
 /**
- * @brief
- *
+ * @brief Store and manage the compatibility mode data associated with a FileHandle.
  */
 class CompatModeManager {
  private:
@@ -60,36 +62,87 @@ class CompatModeManager {
   bool _is_compat_mode_preferred_for_async{true};
 
  public:
+  /**
+   * @brief Construct an empty compatibility mode manager.
+   */
   CompatModeManager() noexcept = default;
+
+  /**
+   * @brief Construct a compatibility mode manager associated with a FileHandle.
+   *
+   * According to the file path, requested compatibility mode, and the system configuration, the
+   * compatibility manager:
+   * - Infers the final compatibility modes for synchronous and asynchronous I/O paths,
+   * respectively.
+   * - Initializes the file wrappers and cuFile handle associated with a FileHandle.
+   *
+   * @param file_path Refer to
+   * FileHandle::FileHandle(std::string const&, std::string const&, mode_t, CompatMode).
+   * @param flags Same as above.
+   * @param mode Same as above.
+   * @param compat_mode_requested Same as above.
+   * @param file_handle Point to the FileHandle object that owns this compatibility mode manager.
+   */
   CompatModeManager(std::string const& file_path,
                     std::string const& flags,
                     mode_t mode,
                     CompatMode compat_mode_requested,
-                    FileWrapper& file_direct_on,
-                    FileWrapper& file_direct_off,
-                    CUFileHandleWrapper& cufile_handle);
+                    FileHandle* file_handle);
+
   ~CompatModeManager() noexcept                              = default;
   CompatModeManager(const CompatModeManager&)                = default;
   CompatModeManager& operator=(const CompatModeManager&)     = default;
   CompatModeManager(CompatModeManager&&) noexcept            = default;
   CompatModeManager& operator=(CompatModeManager&&) noexcept = default;
 
-  void compat_mode_reset(CompatMode compat_mode_requested);
-
+  /**
+   * @brief Functionally identical to `defaults::infer_compat_mode_if_auto(CompatMode)`.
+   *
+   * @param compat_mode Compatibility mode.
+   * @return If the given compatibility mode is CompatMode::AUTO, infer the final compatibility
+   * mode.
+   */
   CompatMode infer_compat_mode_if_auto(CompatMode compat_mode) noexcept;
 
+  /**
+   * @brief Functionally identical to `defaults::is_compat_mode_preferred(CompatMode)`.
+   *
+   * @param compat_mode Compatibility mode.
+   * @return Boolean answer.
+   */
   bool is_compat_mode_preferred(CompatMode compat_mode) noexcept;
 
+  /**
+   * @brief Check if the compatibility mode for synchronous I/O of the associated FileHandle is
+   * expected to be CompatMode::ON.
+   *
+   * @return Boolean answer.
+   */
   bool is_compat_mode_preferred() const noexcept;
 
+  /**
+   * @brief Check if the compatibility mode for asynchronous I/O of the associated FileHandle is
+   * expected to be CompatMode::ON.
+   *
+   * @return Boolean answer.
+   */
   bool is_compat_mode_preferred_for_async() const noexcept;
 
+  /**
+   * @brief Retrieve the original compatibility mode requested.
+   *
+   * @return The original compatibility mode requested.
+   */
   CompatMode compat_mode_requested() const noexcept;
 
   /**
    * @brief Determine if the asynchronous I/O should be performed or not (throw exceptions)
-   * according to `_compat_mode_requested`, `_is_compat_mode_preferred`, and
-   * `_is_compat_mode_preferred_for_async`.
+   * according to the existing compatibility mode data in the manager.
+   *
+   * The asynchronous I/O should not be performed, for instance, when compat_mode_requested() is
+   * CompatMode::OFF, is_compat_mode_preferred() is CompatMode::OFF, but
+   * is_compat_mode_preferred_for_async() is CompatMode::ON (due to missing cuFile stream API or
+   * cuFile configuration file).
    */
   void validate_compat_mode_for_async();
 };
--- a/cpp/include/kvikio/compat_mode.hpp
+++ b/cpp/include/kvikio/compat_mode.hpp
@@ -60,12 +60,18 @@ class CompatModeManager {
   bool _is_compat_mode_preferred_for_async{true};
 
  public:
-  CompatModeManager()                                    = default;
-  ~CompatModeManager() noexcept                          = default;
-  CompatModeManager(const CompatModeManager&)            = default;
-  CompatModeManager& operator=(const CompatModeManager&) = default;
-  CompatModeManager(CompatModeManager&&) noexcept;
-  CompatModeManager& operator=(CompatModeManager&&) noexcept;
+  CompatModeManager(std::string const& file_path,
+                    std::string const& flags,
+                    mode_t mode,
+                    CompatMode compat_mode_requested,
+                    FileWrapper& file_direct_on,
+                    FileWrapper& file_direct_off,
+                    CUFileHandleWrapper& cufile_handle);
+  ~CompatModeManager() noexcept                              = default;
+  CompatModeManager(const CompatModeManager&)                = default;
+  CompatModeManager& operator=(const CompatModeManager&)     = default;
+  CompatModeManager(CompatModeManager&&) noexcept            = default;
+  CompatModeManager& operator=(CompatModeManager&&) noexcept = default;
 
   void compat_mode_reset(CompatMode compat_mode_requested);
 
@@ -78,9 +84,6 @@ class CompatModeManager {
   bool is_compat_mode_preferred_for_async() const noexcept;
 
   CompatMode compat_mode_requested() const noexcept;
-
-  std::tuple<FileWrapper, FileWrapper, CUFileHandleWrapper> resolve_compat_mode_for_file(
-    std::string const& file_path, std::string const& flags, mode_t mode, CompatMode compat_mode);
 
   /**
    * @brief Determine if the asynchronous I/O should be performed or not (throw exceptions)
--- a/cpp/include/kvikio/compat_mode.hpp
+++ b/cpp/include/kvikio/compat_mode.hpp
@@ -96,7 +96,7 @@ class CompatModeManager {
   CompatModeManager& operator=(CompatModeManager&&) noexcept = default;
 
   /**
-   * @brief Functionally identical to `defaults::infer_compat_mode_if_auto(CompatMode)`.
+   * @brief Functionally identical to defaults::infer_compat_mode_if_auto(CompatMode).
    *
    * @param compat_mode Compatibility mode.
    * @return If the given compatibility mode is CompatMode::AUTO, infer the final compatibility
@@ -105,7 +105,7 @@ class CompatModeManager {
   CompatMode infer_compat_mode_if_auto(CompatMode compat_mode) noexcept;
 
   /**
-   * @brief Functionally identical to `defaults::is_compat_mode_preferred(CompatMode)`.
+   * @brief Functionally identical to defaults::is_compat_mode_preferred(CompatMode).
    *
    * @param compat_mode Compatibility mode.
    * @return Boolean answer.
@@ -136,10 +136,10 @@ class CompatModeManager {
   CompatMode compat_mode_requested() const noexcept;
 
   /**
-   * @brief Determine if the asynchronous I/O should be performed or not (throw exceptions)
+   * @brief Determine if the asynchronous I/O can be performed or not (throw exceptions)
    * according to the existing compatibility mode data in the manager.
    *
-   * The asynchronous I/O should not be performed, for instance, when compat_mode_requested() is
+   * The asynchronous I/O  cannot be performed, for instance, when compat_mode_requested() is
    * CompatMode::OFF, is_compat_mode_preferred() is CompatMode::OFF, but
    * is_compat_mode_preferred_for_async() is CompatMode::ON (due to missing cuFile stream API or
    * cuFile configuration file).
--- a/cpp/include/kvikio/compat_mode.hpp
+++ b/cpp/include/kvikio/compat_mode.hpp
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <string>
+
+#include <kvikio/file_utils.hpp>
+
+namespace kvikio {
+/**
+ * @brief I/O compatibility mode.
+ */
+enum class CompatMode : uint8_t {
+  OFF,  ///< Enforce cuFile I/O. GDS will be activated if the system requirements for cuFile are met
+        ///< and cuFile is properly configured. However, if the system is not suited for cuFile, I/O
+        ///< operations under the OFF option may error out.
+  ON,   ///< Enforce POSIX I/O.
+  AUTO,  ///< Try cuFile I/O first, and fall back to POSIX I/O if the system requirements for cuFile
+         ///< are not met.
+};
+
+namespace detail {
+/**
+ * @brief Parse a string into a CompatMode enum.
+ *
+ * @param compat_mode_str Compatibility mode in string format (case-insensitive). Valid values
+ * include:
+ *   - `ON` (alias: `TRUE`, `YES`, `1`)
+ *   - `OFF` (alias: `FALSE`, `NO`, `0`)
+ *   - `AUTO`
+ * @return A CompatMode enum.
+ */
+CompatMode parse_compat_mode_str(std::string_view compat_mode_str);
+
+}  // namespace detail
+
+/**
+ * @brief
+ *
+ */
+class CompatModeManager {
+ private:
+  CompatMode _compat_mode_requested{CompatMode::AUTO};
+  bool _is_compat_mode_preferred{true};
+  bool _is_compat_mode_preferred_for_async{true};
+
+ public:
+  CompatModeManager()                                    = default;
+  ~CompatModeManager() noexcept                          = default;
+  CompatModeManager(const CompatModeManager&)            = delete;
+  CompatModeManager& operator=(const CompatModeManager&) = delete;
+  CompatModeManager(CompatModeManager&&) noexcept;
+  CompatModeManager& operator=(CompatModeManager&&) noexcept;
+
+  void compat_mode_reset(CompatMode compat_mode_requested);
+
+  CompatMode infer_compat_mode_if_auto(CompatMode compat_mode) noexcept;
+
+  bool is_compat_mode_preferred(CompatMode compat_mode) noexcept;
+
+  bool is_compat_mode_preferred() const noexcept;
+
+  bool is_compat_mode_preferred_for_async() const noexcept;
+
+  CompatMode compat_mode_requested() const noexcept;
+
+  std::tuple<FileWrapper, FileWrapper, CUFileHandleWrapper> resolve_compat_mode_for_file(
+    std::string const& file_path, std::string const& flags, mode_t mode, CompatMode compat_mode);
+
+  /**
+   * @brief Determine if the asynchronous I/O should be performed or not (throw exceptions)
+   * according to `_compat_mode_requested`, `_is_compat_mode_preferred`, and
+   * `_is_compat_mode_preferred_for_async`.
+   */
+  void validate_compat_mode_for_async();
+};
+
+}  // namespace kvikio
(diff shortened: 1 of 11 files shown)


madsbk (commented):
I think this is a good idea but I am wondering if we should go a step further and implement a `ResolveCompatMode` class that encapsulate ALL compat-mode logic?

kingcrimsontianyu:
I image that `ResolveCompatMode` would take over `FileHandle`'s constructor's responsibility: (1) to determine the values of `_is_compat_mode_preferred` and `_is_compat_mode_preferred_for_async`, and (2) to potentially throw exceptions if the cuFile path is not possible under the condition `_compat_mode_requested==OFF`. However, the side effect, i.e. returning the file descriptors, seems to be tightly coupled with these steps. I'm not quite sure how to separate them out to let `ResolveCompatMode` solely focus on the compat-mode logic. Otherwise, it feels the code is simply moved from the constructor to a separate function. Please let me know what you think of this!

madsbk:
Right, `ResolveCompatMode` would need something like a `.set_cufile_init_failure()` method.

I am not sure, it just think it would simplify the code to have all the compat infer code in one place, or maybe not :)

kingcrimsontianyu:
I see. We may add this in a future PR before the implementation details become more clear.
Meanwhile, I'm increasing the scope of this PR by addressing https://github.com/rapidsai/kvikio/issues/607. The FileHandle constructor is now slightly leaner.

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Enable documentation of the enum.
+/**
+ * @file
+ */
```
bdice, line 20: Should `#pragma once` and the includes go before this documentation block? Does this documentation block need to have any content?

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include "kvikio/file_utils.hpp"
+
+namespace kvikio {
+/**
+ * @brief I/O compatibility mode.
+ */
+enum class CompatMode : uint8_t {
+  OFF,  ///< Enforce cuFile I/O. GDS will be activated if the system requirements for cuFile are met
+        ///< and cuFile is properly configured. However, if the system is not suited for cuFile, I/O
+        ///< operations under the OFF option may error out, crash or hang.
```
bdice, line 35: Errors are acceptable. Is there a way to prevent crashes and hangs? It seems like we should be able to error in any situation where we would have otherwise fallen back.

cpp/include/kvikio/file_handle.hpp
```diff
@@ -28,11 +28,13 @@
 #include <kvikio/cufile/config.hpp>
 #include <kvikio/defaults.hpp>
 #include <kvikio/error.hpp>
+#include <kvikio/file_utils.hpp>
 #include <kvikio/parallel_operation.hpp>
 #include <kvikio/posix_io.hpp>
 #include <kvikio/shim/cufile.hpp>
 #include <kvikio/stream.hpp>
 #include <kvikio/utils.hpp>
+#include "kvikio/shim/cufile_h_wrapper.hpp"
```
bdice, line 37:
```suggestion
#include <kvikio/shim/cufile_h_wrapper.hpp>
```

cpp/include/kvikio/file_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <optional>
+#include <string>
+
+#include "kvikio/shim/cufile_h_wrapper.hpp"
```
bdice, line 21:
```suggestion
#include <kvikio/shim/cufile_h_wrapper.hpp>
```

cpp/src/file_handle.cpp
```diff
@@ -21,166 +21,51 @@
 #include <cstddef>
 #include <cstdlib>
 #include <stdexcept>
-#include <system_error>
+#include <utility>
 
 #include <kvikio/defaults.hpp>
 #include <kvikio/file_handle.hpp>
+#include <kvikio/file_utils.hpp>
+#include "kvikio/compat_mode.hpp"
```
bdice, line 29:
```suggestion
#include <kvikio/compat_mode.hpp>
```

cpp/include/kvikio/compat_mode.hpp
kingcrimsontianyu: I think the order of this command and `#pragma once` does not matter here. The block doesn't need any content. The presence of `@file` alone enables the `enum` doc generation.

But on a second thought, perhaps a better option is to simply document the namespace kvikio in the `defaults.hpp`. This would automatically generate docs for all namespace scope entities. What do you think @bdice ?

cpp/include/kvikio/compat_mode.hpp
kingcrimsontianyu: Thanks for noticing this. I meant to say "unspecified behavior" and wanted to give a couple of examples, but it looks like returning the non-success error code is my general observation. So I'll remove the "crash or hang" part from here and the doc.

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,78 @@
...
+  ON,   ///< Enforce POSIX I/O.
+  AUTO,  ///< Try cuFile I/O first, and fall back to POSIX I/O if the system requirements for cuFile
+         ///< are not met.
+};
+
+namespace detail {
+/**
+ * @brief Parse a string into a CompatMode enum.
+ *
+ * @param compat_mode_str Compatibility mode in string format(case-insensitive). Valid values
```
madsbk, line 45:
```suggestion
 * @param compat_mode_str Compatibility mode in string format (case-insensitive). Valid values
```

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * include:
+ *   - `ON` (alias: `TRUE`, `YES`, `1`)
+ *   - `OFF` (alias: `FALSE`, `NO`, `0`)
+ *   - `AUTO`
+ * @return A CompatMode enum.
+ */
+CompatMode parse_compat_mode_str(std::string_view compat_mode_str);
+
+}  // namespace detail
+
```
madsbk, line 55: Docs of `CompatModeManager` would be good

cpp/include/kvikio/defaults.hpp
```diff
@@ -93,6 +67,8 @@ class defaults {
   std::size_t _gds_threshold;
   std::size_t _bounce_buffer_size;
 
+  inline static CompatModeManager _compat_mode_manager{};
```
madsbk, line 70:
```suggestion
  static CompatModeManager _compat_mode_manager{};
```

cpp/include/kvikio/file_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace kvikio {
```
madsbk, line 23: Do we want this under `kvikio:detail`?

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,78 @@
...
+class CompatModeManager {
+ public:
+  CompatMode infer_compat_mode_if_auto(CompatMode compat_mode) noexcept;
+
+  bool is_compat_mode_preferred(CompatMode compat_mode) noexcept;
+
+  std::tuple<FileWrapper, FileWrapper, CUFileHandleWrapper, bool, bool>
+  resolve_compat_mode_for_file(std::string const& file_path,
```
madsbk, line 63: Docs of `resolve_compat_mode_for_file` would be good

cpp/src/file_handle.cpp
```diff
@@ -21,166 +21,51 @@
...
-  if (_compat_mode == CompatMode::AUTO && error_code.err != CU_FILE_SUCCESS) {
-    _compat_mode = CompatMode::ON;
-  } else {
-    CUFILE_TRY(error_code);
-  }
+  std::tie(_fd_direct_off,
+           _fd_direct_on,
+           _handle,
+           _is_compat_mode_preferred,
+           _is_compat_mode_preferred_for_async) =
```
madsbk, line 43: What about including them in the state of `CompatModeManager` ?
And replace  `compat_mode_requested`, `is_compat_mode_preferred`, `is_compat_mode_preferred_for_async`. with a const reference to a `CompatModeManager` instance?

cpp/include/kvikio/compat_mode.hpp
bdice: I think we try to keep `#pragma once` at the top of the file. Your proposal to automatically generate docs sounds reasonable on the surface but I am not familiar enough with how Doxygen works to have strong opinions here.

cpp/include/kvikio/compat_mode.hpp
kingcrimsontianyu: Completed.

cpp/include/kvikio/file_handle.hpp
kingcrimsontianyu: Thanks. Done.

cpp/src/file_handle.cpp
kingcrimsontianyu: Done.

cpp/include/kvikio/compat_mode.hpp
kingcrimsontianyu: Thanks. Done.

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: The `inline` specifier allows in-class initialization of the static `_compat_mode_manager` with an initializer `{}`. But since now `CompatModeManager` is stateful and is a non-static data member, both `inline` and `static` are scrapped.

cpp/include/kvikio/error.hpp
```diff
@@ -71,7 +72,7 @@ void cuda_driver_try_2(CUresult error, int line_number, char const* filename)
 {
   if (error == CUDA_ERROR_STUB_LIBRARY) {
     throw Exception{std::string{"CUDA error at: "} + std::string(filename) + ":" +
-                    KVIKIO_STRINGIFY(line_number) +
+                    std::to_string(line_number) +
```
kingcrimsontianyu, line 75: Just noticed a bug introduced here (https://github.com/rapidsai/kvikio/pull/581/files#diff-c7587291fbcd959f8bc842352115fb8ad2ecacdcc888a5dad99b3a0e2da6c4a3L53). `KVIKIO_STRINGIFY` applied to the variable `line_number` would expand to the string literal `"line_number"`. Fixed. @madsbk

cpp/include/kvikio/compat_mode.hpp
kingcrimsontianyu: Sure! Will do!

cpp/src/file_handle.cpp
kingcrimsontianyu: To clarify, my current implementation is to make the `defaults` singleton hold a `CompatModeManager` as its data member, and likewise make each `FileHandle` hold a per-instance `CompatModeManager`.

cpp/src/compat_mode.cpp
```diff
@@ -0,0 +1,171 @@
...
+  return _is_compat_mode_preferred_for_async;
+}
+
+CompatMode CompatModeManager::compat_mode_requested() const noexcept
+{
+  return _compat_mode_requested;
+}
+
+std::tuple<FileWrapper, FileWrapper, CUFileHandleWrapper>
+CompatModeManager::resolve_compat_mode_for_file(std::string const& file_path,
```
madsbk, line 107: Could this be moved to the ctor? I think it is a bit surprising that it changes the state of `CompatModeManager`

madsbk (commented):
@kingcrimsontianyu If it is not too much work, I think it would help review and design discussion if you move `FileWrapper` to its own PR?

kingcrimsontianyu:
Reducing the scope of this PR as suggested. File RAII addressed in another PR: https://github.com/rapidsai/kvikio/pull/614

cpp/src/compat_mode.cpp
kingcrimsontianyu: My initial intention is to default initialize `CompatModeManager`, and derive the data member at a later time, i.e:

- For the compat mode manager in `defaults` class that does not involve files, in `defaults` 's constructor use `compat_mode_reset()` to save the requested compat mode and determine the final compat mode.
- For the compat mode manager in `FileHandle` class, use `resolve_compat_mode_for_file` to save the requested compat mode and determine the final compat mode for sync and async I/O (this function also returns file objects that the constructor cannot do).

Would it be better if we:
- Still use `CompatModeManager` for the `defaults` class, but remove any async-related data member and method.
- Have a subclass `CompatModeForFileManager` to include the async-related data member and method, put the "resolve_compat_mode_for_file" logic to the constructor as you suggested, and have a getter method to retrieve the generated file and handle objects?

kingcrimsontianyu:
If I directly rebase this branch on 25.04, there would too many repeated rebase conflicts to resolve (as a result of some commits on this branch making back-and-forth changes).

But I do have a branch where I simply squash the commits of this branch and then rebase to 25.04. Should I force-push that branch to here? Or should I close this PR and open a new one for the squashed-commit branch? I asked with the hope to not lose the review comments above. Thanks for any suggestion! @bdice

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,93 @@
...
+  bool _is_compat_mode_preferred{true};
+  bool _is_compat_mode_preferred_for_async{true};
+
+ public:
+  CompatModeManager()                                    = default;
+  ~CompatModeManager() noexcept                          = default;
+  CompatModeManager(const CompatModeManager&)            = delete;
+  CompatModeManager& operator=(const CompatModeManager&) = delete;
+  CompatModeManager(CompatModeManager&&) noexcept;
+  CompatModeManager& operator=(CompatModeManager&&) noexcept;
```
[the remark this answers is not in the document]
kingcrimsontianyu: Should we still keep the move constructor and move assignment operator explicitly defined? If made `default`, the moved-from object will not have proper "reset" values. The `std::exchange()` used in the current explicitly defined version avoids this problem.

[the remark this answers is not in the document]
madsbk: I don't think we need to reset the values? I mean, if `CompatModeManager` is accessed after move, I don't think resetting its values helps?

cpp/src/compat_mode.cpp
madsbk: What about keeping `defaults.cpp` as it was prior to this PR and then introduce `CompatModeManager` for file handlers?
I don't think the current `defaults.cpp` code it too complex. The complex starts when files are involved :)

madsbk:
> If I directly rebase this branch on 25.04, there would too many repeated rebase conflicts to resolve (as a result of some commits on this branch making back-and-forth changes).
> 
> But I do have a branch where I simply squash the commits of this branch and then rebase to 25.04. Should I force-push that branch to here? Or should I close this PR and open a new one for the squashed-commit branch? I asked with the hope to not lose the review comments above. Thanks for any suggestion! @bdice

Squash and force-push is fine.

cpp/include/kvikio/compat_mode.hpp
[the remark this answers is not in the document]
kingcrimsontianyu: What I was thinking of is something like this:
```cpp
// std::exchange currently used in FileHandle's move constructor "resets" old_file_handle's fd to -1.
auto new_file_handle = std::move(old_file_handle);

// So this is valid, as fd has a specified value.
query(old_file_handle.fd()); 

// But small whoopsy. CompatModeManager is default moved. 
// After moved from, all the data members of `old_file_handle._compat_mode_manager` are left unspecified.
query(old_file_handle.compat_mode_requested()); 
query(old_file_handle.is_compat_mode_preferred()); 
query(old_file_handle.is_compat_mode_preferred_for_async()); 
```
Maybe I worried too much about this. :smile:

cpp/src/compat_mode.cpp
kingcrimsontianyu: Reverting `defaults.cpp` will do. 

The flip side would be a slight impediment to mocking. I had intended to thoroughly mock the following code, including the `is_cufile_available()` function:

https://github.com/rapidsai/kvikio/blob/b5758ec8483682cb7e50e4b7a4fb3b139b368534/cpp/src/defaults.cpp#L146

https://github.com/rapidsai/kvikio/blob/b5758ec8483682cb7e50e4b7a4fb3b139b368534/cpp/src/defaults.cpp#L157

A compromise  I could imagine would be to simply have these constructor overloads:
```cpp
// For defaults
// File-related members are default initialized. No actual files or handles are involved.
CompatModeManager();

// For FileHandle
CompatModeManager(std::string const& file_path, std::string const& flags, mode_t mode, CompatMode compat_mode_requested);
```
That is to tolerate the dual use of `CompatModeManager` for `defaults` and `FileHandle`, with the slight benefit of thorough mocking.

cpp/include/kvikio/compat_mode.hpp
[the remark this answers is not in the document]
madsbk: I think that is a bit too defensive :)

cpp/src/compat_mode.cpp
madsbk: That could also work, but I must admit, I prefer avoiding the dual use of `CompatModeManager` altogether.

kingcrimsontianyu: Got it! I'll restrict `CompatModeManager` to file only and revert `defaults`. 
To facilitate mocking in another PR I'll also keep `infer_compat_mode_if_auto` and `is_compat_mode_preferred` in `CompatModeManager` (mild code duplication).

cpp/include/kvikio/compat_mode.hpp
kingcrimsontianyu: Done.

cpp/include/kvikio/compat_mode.hpp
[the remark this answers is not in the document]
kingcrimsontianyu: Done.

cpp/src/compat_mode.cpp
kingcrimsontianyu: Completed.

cpp/include/kvikio/file_handle.hpp
```diff
@@ -465,6 +457,13 @@ class FileHandle {
    * @return Boolean answer.
    */
   [[nodiscard]] bool is_compat_mode_preferred_for_async() const noexcept;
+
+  /**
+   * @brief Returns the initially requested compatibility mode.
+   *
+   * @return The compatibility mode initially requested.
+   */
+  CompatMode compat_mode_requested() const noexcept;
```
madsbk, line 466: Maybe return a const reference to `_compat_mode_manager` instead of  `compat_mode_requested`, `is_compat_mode_preferred`, and `is_compat_mode_preferred_for_async`. Most users will never have to use them, so I think an extra indirection is fine.

madsbk (commented):
Looks great, I only have an one suggestion

cpp/include/kvikio/file_handle.hpp
kingcrimsontianyu: I agree. Done!
Updated the PR to reflect this slightly breaking change.

madsbk (approved):
Looks great

cpp/src/compat_mode.cpp
```diff
@@ -0,0 +1,146 @@
...
+  } else {
+    CUFILE_TRY(error_code.value());
+  }
+
+  // Check cuFile async API
+  static bool is_extra_symbol_available = is_stream_api_available();
+  static bool is_config_path_empty      = config_path().empty();
+  _is_compat_mode_preferred_for_async =
+    _is_compat_mode_preferred || !is_extra_symbol_available || is_config_path_empty;
+  return;
```
wence-, line 128: *nit*: This part means this object isn't thread safe. Is that problematic?

wence- (commented):
Some small questions/nits

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,150 @@
...
+  AUTO,  ///< Try cuFile I/O first, and fall back to POSIX I/O if the system requirements for cuFile
+         ///< are not met.
+};
+
+namespace detail {
+/**
+ * @brief Parse a string into a CompatMode enum.
+ *
+ * @param compat_mode_str Compatibility mode in string format (case-insensitive). Valid values
+ * include:
```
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks. Done.

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,150 @@
...
+   * - Infers the final compatibility modes for synchronous and asynchronous I/O paths,
+   * respectively.
+   * - Initializes the file wrappers and cuFile handle associated with a FileHandle.
+   *
+   * @param file_path Refer to
+   * FileHandle::FileHandle(std::string const&, std::string const&, mode_t, CompatMode).
+   * @param flags Same as above.
+   * @param mode Same as above.
+   * @param compat_mode_requested Same as above.
+   * @param file_handle Point to the FileHandle object that owns this compatibility mode manager.
```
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks. Done.

cpp/include/kvikio/compat_mode.hpp
```diff
@@ -0,0 +1,150 @@
...
+  CompatMode compat_mode_requested() const noexcept;
+
+  /**
+   * @brief Determine if the asynchronous I/O can be performed or not (throw exceptions)
+   * according to the existing compatibility mode data in the manager.
+   *
+   * The asynchronous I/O cannot be performed, for instance, when compat_mode_requested() is
+   * CompatMode::OFF, is_compat_mode_preferred() is CompatMode::OFF, but
+   * is_compat_mode_preferred_for_async() is CompatMode::ON (due to missing cuFile stream API or
+   * cuFile configuration file).
```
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks. Done as suggested.

cpp/src/compat_mode.cpp
```diff
@@ -0,0 +1,146 @@
...
+  if (tmp == "on" || tmp == "true" || tmp == "yes" || tmp == "1") {
+    res = CompatMode::ON;
+  } else if (tmp == "off" || tmp == "false" || tmp == "no" || tmp == "0") {
+    res = CompatMode::OFF;
+  } else if (tmp == "auto") {
+    res = CompatMode::AUTO;
+  } else {
+    throw std::invalid_argument("Unknown compatibility mode: " + std::string{tmp});
+  }
+  return res;
```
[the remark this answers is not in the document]
kingcrimsontianyu: I think it's a matter of either using the early return style or using the single point of return style. In this case return within the branches does look more concise. Done as suggested!

cpp/src/compat_mode.cpp
kingcrimsontianyu: Could you elaborate a bit? I'm not quite sure which case could potentially cause a data race here.
PS: Removed the superfluous `return`.

kingcrimsontianyu: PPS: Added `const` to the function-local static variables.

wence-: I thought the initialisation of the function-local statics might be racy if multiple threads are here simultaneously, however https://eel.is/c++draft/stmt.dcl#3 indicates that I am incorrect. effectively static initialisation is wrapped in a "critical" section by the compiler. (assuming we are compiling without `-fno-threadsafe-statics`).

So nothing to do here, thanks.

wence- (approved):
Thanks @kingcrimsontianyu

PointKernel:
For your reference, this PR seems to break the cudf CI: https://github.com/rapidsai/cudf/actions/runs/13376114806/job/37355755788?pr=18013

kingcrimsontianyu:
@PointKernel Thanks Yunsong! I'm making a cuDF PR for the fix.

kingcrimsontianyu:
FYI: The cuDF fix is here https://github.com/rapidsai/cudf/pull/18025

PR #614, opened 2025-02-04
Add RAII file wrappers to avoid resource leak

### LINKED ISSUE #607: FileHandle file resource may leak if the constructor throws an exception

kingcrimsontianyu:
Closes #607

### DIFF
--- a/cpp/src/file_utils.cpp
+++ b/cpp/src/file_utils.cpp
@@ -52,7 +52,7 @@ void FileWrapper::open(std::string const& file_path,
   if (!opened()) { _fd = open_fd(file_path, flags, o_direct, mode); }
 }
 
-bool FileWrapper::opened() noexcept { return _fd != -1; }
+bool FileWrapper::opened() const noexcept { return _fd != -1; }
 
 void FileWrapper::close() noexcept
 {
--- a/cpp/src/file_utils.cpp
+++ b/cpp/src/file_utils.cpp
@@ -21,6 +21,7 @@
 #include <system_error>
 #include <utility>
 
+#include <kvikio/error.hpp>
 #include <kvikio/file_utils.hpp>
 #include <kvikio/shim/cufile.hpp>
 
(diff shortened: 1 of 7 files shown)


KyleFromNVIDIA (approved):
Approved trivial CMake changes

madsbk (approved):
Looks great, thanks!

cpp/src/file_utils.cpp
```diff
@@ -0,0 +1,163 @@
...
+}
+
+bool CUFileHandleWrapper::registered() const noexcept { return _registered; }
+
+CUfileHandle_t CUFileHandleWrapper::handle() const noexcept { return _handle; }
+
+void CUFileHandleWrapper::unregister_handle() noexcept
+{
+  if (registered()) {
+    cuFileAPI::instance().HandleDeregister(_handle);
```
kingcrimsontianyu, line 103: Nit: Non-success error codes from the system call `close()` and `cuFileHandleDeregister()` are currently left unhandled in order to keep the destructor `noexcept`. Maybe in the future we could consider having some logging mechanism or having some global error states for these rare but not impossible cases? Or simply removing the `noexcept` promise from the destructor? @madsbk

madsbk: Let's keep the dtor `noexcept`, but maybe print to stderr?

kingcrimsontianyu: Done. Added to `error.hpp/cpp` . It looks like `cuFileDeregister()` returns void, so this rare error logging is only done for `::close()` atm.

madsbk (approved):

madsbk:
@kingcrimsontianyu, please feel free to merge when ready.

vuule (approved):

PR #616, opened 2025-02-05
Fix package versioning

vyasr:
The nightly package versions are currently not being correctly assigned.

PR #622, opened 2025-02-05
Use `rapids-generate-version` for env var

jakirkham:
Follow up on this thread: https://github.com/rapidsai/kvikio/pull/616#discussion_r1942351193

### DIFF
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -22,7 +22,7 @@ conda config --set path_conflict prevent
 
 sccache --zero-stats
 
-RAPIDS_PACKAGE_VERSION=$(head -1 ./VERSION) rapids-conda-retry mambabuild \
+RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry mambabuild \
   --channel "${CPP_CHANNEL}" \
   conda/recipes/kvikio
 


ci/build_python.sh
```diff
@@ -22,7 +22,7 @@ conda config --set path_conflict prevent
 
 sccache --zero-stats
 
-RAPIDS_PACKAGE_VERSION=$(head -1 ./VERSION) rapids-conda-retry mambabuild \
+RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry mambabuild \
```
vyasr, line 25: Note that the function is being called once above, hence why I previously just read the output.

jakirkham: Here is the line Vyas referenced

https://github.com/rapidsai/kvikio/blob/b00141379b8ec777b0f3081069986c70552f880e/ci/build_python.sh#L16

And we need to that line as the `VERSION` file has to be written out there

madsbk: I think this is slightly preferably. I misunderstood the original code, I though that `./VERSION` was written just to be read here :)

madsbk (approved):

ci/build_python.sh
jakirkham: Added a comment to that line. Also shared a few examples below

madsbk:
Thanks @jakirkham

PR #628, opened 2025-02-07
Use rapids-generate-version for package versions

bdice:
This uses the `RAPIDS_PACKAGE_VERSION` values set in #616. This ensures we have consistent nightly versions.

This PR is independent of #622 (it is needed regardless of whether that PR is closed or merged).


jameslamb (approved):
agree!

vyasr:
Wow sorry I missed that. Thanks.

jakirkham:
Good catch! Thanks Bradley 🙏

PR #635, opened 2025-02-13
Require CMake 3.30.4

robertmaynard:
## Description

Update CMake minimum required to 3.30.4 across all of RAPIDS

### DIFF
--- a/java/src/main/native/CMakeLists.txt
+++ b/java/src/main/native/CMakeLists.txt
@@ -16,7 +16,7 @@ cmake_minimum_required(VERSION 3.30.4 FATAL_ERROR)
 
 project(CuFileJNI LANGUAGES CXX CUDA)
 
-find_package(CUDA REQUIRED)
+find_package(CUDAToolkit REQUIRED)
 
 find_package(JNI REQUIRED)
 


jameslamb (approved):
Approving based on my understanding of https://github.com/rapidsai/rapids-cmake/pull/753

It would be helpful to have a tracking issue similar to https://github.com/rapidsai/build-planning/issues/80 linking all these PRs together and with documentation for why we're doing this. Or https://github.com/rapidsai/rapids-cmake/issues/680 could even become that.

robertmaynard:
I have made https://github.com/rapidsai/build-planning/issues/154 to track RAPIDS wide status

bdice:
@robertmaynard The newest CMake Maven plugin I see is version `3.30.2-b1`. https://repo.maven.apache.org/maven2/com/googlecode/cmake-maven-project/cmake-maven-plugin/

I'm not sure if we can update to 3.30.4 until that is updated.

bdice:
I filed an issue on the upstream repository (since Google Code no longer exists, this appears to be the current source): https://github.com/cmake-maven-project/cmake-maven-project/issues/68

robertmaynard:
> @robertmaynard The newest CMake Maven plugin I see is version `3.30.2-b1`. https://repo.maven.apache.org/maven2/com/googlecode/cmake-maven-project/cmake-maven-plugin/
> 
> I'm not sure if we can update to 3.30.4 until that is updated.

I don't think that will block us. It supports using an local install of CMake ( https://github.com/cmake-maven-project/cmake-maven-project/blob/master/README.md#using-a-local-cmake-installation ) and not downloading the tightly coupled version.

PR #630, opened 2025-02-09
Enable tracing of thread pool tasks using NVTX

kingcrimsontianyu:
This PR implements the basic feature outlined in #631. 
The two good-to-haves are currently blocked.

PR #637, opened 2025-02-18
Minor fixes on the file permission, host allocation flag, and worker thread initialization

kingcrimsontianyu:
This PR makes the following minor fixes:
- Use the correct file permission flags corresponding to the `644` code.
- Use the correct flag for the `cuMemHostAlloc` call.
- For the thread pool, replace the `thread_local` call-once section (which may negatively affect performance; see https://github.com/rapidsai/kvikio/pull/630#discussion_r1960285383) with more idiomatic worker thread initialization function.


cpp/src/bounce_buffer.cpp
```diff
@@ -74,7 +74,8 @@ AllocRetain::Alloc AllocRetain::get()
   // If no available allocation, allocate and register a new one
   void* alloc{};
   // Allocate page-locked host memory
-  CUDA_DRIVER_TRY(cudaAPI::instance().MemHostAlloc(&alloc, _size, CU_MEMHOSTREGISTER_PORTABLE));
+  // Under unified addressing, host memory allocated this way is automatically portable and mapped.
+  CUDA_DRIVER_TRY(cudaAPI::instance().MemHostAlloc(&alloc, _size, CU_MEMHOSTALLOC_PORTABLE));
```
vuule, line 78: oh, this was using a macro from a different CUDA API?

kingcrimsontianyu: Yes. Fortunately they are both defined as `0x1`, so we haven't run into any issues.

madsbk: Good catch!

madsbk (approved):

cpp/include/kvikio/threadpool_wrapper.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <functional>
+
+#include <BS_thread_pool.hpp>
+
+#include <kvikio/nvtx.hpp>
+
+namespace kvikio {
+
+template <typename pool_type>
+class thread_pool_wrapper : public pool_type {
```
kingcrimsontianyu, line 28: Created a thread pool wrapper to avoid burdening the `defaults` class with the detail of `worker_thread_init_func`.

cpp/include/kvikio/threadpool_wrapper.hpp
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * @brief Reset the number of threads in the thread pool, and invoke a pre-defined initialization
+   * function in each worker thread.
+   *
+   * @param nthreads The number of threads to use.
+   */
+  void reset(unsigned int nthreads) { pool_type::reset(nthreads, worker_thread_init_func); }
+
+ private:
+  inline static std::function<void()> worker_thread_init_func{[] {
```
kingcrimsontianyu, line 47: `inline` here permits in-class definition of the static data member with a brace initializer. Ref https://eel.is/c++draft/class.static#data-4

madsbk (approved):
Nice!

kingcrimsontianyu:
Just fixed a style consistency issue (`nvtx_manager` should use CamelCase). Will merge once CI passes. Thanks for the review!

PR #643, opened 2025-02-24
Added timeout to kvikio requests

### LINKED ISSUE #613: Set timeouts for HTTP requests
xref https://github.com/rapidsai/kvikio/pull/603#issuecomment-2634061687

Investigate whether it's possible for the current remote IO implementation to hang indefinitely. https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html describes the option we probably want. We can add that as an option in `libcurl.cpp` next to the other CURLOPTs.

We'd need to figure out a good default value for this parameter and a way to configure it.

There's an interaction with #603: timeout errors raised by libcurl is one of the non-completed HTTP request / response errors that should be retried automatically (along with things like closed sockets).

TomAugspurger:
This PR adds a timeout configuration value, which controls [CURLOPT_TIMEOUT](https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html) on the remote IO requests we make with libcurl.

The default timeout is 60 seconds. Timeouts are considered retryable errors, so the retry logic has been updated to handle that in addition to HTTP status codes.

Closes #613.


madsbk:
I guess it should follow the style of https://github.com/rapidsai/kvikio/pull/644 ?
cc. @kingcrimsontianyu

TomAugspurger:
Updated to use the new config system.

cpp/src/defaults.cpp
```diff
@@ -129,6 +129,16 @@ defaults::defaults()
     }
     _http_max_attempts = env;
   }
+
+  // Determine the default value of `http_timeout`
+  {
+    const long env = getenv_or("KVIKIO_HTTP_TIMEOUT", 60);
```
madsbk, line 135:
```suggestion
    long const env = getenv_or("KVIKIO_HTTP_TIMEOUT", 60);
```

madsbk (approved):

Issue #642, opened 2025-02-23
Confusion around setting config values


TomAugspurger:
In https://github.com/rapidsai/kvikio/pull/603#discussion_r1966604136, @bdice raised an issue that I also ran into with the `kvikio.defaults` system. We have three methods per option, typically:

- `kvikio.defaults.<option>`
- `kvikio.defaults.set_<option>`
- `kvikio.defaults.<option>_reset`

It wasn't initially clear to me which of `set_<option>` and `<option>_reset` is method that immediately sets the value, and which is the one that must be used as a context manager. And I still mix it up occasionally.

I like the way [donfig](https://donfig.readthedocs.io/en/latest/configuration.html#directly-within-python), which was based on the Dask config system, combines that into a single object that handles both immediately setting the value and is usable as a context manager.

```python
>>> kvikio.defaults.set({option: value})  # set it permenantly

>>> with kvikio.defaults.set({option: value}): ...  # set it for the lifetime of the context manager
```

bdice:
Yes, this API feels more Pythonic to me. I would support refactoring in this direction.

madsbk:
Agree, looks like a good idea!

I would be good to include config of cuFile's [`DriverProperties`](https://github.com/rapidsai/kvikio/blob/branch-25.04/python/kvikio/kvikio/_lib/cufile_driver.pyx#L47), which is currently just exposed as a class: https://github.com/rapidsai/kvikio/blob/branch-25.04/python/kvikio/kvikio/cufile_driver.py#L11

kingcrimsontianyu:
I think `DriverProperties` has already been made pretty Pythonic with the `@property` decorator. Is this what you have in mind as further improvement? @madsbk 
```python
with kvikio.cufile_driver.set({"poll_mode": True, "max_device_cache_size": 4096})
```

madsbk:
Yes, I think it would be good to use the API style.

PR #644, opened 2025-02-24
Improve config handling in cpp and python

### LINKED ISSUE #526: Pythonic bindings to cuFIle's driver properties
I would be good to expose `kvikio.cufile_driver.DriverProperties` in a Pythonic way.

kingcrimsontianyu:
Closes https://github.com/rapidsai/kvikio/issues/642
Closes #526 

This PR makes the following changes:

## Improve config setter/getter

This PR lets the properties in `kvikio.defaults` and `kvikio.cufile_driver` be set using a friendly Pythonic interface (C++ interface has also been adjusted for improved clarity). The expression below can appear standalone which sets the properties globally, or in a `with` statement which sets the properties via a context manager.

- To set one or more properties

```python
# Set the property globally.
kvikio.defaults.set({"prop1": value1, "prop2": value2})
kvikio.cufile_driver.set({"prop1": value1, "prop2": value2})

# Set the property with a context manager.
# The property automatically reverts to its old value
# after leaving the `with` block.
with kvikio.defaults.set({"prop1": value1, "prop2": value2}):
with kvikio.cufile_driver.set({"prop1": value1, "prop2": value2}):
    ...
```

- To set a single property

```python
# Set the property globally.
kvikio.defaults.set("prop", value)
kvikio.cufile_driver.set("prop", value)

# Set the property with a context manager.
# The property automatically reverts to its old value
# after leaving the `with` block.
with kvikio.defaults.set("prop", value):
with kvikio.cufile_driver.set("prop", value):
    ...
```

In addition, this PR supports the getter method:
```python
p = kvikio.defaults.get("prop",)
p = kvikio.cufile_driver.get("prop",)
```

## Deprecate the old ways of setting configs

The `<prop>_reset` and `set_<prop>` functions are now deprecated. Deprecation warnings (of type `FutureWarning`) are issued at runtime.

## Add Python function is_compat_mode_preferred

The C++ function `kvikio::defaults::is_compat_mode_preferred()` resolves and queries the current global compat mode. This PR introduces this function to Python.

### DIFF
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -157,7 +157,7 @@ class defaults {
    *
    * Notice, it is not possible to change the default thread pool. KvikIO will
    * always use the same thread pool however it is possible to change number of
-   * threads in the pool (see `kvikio::default::set_thread_pool_nthreads`).
+   * threads in the pool (see `kvikio::default::set_thread_pool_nthreads()`).
    *
    * @return The default thread pool instance.
    */
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -166,7 +166,7 @@ class defaults {
   /**
    * @brief Get the number of threads in the default thread pool.
    *
-   * Set the default value using `kvikio::default::set_thread_pool_nthreads` or by
+   * Set the default value using `kvikio::default::set_thread_pool_nthreads()` or by
    * setting the `KVIKIO_NTHREADS` environment variable. If not set, the default value is 1.
    *
    * @return The number of threads.
@@ -187,7 +187,7 @@ class defaults {
   /**
    * @brief Get the default task size used for parallel IO operations.
    *
-   * Set the default value using `kvikio::default::set_task_size` or by setting
+   * Set the default value using `kvikio::default::set_task_size()` or by setting
    * the `KVIKIO_TASK_SIZE` environment variable. If not set, the default value is 4 MiB.
    *
    * @return The default task size in bytes.
@@ -207,7 +207,7 @@ class defaults {
    * In order to improve performance of small IO, `.pread()` and `.pwrite()` implement a shortcut
    * that circumvent the threadpool and use the POSIX backend directly.
    *
-   * Set the default value using `kvikio::default::set_gds_threshold` or by setting the
+   * Set the default value using `kvikio::default::set_gds_threshold()` or by setting the
    * `KVIKIO_GDS_THRESHOLD` environment variable. If not set, the default value is 1 MiB.
    *
    * @return The default GDS threshold size in bytes.
@@ -258,7 +258,7 @@ class defaults {
   /**
    * @brief The list of HTTP status codes to retry.
    *
-   * Set the value using `kvikio::default::http_status_codes()` or by setting the
+   * Set the value using `kvikio::default::set_http_status_codes()` or by setting the
    * `KVIKIO_HTTP_STATUS_CODES` environment variable. If not set, the default value is
    *
    * - 429
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -177,8 +177,7 @@ class defaults {
    * @brief Set the number of threads in the default thread pool. Waits for all currently running
    * tasks to be completed, then destroys all threads in the pool and creates a new thread pool with
    * the new number of threads. Any tasks that were waiting in the queue before the pool was reset
-   * will then be executed by the new threads. If the pool was paused before resetting it, the new
-   * pool will be paused as well.
+   * will then be executed by the new threads.
    *
    * @param nthreads The number of threads to use.
    */
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -99,14 +99,14 @@ class defaults {
   [[nodiscard]] static CompatMode compat_mode();
 
   /**
-   * @brief Reset the value of `kvikio::defaults::compat_mode()`.
+   * @brief Set the value of `kvikio::defaults::compat_mode()`.
    *
    * Changing the compatibility mode affects all the new FileHandles whose `compat_mode` argument is
    * not explicitly set, but it never affects existing FileHandles.
    *
    * @param compat_mode Compatibility mode.
    */
-  static void compat_mode_reset(CompatMode compat_mode);
+  static void set_compat_mode(CompatMode compat_mode);
 
   /**
    * @brief Infer the `AUTO` compatibility mode from the system runtime.
@@ -157,7 +157,7 @@ class defaults {
    *
    * Notice, it is not possible to change the default thread pool. KvikIO will
    * always use the same thread pool however it is possible to change number of
-   * threads in the pool (see `kvikio::default::thread_pool_nthreads_reset()`).
+   * threads in the pool (see `kvikio::default::set_thread_pool_nthreads`).
    *
    * @return The default thread pool instance.
    */
@@ -166,7 +166,7 @@ class defaults {
   /**
    * @brief Get the number of threads in the default thread pool.
    *
-   * Set the default value using `kvikio::default::thread_pool_nthreads_reset()` or by
+   * Set the default value using `kvikio::default::set_thread_pool_nthreads` or by
    * setting the `KVIKIO_NTHREADS` environment variable. If not set, the default value is 1.
    *
    * @return The number of threads.
@@ -174,7 +174,7 @@ class defaults {
   [[nodiscard]] static unsigned int thread_pool_nthreads();
 
   /**
-   * @brief Reset the number of threads in the default thread pool. Waits for all currently running
+   * @brief Set the number of threads in the default thread pool. Waits for all currently running
    * tasks to be completed, then destroys all threads in the pool and creates a new thread pool with
    * the new number of threads. Any tasks that were waiting in the queue before the pool was reset
    * will then be executed by the new threads. If the pool was paused before resetting it, the new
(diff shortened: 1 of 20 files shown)


kingcrimsontianyu:
Forgot about cuFile's `DriverProperties`. Sending this PR to draft and working on that.

cpp/include/kvikio/defaults.hpp
```diff
@@ -166,81 +166,81 @@ class defaults {
...
    */
   [[nodiscard]] static unsigned int thread_pool_nthreads();
 
   /**
-   * @brief Reset the number of threads in the default thread pool. Waits for all currently running
+   * @brief Set the number of threads in the default thread pool. Waits for all currently running
    * tasks to be completed, then destroys all threads in the pool and creates a new thread pool with
    * the new number of threads. Any tasks that were waiting in the queue before the pool was reset
    * will then be executed by the new threads. If the pool was paused before resetting it, the new
    * pool will be paused as well.
```
madsbk, line 181:
```suggestion
   * will then be executed by the new threads.
```

python/kvikio/kvikio/defaults.py
```diff
@@ -2,11 +2,81 @@
...
+        return None
+
+    def __exit__(self, type, value, traceback):
+        for key, value in self._old_properties.items():
+            self._set_property(key, value)
+
+    def _get_property(self, property: str) -> Any:
+        if property == "num_threads":
+            property = "thread_pool_nthreads"
+        func = getattr(kvikio._lib.defaults, property)
```
TomAugspurger, line 28: What exception type do we want if a user gives a bad key name? Right now I think we'd get an `AttributeError`. Maybe a `KeyError` would be more appropriate.

python/kvikio/kvikio/defaults.py
```diff
@@ -2,11 +2,81 @@
...
+        return func()
+
+    def _set_property(self, property: str, value: Any):
+        if property == "num_threads":
+            property = "thread_pool_nthreads"
+        func = getattr(kvikio._lib.defaults, f"set_{property}")
+        func(value)
+
+
+def set(*config: Any) -> ConfigContextManager:
```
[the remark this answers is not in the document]
TomAugspurger: If we want to get fancy, we can have the `str` be a `typing.Literal` or `enum.Enum` class with the valid setting names.

python/kvikio/kvikio/defaults.py
```diff
@@ -32,45 +102,10 @@ def compat_mode() -> kvikio.CompatMode:
     return kvikio._lib.defaults.compat_mode()
 
 
-def compat_mode_reset(compatmode: kvikio.CompatMode) -> None:
```
TomAugspurger, line 35: Is it worth keeping the old way of doing things around for a bit, maybe with a `FutureWarning` warning that they'll be removed soon?

Since the `ConfigContext` handles both the `set` and `reset` modes, they can have the same implementation, which can defer to calling `ConfigContext`.

python/kvikio/kvikio/defaults.py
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks. I didn't know Python has the `@overload` decorator for type checking. Great to learn something new. Modified as suggested.

This has helped to unveil a type issue that fails the type checker: `set(*config)` indicates that `config` is a positional-only parameter, whereas the overloads mean the parameters are [positional-or-keywords](https://docs.python.org/3/glossary.html#term-parameter). I've added `/` behind the parameters in the overloaded functions to indicate positional-only.

python/kvikio/kvikio/defaults.py
kingcrimsontianyu: Thanks. That makes sense. Modified as suggested.

What I've done now is to scan the `kvikio._lib.defaults` modules (on the first use of the context manager; result gets cached) to get all the getter and setter names and their function objects, and put them in getter/setter dictionaries. Now the `KeyError` will be triggered for a bad key.

python/kvikio/kvikio/defaults.py
kingcrimsontianyu: I'll add back the old API with the deprecation notice. What I'm not sure of though, is whether the deprecation warnings will be treated as errors to fail the CI. Our C++ compiler is made behave this way, so the deprecation attribute cannot be applied... I'll test the Python part.

kingcrimsontianyu: CI has passed. So the change should be good.

python/kvikio/kvikio/defaults.py
kingcrimsontianyu: Simplified implementation. Done.

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -1,14 +1,116 @@
...
-from typing import Tuple
+from typing import Any, Tuple, overload
 
+import kvikio.utils
 from kvikio._lib import cufile_driver  # type: ignore
 
-# TODO: Wrap nicely, maybe as a dataclass?
-# <https://github.com/rapidsai/kvikio/issues/526>
-DriverProperties = cufile_driver.DriverProperties
+properties = cufile_driver.DriverProperties()
```
kingcrimsontianyu, line 10: I think having an instance of `DriverProperties()` would suffice. So now the settable properties in cufile driver can be set in the following ways:
```python
kvikio.cuda_driver.properties.<prop> = <value>
kvikio.cuda_driver.set("<prop>", <value>) # Can also be used in a "with" statement
kvikio.cuda_driver.set({"<prop>": <value>}) # Can also be used in a "with" statement
```
Let me know if this looks good enough to you @madsbk

madsbk (commented):
Don't we also want to use `get` for configs like `task_size`  ?

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -1,14 +1,116 @@
...
+
+
+class ConfigContextManager:
```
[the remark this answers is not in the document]
kingcrimsontianyu: Done. Also added the docstring to `defaults.ConfigContextManager`.

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -1,14 +1,116 @@
...
+      .. code-block:: python
+
+         kvikio.cufile_driver.properties.set("prop", value)
+
+    Parameters
+    ----------
+    config
+        The configurations. Can either be a single parameter (dict) consisting of one
+        or more properties, or two parameters key (string) and value (Any)
+        indicating a single property.
```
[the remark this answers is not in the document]
kingcrimsontianyu: Done. Also enumerated all valid values:
image

python/kvikio/kvikio/cufile_driver.py
```diff
@@ -1,14 +1,116 @@
...
+
+      .. code-block:: python
+
+         kvikio.cufile_driver.properties.set({"prop1": value1, "prop2": value2})
+
+    - To set a single property
+
+      .. code-block:: python
+
+         kvikio.cufile_driver.properties.set("prop", value)
```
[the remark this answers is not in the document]
kingcrimsontianyu: Done.

python/kvikio/kvikio/_lib/defaults.pyx
```diff
@@ -14,80 +14,86 @@ cdef extern from "<kvikio/defaults.hpp>" namespace "kvikio" nogil:
         OFF = 0
         ON = 1
         AUTO = 2
+    bool cpp_is_compat_mode_preferred \
```
kingcrimsontianyu, line 17: This additional C++ function is needed.

python/kvikio/tests/test_basic_io.py
```diff
@@ -100,17 +98,16 @@ def test_incorrect_open_mode_error(tmp_path, xp):
 
 
 @pytest.mark.skipif(
-    kvikio.defaults.compat_mode(),
```
kingcrimsontianyu, line 103: This is a minor bug. By default, compat mode is `AUTO`, not `ON`. Here we need to resolve it by using `is_compat_mode_preferred()` (which necessitates importing its C++ function as mentioned above).

docs/source/api.rst
```diff
@@ -11,6 +11,24 @@ CuFile
 .. autoclass:: IOFuture
     :members:
 
+CuFile driver
```
kingcrimsontianyu, line 14: Added this section, which is missing on the dev branch.

kingcrimsontianyu:
>Don't we also want to use get for configs like task_size ?

Done. Added the `get` method for both `kvikio.defaults` and `kvikio.cufile_driver`.

madsbk (approved):
Looks great!
Only question is if we should raise a [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) ?

TomAugspurger:
> Only question is if we should raise a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) ?

My preference is usually for a `FutureWarning` if we want people to actually see them (outside of `pytest` invoked unit tests, which un-hides DeprecationWarnings).

madsbk:
> An exception to this is if it's a component that's typically used by other libraries. Maybe we do want a `DeprecationWarning` initially so that we can update `cuDF` to use the new system (I haven't checked if it actually uses the config system anywhere, rather than just docs)? But maybe that's more work than it's worth.

Exactly, KvikIO is primarily used by other libraries :)

bdice:
> My preference is usually for a `FutureWarning` if we want people to actually see them (outside of `pytest` invoked unit tests, which un-hides DeprecationWarnings).

We exclusively use FutureWarning in cuDF and have recommended all RAPIDS libraries should match that. It’s also how most PyData libraries we’ve studied handle deprecations, to ensure visibility.

kingcrimsontianyu:
Thanks for the review!

vyasr:
For posterity, some of our historical discussions around DeprecationWarning and FutureWarning can be found in:
- https://github.com/rapidsai/cudf/pull/8603#discussion_r658298855
- https://github.com/rapidsai/cudf/pull/9347#discussion_r719464496

Also worth noting that we have a pre-commit hook in cudf to enforce never using DeprecationWarning, see https://github.com/rapidsai/cudf/issues/11251 and https://github.com/rapidsai/cudf/pull/11255.

PR #650, opened 2025-02-27
nvCOMP : don't use legacy API

madsbk:


vuule (approved):

wence- (approved):
Approving, on the assumption that there are other ways to provide stream/device?

madsbk:
> Approving, on the assumption that there are other ways to provide stream/device?

The nvCOMP API never supported stream/device :/

PR #656, opened 2025-03-07
Remove libcufile-dev packages from conda dependencies now that libkvikio is a shared library.

bdice:
For a long time, `libkvikio` was a header-only library. To make it usable by consumers, we included `libcufile-dev` in its run dependencies. Now that `libkvikio` is a shared library, we do not need to do that.

Also, ARM packages of `libkvikio` were depending on `libcufile` and getting it from the `nvidia` channel, which breaks things because the version is unconstrained by `cuda-version`. We cannot include a run dependency on `libcufile` on ARM until the next CUDA major version, because we need ARM packages to work with CUDA 12.0 and `libcufile` was only introduced in CUDA 12.2+.


bdice:
Closing. It's not clear what to do with this. We tried to improve this in #678 but ended up reverting to the status quo.

Issue #606, opened 2025-01-30
Add C++ benchmarks for local IO throughput


vuule:
Benchmarks should cover different workloads (size/number of reads/writes).
Multi-threaded benchmarks can also be considered.
Is possible, the throughput can be compared to the results from utilities like fio and gdsio.

JigaoLuo:
One point I’d like to raise, as I discovered this WIP today: I hope the tool could also support NVMe‑over‑Fabric. 
In principle, it should not require extra work, since once NVMeoF is set up, both fio and gdsio run seamlessly on it, once gdscheck shows NVMeoF as well.

JigaoLuo:
Thanks. I hope it’s alright that I include these off‑topic notes in this issue. I can fold them in later if needed.

That’s another story, as we update the DGX-OS6 kernel, then the driver for CUDA 13, and then redo the GDS setup on our DGX‑A100. We posted in:
- https://forums.developer.nvidia.com/t/kernel-bug-in-gds-on-dgx-os6-nvidia-fs1383-incompatible-page-table-version-0x00020000/350209/8?u=user157267
- https://github.com/NVIDIA/gds-nvidia-fs/issues/66 

One issue or one hesitation we had is that the documentation is not very clear for our case: updating the driver and, at the same time, enabling GDS.

The problem is that there are two separate document sections, one for updating the driver and another for installing GDS, but they are only loosely connected. In our case, we couldn’t find a confident way to follow everything successfully. That’s why we first tried the proprietary driver, which led to a kernel panic pasted in the issue66 above, and then switched to the open driver, which worked in the end. (The choice between proprietary and open drivers is still tied to a legacy version issue in GDS.)

kingcrimsontianyu:
@JigaoLuo  Thanks for raising this key documentation issue. It has been forwarded to the GDS team.

PR #659, opened 2025-03-12
Add C++ benchmarks (part 1/n)

kingcrimsontianyu:
Partially addresses https://github.com/rapidsai/kvikio/issues/606

This PR kick-starts the process of adding benchmarks to KvikIO. The following tasks are done:
- Add CMake scripts for the benchmark.
- Add a simple benchmark for the threadpool.

To build and run the benchmark programs in the dev container:
```
# Build the benchmarks
build-kvikio-cpp -DBUILD_TESTS=ON -DBUILD_BENCHMARKS=ON -j 16

# Run specific benchmark
~/kvikio/cpp/build/latest/benchmarks/<benchmark-name>
```

Sample output:
```
2025-03-14T04:38:46+00:00
Running ./THREADPOOL_BENCHMARK
Run on (16 X 5050 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1024 KiB (x8)
  L3 Unified 98304 KiB (x1)
Load Average: 0.34, 1.07, 1.22
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
---------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                 Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------------------------------------------------------------
BM_threadpool_compute:strong_scaling/1/min_time:2.000/real_time         162 ms        0.728 ms           17 threads=1
BM_threadpool_compute:strong_scaling/2/min_time:2.000/real_time        81.6 ms        0.885 ms           34 threads=2
BM_threadpool_compute:strong_scaling/4/min_time:2.000/real_time        41.1 ms        0.972 ms           68 threads=4
BM_threadpool_compute:strong_scaling/8/min_time:2.000/real_time        21.3 ms         1.40 ms          127 threads=8
BM_threadpool_compute:strong_scaling/16/min_time:2.000/real_time       18.7 ms         1.84 ms          150 threads=16
BM_threadpool_compute:strong_scaling/32/min_time:2.000/real_time       19.2 ms         2.76 ms          145 threads=32
BM_threadpool_compute:strong_scaling/64/min_time:2.000/real_time       20.8 ms         5.46 ms          139 threads=64
BM_threadpool_compute:weak_scaling/1/min_time:2.000/real_time          16.2 ms        0.135 ms          172 threads=1
BM_threadpool_compute:weak_scaling/2/min_time:2.000/real_time          16.3 ms        0.260 ms          171 threads=2
BM_threadpool_compute:weak_scaling/4/min_time:2.000/real_time          16.6 ms        0.527 ms          168 threads=4
BM_threadpool_compute:weak_scaling/8/min_time:2.000/real_time          17.2 ms         1.05 ms          164 threads=8
BM_threadpool_compute:weak_scaling/16/min_time:2.000/real_time         29.7 ms         2.66 ms           94 threads=16
BM_threadpool_compute:weak_scaling/32/min_time:2.000/real_time         60.9 ms         8.04 ms           46 threads=32
BM_threadpool_compute:weak_scaling/64/min_time:2.000/real_time          133 ms         36.0 ms           21 threads=64
```

### DIFF
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -15,19 +15,18 @@
 add_executable(THREADPOOL_BENCHMARK "threadpool/threadpool_benchmark.cpp")
 set_target_properties(THREADPOOL_BENCHMARK PROPERTIES INSTALL_RPATH "\$ORIGIN/../../lib")
 target_include_directories(THREADPOOL_BENCHMARK PRIVATE ../include ${cuFile_INCLUDE_DIRS})
-target_link_libraries(THREADPOOL_BENCHMARK
-PUBLIC benchmark::benchmark benchmark::benchmark_main kvikio)
+target_link_libraries(THREADPOOL_BENCHMARK PUBLIC benchmark::benchmark kvikio)
 
 if(CMAKE_COMPILER_IS_GNUCXX)
-set(KVIKIO_CXX_FLAGS "-Wall;-Werror;-Wno-unknown-pragmas")
-target_compile_options(
+  set(KVIKIO_CXX_FLAGS "-Wall;-Werror;-Wno-unknown-pragmas")
+  target_compile_options(
     THREADPOOL_BENCHMARK PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${KVIKIO_CXX_FLAGS}>"
-)
+  )
 endif()
 
 install(
-TARGETS THREADPOOL_BENCHMARK
-COMPONENT testing
-DESTINATION bin/benchmarks/libkvikio
-EXCLUDE_FROM_ALL
+  TARGETS THREADPOOL_BENCHMARK
+  COMPONENT testing
+  DESTINATION bin/benchmarks/libkvikio
+  EXCLUDE_FROM_ALL
 )
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -42,7 +42,7 @@ function(kvikio_add_benchmark)
   endif()
 
   add_executable(${_KVIKIO_NAME} ${_KVIKIO_SOURCES})
-  set_target_properties(${_KVIKIO_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN/../../lib")
+  set_target_properties(${_KVIKIO_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
   target_include_directories(${_KVIKIO_NAME} PRIVATE ../include ${cuFile_INCLUDE_DIRS})
   target_link_libraries(${_KVIKIO_NAME} PUBLIC benchmark::benchmark kvikio)
 
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -1,0 +1,33 @@
+# =============================================================================
+# Copyright (c) 2025, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+# =============================================================================
+
+add_executable(THREADPOOL_BENCHMARK "threadpool/threadpool_benchmark.cpp")
+set_target_properties(THREADPOOL_BENCHMARK PROPERTIES INSTALL_RPATH "\$ORIGIN/../../lib")
+target_include_directories(THREADPOOL_BENCHMARK PRIVATE ../include ${cuFile_INCLUDE_DIRS})
+target_link_libraries(THREADPOOL_BENCHMARK
+PUBLIC benchmark::benchmark benchmark::benchmark_main kvikio)
+
+if(CMAKE_COMPILER_IS_GNUCXX)
+set(KVIKIO_CXX_FLAGS "-Wall;-Werror;-Wno-unknown-pragmas")
+target_compile_options(
+    THREADPOOL_BENCHMARK PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${KVIKIO_CXX_FLAGS}>"
+)
+endif()
+
+install(
+TARGETS THREADPOOL_BENCHMARK
+COMPONENT testing
+DESTINATION bin/benchmarks/libkvikio
+EXCLUDE_FROM_ALL
+)
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -12,21 +12,51 @@
 # the License.
 # =============================================================================
 
-add_executable(THREADPOOL_BENCHMARK "threadpool/threadpool_benchmark.cpp")
-set_target_properties(THREADPOOL_BENCHMARK PROPERTIES INSTALL_RPATH "\$ORIGIN/../../lib")
-target_include_directories(THREADPOOL_BENCHMARK PRIVATE ../include ${cuFile_INCLUDE_DIRS})
-target_link_libraries(THREADPOOL_BENCHMARK PUBLIC benchmark::benchmark kvikio)
+#[=======================================================================[.rst:
+kvikio_add_benchmark
+--------------------
 
-if(CMAKE_COMPILER_IS_GNUCXX)
-  set(KVIKIO_CXX_FLAGS "-Wall;-Werror;-Wno-unknown-pragmas")
-  target_compile_options(
-    THREADPOOL_BENCHMARK PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${KVIKIO_CXX_FLAGS}>"
+Create a KvikIO benchmark.
+
+.. code-block:: cmake
+
+   kvikio_add_benchmark(NAME <name> SOURCES <sources>)
+
+   ``NAME``
+   Benchmark name. Single-value argument.
+
+   ``SOURCES``
+   List of source files for the benchmark. Multi-value argument.
+#]=======================================================================]
+function(kvikio_add_benchmark)
+  cmake_parse_arguments(
+    _KVIKIO # prefix
+    "" # optional
+    "NAME" # single value
+    "SOURCES" # multi-value
+    ${ARGN}
   )
-endif()
 
-install(
-  TARGETS THREADPOOL_BENCHMARK
-  COMPONENT testing
-  DESTINATION bin/benchmarks/libkvikio
-  EXCLUDE_FROM_ALL
-)
+  if(DEFINED _KVIKIO_UNPARSED_ARGUMENTS)
+    message(FATAL_ERROR "Unknown argument: ${_KVIKIO_UNPARSED_ARGUMENTS}")
+  endif()
+
+  add_executable(${_KVIKIO_NAME} ${_KVIKIO_SOURCES})
+  set_target_properties(${_KVIKIO_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN/../../lib")
+  target_include_directories(${_KVIKIO_NAME} PRIVATE ../include ${cuFile_INCLUDE_DIRS})
+  target_link_libraries(${_KVIKIO_NAME} PUBLIC benchmark::benchmark kvikio)
+
+  if(CMAKE_COMPILER_IS_GNUCXX)
+    set(KVIKIO_CXX_FLAGS "-Wall;-Werror;-Wno-unknown-pragmas")
+    target_compile_options(${_KVIKIO_NAME} PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${KVIKIO_CXX_FLAGS}>")
+  endif()
+
+  install(
+    TARGETS ${_KVIKIO_NAME}
+    COMPONENT testing
+    DESTINATION bin/benchmarks/libkvikio
+    EXCLUDE_FROM_ALL
+  )
+endfunction()
+
+kvikio_add_benchmark(NAME THREADPOOL_BENCHMARK SOURCES "threadpool/threadpool_benchmark.cpp")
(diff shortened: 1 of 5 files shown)


cpp/examples/CMakeLists.txt
```diff
@@ -13,14 +13,15 @@
...
+get_target_property(KVIKIO_INCLUDES kvikio::kvikio INCLUDE_DIRECTORIES)
 
 # Example: basic_io
 
 if(CUDAToolkit_FOUND)
   add_executable(BASIC_IO_EXAMPLE basic_io.cpp)
   set_target_properties(BASIC_IO_EXAMPLE PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
-  target_include_directories(BASIC_IO_EXAMPLE PRIVATE ../include ${cuFile_INCLUDE_DIRS})
-  target_link_libraries(BASIC_IO_EXAMPLE PRIVATE kvikio CUDA::cudart)
+  target_include_directories(BASIC_IO_EXAMPLE PRIVATE "${KVIKIO_INCLUDES}")
```
KyleFromNVIDIA, line 23: This is both redundant and wrong. It's redundant because the `target_link_libraries(kvikio::kvikio)` below causes `BASIC_IO_EXAMPLE` to already include `kvikio::kvikio`'s `INTERFACE_INCLUDE_DIRECTORIES` as well as linking to the library itself, as part of CMake's *usage requirements* functionality. It's wrong because you're including `kvikio::kvikio`'s `INCLUDE_DIRECTORIES`, which contains its `PRIVATE` and `PUBLIC` includes, rather than `INTERFACE_INCLUDE_DIRECTORIES`, which contains its `INTERFACE` and `PUBLIC` includes. If we look at https://github.com/rapidsai/kvikio/blob/3adbe7e6435c728dc8a03aba97094519f2a299c5/cpp/CMakeLists.txt#L172-L177, we can see that `kvikio::kvikio` doesn't get any strictly `PRIVATE` include directories. They're all `PUBLIC` or `INTERFACE`, and the lone `INTERFACE` directory is not relevant because it's only activated in an install context.

In short, you can take out this `target_include_directories()`, the one below, and the `get_target_property()` above.

cpp/examples/CMakeLists.txt
```diff
@@ -43,8 +44,8 @@ endif()
 
 add_executable(BASIC_NO_CUDA_EXAMPLE basic_no_cuda.cpp)
 set_target_properties(BASIC_NO_CUDA_EXAMPLE PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
-target_include_directories(BASIC_NO_CUDA_EXAMPLE PRIVATE ../include)
-target_link_libraries(BASIC_NO_CUDA_EXAMPLE PRIVATE kvikio)
+target_include_directories(BASIC_NO_CUDA_EXAMPLE PRIVATE "${KVIKIO_INCLUDES}")
```
KyleFromNVIDIA, line 47: As per my earlier comment, this can be removed.

cpp/examples/CMakeLists.txt
```diff
@@ -13,14 +13,15 @@
 # =============================================================================
 
 set(TEST_INSTALL_PATH bin/tests/libkvikio)
+get_target_property(KVIKIO_INCLUDES kvikio::kvikio INCLUDE_DIRECTORIES)
```
KyleFromNVIDIA, line 16: As per my earlier comment, this can be removed.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+  if(DEFINED _KVIKIO_UNPARSED_ARGUMENTS)
+    message(FATAL_ERROR "Unknown argument: ${_KVIKIO_UNPARSED_ARGUMENTS}")
+  endif()
+
+  add_executable(${_KVIKIO_NAME} ${_KVIKIO_SOURCES})
+  set_target_properties(${_KVIKIO_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
+
+  get_target_property(KVIKIO_INCLUDES kvikio::kvikio INCLUDE_DIRECTORIES)
+  target_include_directories(${_KVIKIO_NAME} PRIVATE "${KVIKIO_INCLUDES}")
```
KyleFromNVIDIA, line 48: As per my earlier comment, these lines can be removed.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+  target_link_libraries(${_KVIKIO_NAME} PUBLIC benchmark::benchmark kvikio::kvikio)
+
+  if(CMAKE_COMPILER_IS_GNUCXX)
+    set(KVIKIO_CXX_FLAGS "-Wall;-Werror;-Wno-unknown-pragmas")
+    target_compile_options(${_KVIKIO_NAME} PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${KVIKIO_CXX_FLAGS}>")
+  endif()
```
KyleFromNVIDIA, line 54: `CMAKE_COMPILER_IS_GNUCXX` is [deprecated](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html#deprecated-variables-for-languages). Use the [`CMAKE_CXX_COMPILER_ID`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html) variable instead:

```suggestion
  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    set(KVIKIO_CXX_FLAGS "-Wall;-Werror;-Wno-unknown-pragmas")
    target_compile_options(${_KVIKIO_NAME} PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${KVIKIO_CXX_FLAGS}>")
  endif()
```

KyleFromNVIDIA (changes requested):
Just a few small CMake issues that should be fixed.

cpp/benchmarks/CMakeLists.txt
kingcrimsontianyu: Thanks! Also updated other places where `CMAKE_COMPILER_IS_GNUCXX` was used.

cpp/examples/CMakeLists.txt
kingcrimsontianyu: Thanks! Great to know.

GregoryKimball:
Thanks @kingcrimsontianyu this is a great start! 

I think the "Time" column makes sense to me, showing an optimum of 8 threads in both the weak and strong scaling cases. The "CPU" column doesn't make much sense to me though.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include <cmath>
+#include <cstdint>
+
+#include <benchmark/benchmark.h>
+#include <kvikio/defaults.hpp>
+
+namespace kvikio {
+enum class ScalingType : uint8_t {
+  StrongScaling,
```
mhaseeb123, line 32: I think these need be in all caps unless the styling is not the same across libcudf and kvikio

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  std::string label;
+  std::size_t num_compute_tasks;
+  if constexpr (scaling_type == ScalingType::StrongScaling) {
+    num_compute_tasks = 10'000;
+  } else {
+    num_compute_tasks = 1'000 * num_threads;
+  }
+
+  std::size_t const num_compute_iterations{100'000};
```
mhaseeb123, line 58: Should this be varied on an axis?

```suggestion
  auto constexpr num_compute_iterations = std::size_t{100'000};
```

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+  for (std::size_t i = 0u; i < num_compute_iterations; ++i) {
+    auto x{static_cast<double>(i)};
+    res += std::sqrt(x) + std::cbrt(x) + std::sin(x);
+  }
+}
+
+template <ScalingType scaling_type>
+void BM_threadpool_compute(benchmark::State& state)
+{
+  auto num_threads = state.range(0);
```
mhaseeb123, line 48:
```suggestion
  auto const num_threads = state.range(0);
```

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  std::string label;
```
mhaseeb123, line 50: I think this is unused

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+    num_compute_tasks = 10'000;
+  } else {
+    num_compute_tasks = 1'000 * num_threads;
+  }
+
+  std::size_t const num_compute_iterations{100'000};
+  kvikio::defaults::set_thread_pool_nthreads(num_threads);
+
+  for (auto _ : state) {
+    for (std::size_t i = 0u; i < num_compute_tasks; ++i) {
```
mhaseeb123, line 62: nit: 
```suggestion
    for (auto i = std::size_t{0}; i < num_compute_tasks; ++i) {
```

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+      [[maybe_unused]] auto fut = kvikio::defaults::thread_pool().submit_task(
```
mhaseeb123, line 63: Would this make sense if we just wanna ignore the returned future
```suggestion
      std::ignore = kvikio::defaults::thread_pool().submit_task(
```

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+  auto num_threads = state.range(0);
+
+  std::string label;
+  std::size_t num_compute_tasks;
+  if constexpr (scaling_type == ScalingType::StrongScaling) {
+    num_compute_tasks = 10'000;
+  } else {
+    num_compute_tasks = 1'000 * num_threads;
+  }
```
mhaseeb123, line 56:
```suggestion
  auto const num_compute_tasks = [&](){
  if constexpr (scaling_type == ScalingType::StrongScaling) {
    return static_cast<size_t>(10'000);
  } else {
    return static_cast<size_t>(1'000 * num_threads);
  }
  }();
```

mhaseeb123 (commented):
Would appreciate some comments in `threadpool_benchmark.cpp`

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: Thanks! Done.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: For this PR I'd like to keep it single-valued and hardcoded, and make the results concise and clean. In the future, it would be a good idea to make it a parameter and/or allow sweeping.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: Thanks. Done.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: Thanks. Removed.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: Done.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: Both are acceptable, and the choice is more opinion-based. I think `std::ignore` is primarily designed for tuples, and the attribute feels a bit more C++17 idiomatic.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: Thanks! Considering that `scaling_type` is a compile-time constant, I've made the expression more concise while making it `const`:
```cpp
  std::size_t const num_compute_tasks =
    (scaling_type == ScalingType::STRONG_SCALING) ? 10'000 : (1'000 * num_threads);
```

kingcrimsontianyu:
@GregoryKimball `Time` refers to the wall-clock time, and `CPU` is the CPU time of the main thread (implemented using `CLOCK_THREAD_CPUTIME_ID`  from [clock_gettime](https://www.man7.org/linux/man-pages/man3/clock_gettime.3.html)). For this benchmark, `CPU` is a measure of the overhead of thread pool management that takes place on the main thread. I think this will help us find how well the BS pool implementation would scale on a many-core system (Grace), where the use of single task queue with mutexes and condition variables might turn out suboptimal.

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
```diff
@@ -0,0 +1,95 @@
...
+    (scaling_type == ScalingType::STRONG_SCALING) ? 10'000 : (1'000 * num_threads);
+
+  std::size_t constexpr num_compute_iterations{1'000};
+  kvikio::defaults::set_thread_pool_nthreads(num_threads);
+
+  for (auto _ : state) {
+    // Submit a total of "num_compute_tasks" tasks to the thread pool.
+    for (auto i = std::size_t{0}; i < num_compute_tasks; ++i) {
+      [[maybe_unused]] auto fut = kvikio::defaults::thread_pool().submit_task(
+        [num_compute_iterations = num_compute_iterations] {
```
madsbk, line 62:
```suggestion
        [num_compute_iterations] {
```

madsbk (approved):
Looks great

cpp/benchmarks/threadpool/threadpool_benchmark.cpp
kingcrimsontianyu: Thanks. Actually, because `num_compute_iteration` is now a `constexpr` variable, if the init-capture is not used, then it is not necessary to capture `num_compute_iteration`.

KyleFromNVIDIA (approved):

kingcrimsontianyu:
To avoid confusion, explanation of the metrics is added to the result in another ongoing PR (https://github.com/rapidsai/kvikio/pull/664).

Screenshot from 2025-03-14 11-01-10

mhaseeb123 (approved):
Approving C++ changes

PR #662, opened 2025-03-13
Add pom.xml to update-version.sh

raydouglass:
Update the java `pom.xml` in `update-version.sh`.

Copied from https://github.com/rapidsai/cudf/blob/branch-25.04/ci/release/update-version.sh#L90-L92

Skipping CI since this file is not tested in CI.

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -60,3 +60,7 @@ done
 
 # The example of a downstream project
 sed_runner "s/find_and_configure_kvikio(.*)/find_and_configure_kvikio(\"${NEXT_SHORT_TAG}\")/g" "cpp/examples/downstream/cmake/get_kvikio.cmake"
+
+# Java files
+NEXT_FULL_JAVA_TAG="${NEXT_SHORT_TAG}.${PATCH_PEP440}-SNAPSHOT"
+sed_runner "s|<version>.*-SNAPSHOT</version>|<version>${NEXT_FULL_JAVA_TAG}</version>|g" java/pom.xml


ci/release/update-version.sh
```diff
@@ -60,3 +60,7 @@ done
 
 # The example of a downstream project
 sed_runner "s/find_and_configure_kvikio(.*)/find_and_configure_kvikio(\"${NEXT_SHORT_TAG}\")/g" "cpp/examples/downstream/cmake/get_kvikio.cmake"
+
+# Java files
+NEXT_FULL_JAVA_TAG="${NEXT_SHORT_TAG}.${PATCH_PEP440}-SNAPSHOT"
+sed_runner "s|<version>.*-SNAPSHOT</version>|<version>${NEXT_FULL_JAVA_TAG}</version>|g" java/pom.xml
```
jameslamb, line 66:
```suggestion
sed_runner "s|^  <version>.*-SNAPSHOT</version>|  <version>${NEXT_FULL_JAVA_TAG}</version>|g" java/pom.xml
```

There are a lot of `<version></version>` tags in that file, but currently only 1 matching `-SNAPSHOT`, so this should be safe enouehg for now.

I was trying to think of ways to make this even safer, and think I have one. I noticed there's only 1 "top-level" version preceded by exactly 2 spaces, and that's the one we want to modify... would you consider padding spaces into this?

Alternatively, I *think* XML supports arbitrary data on tags similar to HTML, so maybe something like this to help identify the line?

```xml
  <version  id="kvikio-version">25.04.0-SNAPSHOT</version>
```

jameslamb (approved):
I left a small suggestion, but it's VERY small... feel free to ignore it and merge this if you'd like.

ci/release/update-version.sh
raydouglass: I think these suggestions can be a follow up. `cudf` has the same problem. But `cuvs` [uses comments](https://github.com/rapidsai/cuvs/blob/branch-25.04/ci/release/update-version.sh#L106) to be more selective.

We should probably standardize on something, so I'll open an issue.

raydouglass: Filed https://github.com/rapidsai/build-planning/issues/163

PR #666, opened 2025-03-14
Make created CUDA streams non-blocking to avoid implicit sync with the default stream

kingcrimsontianyu:


madsbk:
Did it do anything for performance?

PR #669, opened 2025-03-18
Fix comment typo in buffer_register

JigaoLuo:
- No code change
- Only typo fix in function signature comment: should be `@param size` instead of `@param length`

https://github.com/rapidsai/kvikio/blob/93a447f451dba0b1777be8753f4e4d180c40643b/cpp/include/kvikio/buffer.hpp#L26-L41


vuule:
Thank you for the PR! I'm moving it to 25.06 because we're already in 25.04 burndown.

bdice (approved):

PR #671, opened 2025-03-20
Improve NVTX annotation for functions

kingcrimsontianyu:
## Background
Previously, to add NVTX function annotation for profiling, one of the following methods is used:
```cpp
namespace kvikio::detail {
// Method 1
void f() {
    KVIKIO_NVTX_FUNC_RANGE();
}

// Method 2
void f() {
    KVIKIO_NVTX_SCOPED_RANGE("detail::f()", 0);
}
} // end namespace kvikio::detail
```

Method 1 has the limitation that:
- `__func__` is used as the function name, which is unqualified and contains no information on namespace. Function overloads cannot be well handled.
- Users cannot specify NVTX payload or color.
- There is no way to annotate a lambda. Any lambda is simply named `operator()`.

Method 2 has the limitation that:
- Users have to manually enter the function name. Naming inconsistency has been observed across different source files.
- Users have to specify an NVTX payload value.

## This PR
This PR eliminates the limitations above to improve the annotation experience. More concretely:
- `KVIKIO_NVTX_FUNC_RANGE()` now uses fully qualified name with namespace, return type and parameter types. Function overloads can be properly handled.
- `KVIKIO_NVTX_FUNC_RANGE()` now works for lambdas. For example, for a lambda inside a function `kvikio::sample::foo()`, the name of the lambda will be: `kvikio::sample::foo::<lambda()>`
- `KVIKIO_NVTX_FUNC_RANGE()` now accepts two optional parameters: payload and color.
- `KVIKIO_NVTX_SCOPED_RANGE()` now only has one mandatory parameter, which is the message; both the payload and color have become optional.

With the flexible `KVIKIO_NVTX_FUNC_RANGE()` available, **this PR adds additional traces to many functions**.

### DIFF
--- a/cpp/include/kvikio/nvtx.hpp
+++ b/cpp/include/kvikio/nvtx.hpp
@@ -50,9 +50,29 @@ using nvtx_registered_string_type = nvtx3::registered_string_in<libkvikio_domain
   }(message)
 
 // Implementation of KVIKIO_NVTX_FUNC_RANGE()
-#define KVIKIO_NVTX_FUNC_RANGE_IMPL() NVTX3_FUNC_RANGE_IN(kvikio::libkvikio_domain)
+#define KVIKIO_NVTX_FUNC_RANGE_IMPL_0() KVIKIO_NVTX_SCOPED_RANGE_IMPL_1(__PRETTY_FUNCTION__)
+#define KVIKIO_NVTX_FUNC_RANGE_IMPL_1(payload) \
+  KVIKIO_NVTX_SCOPED_RANGE_IMPL_2(__PRETTY_FUNCTION__, payload)
+#define KVIKIO_NVTX_FUNC_RANGE_IMPL_2(payload, color) \
+  KVIKIO_NVTX_SCOPED_RANGE_IMPL_3(__PRETTY_FUNCTION__, payload, color)
+#define KVIKIO_NVTX_FUNC_RANGE_SELECTOR(_0, _1, _2, NAME, ...) NAME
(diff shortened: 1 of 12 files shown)


kingcrimsontianyu:
## Sample code and resulting annotations

### Example: pass no argument to `KVIKIO_NVTX_FUNC_RANGE()`
```cpp
namespace quake {
namespace ranger {
void foo0(double, float)
{
  KVIKIO_NVTX_FUNC_RANGE();
  std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
}  // namespace ranger
}  // namespace quake
```
image

### Example: add payload
```cpp
namespace quake {
namespace ranger {
void foo1(double, float)
{
  KVIKIO_NVTX_FUNC_RANGE(123);
  std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
}  // namespace ranger
}  // namespace quake
```
image


### Example: add payload and color
```cpp
namespace quake {
namespace ranger {
void foo2(double, float)
{
  nvtx3::rgb color{0, 255, 0};
  KVIKIO_NVTX_FUNC_RANGE(123, color);
  std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
}  // namespace ranger
}  // namespace quake
```
image

### Example: trace lambda
```cpp
namespace quake {
namespace ranger {
void foo2(double, float)
{
  nvtx3::rgb color{0, 255, 0};
  KVIKIO_NVTX_FUNC_RANGE(123, color);
  std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
}  // namespace ranger
}  // namespace quake
```
image

madsbk (approved):
Nice!

PR #674, opened 2025-03-21
chore(conda): enable strict channel priority for conda build

gforsyth:
This PR enables strict channel priority for building conda packages with `conda-build`.

xref rapidsai/build-planning#84


gforsyth:
It isn't going to be possible to use `strict` priority with some cuda11 builds because of `libcufile` versions spread across `conda-forge` and `nvidia` (newer version are all available on `conda-forge` but older versions are only on `nvidia`).

I'll have to update the build scripts to not set strict priority for 11.8 builds on x86_64

jakirkham:
In that case, could you please note this in the OP of issue ( https://github.com/rapidsai/build-planning/issues/84 )?

Since it sounds like we need two passes of all RAPIDS projects to implement this, would also update the checklist to reflect this is part 1

ci/build_python.sh
```diff
@@ -5,6 +5,20 @@ set -euo pipefail
...
+  rapids-logger "Channel priority disabled for CUDA 11 builds"
+else
+  rapids-logger "Setting strict channel priority for CUDA 12 builds"
+  conda config --set channel_priority strict
+fi
+
+# Remove `rapidsai` channel for non-release builds
+if ! rapids-is-release-build; then
+  conda config --remove channels rapidsai
+fi
```
msarahan, line 20: Not terribly important, but this boilerplate seems like it could be added to a function in [gha-tools](https://github.com/rapidsai/gha-tools) to reduce the copy-paste stuff.

msarahan (approved):

ci/build_python.sh
gforsyth: Yep, strong agree -- I have a draft PR open for that here https://github.com/rapidsai/gha-tools/pull/153 but I didn't want to merge it in mid-release.
I think I'm going to hold off on any more channel priority PRs until I can change things more easily.

gforsyth:
This is ready for a look now that the upstream `gha-tools` changes have been made.

Issue #530, opened 2024-10-31
[FEA] Add an alternative method of reading from files that uses `mmap`


vuule:
Memory mapping could be faster in some use cases and/or some systems. 
This is a proposal to add an option to replace file reads with copying from the mapped-to buffer.
Some open (AFAIK) design decisions:

- How should this interact with compatibility mode options.
- Should this be used for reads to device memory, or when.

vuule:
this is addressed now

kingcrimsontianyu:
# Related PRs

- [ ] C++ implementation: #740 
- [ ] Python binding: #742 
- [ ] Example: #743 
- [ ] Benchmark: TBA

# Benchmark results
 
## GH200 (C2C)
### t1: PDS all queries 

- t1.1: Hot page cache
  - Each query is repeated 11 times. The average of the elapsed time of the last 10 repetitions is used. Same for the case of cold page cache.

  ![Image](https://github.com/user-attachments/assets/30fde2c4-3f23-4e67-8701-5a83a4b4b4d0)
 
- t1.2: Cold page cache

  ![Image](https://github.com/user-attachments/assets/e00cb5f4-7506-43dd-85d8-39070dc6a405)

### t2: PDF Q5 parameter sweep (number of threads, task size)
- Query 5 at a certain (number of threads, task size) coordinate is repeated 5 times. THe average of the elapsed time of the last 4 repetitions is used.

![Image](https://github.com/user-attachments/assets/54396080-ecbb-4021-b428-13971af409c4)

 
### t3: cuDF benchmark
- t3.1: Parquet
  
  ```
  PARQUET_READER_NVBENCH -d 0 -b parquet_read_io_compression -a compression_type=NONE -a io_type=FILEPATH -a cardinality=0 -a run_length=1 --min-samples 80 --timeout 60 --csv stdout
  ```

  || Standard I/O (s)| Mapped I/O (s) |
  | -------- | ------- | ------- |
  |Hot page cache|0.02365 (0.5381%)|0.02318 (2.163%)|
  | Cold page cache|0.1651 (5.02%) |0.1672 (2.143%) |

- t3.2: ORC
  
  || Standard I/O (s)| Mapped I/O (s) |
  | -------- | ------- | ------- |
  |Hot page cache|0.03901 (0.2444%)|0.03688 (1.46%)|
  | Cold page cache|0.1803 (4.862%)|0.1761 (4.605%)|

- t3.3: Json

[...]

PR #677, opened 2025-03-24
Make page size portable by querying it at runtime

kingcrimsontianyu:
So far the parameter page size in KvikIO has been a compile-time constant (4K). On GH200, however, the page size is 64K. This PR makes it portable by querying its value at runtime.

### DIFF
--- a/cpp/include/kvikio/parallel_operation.hpp
+++ b/cpp/include/kvikio/parallel_operation.hpp
@@ -156,7 +156,7 @@ std::future<std::size_t> parallel_io(F op,
                                       decltype(devPtr_offset)>);
 
   // Single-task guard
-  if (task_size >= size || page_size >= size) {
+  if (task_size >= size || get_page_size() >= size) {
     return detail::submit_task(op, buf, size, file_offset, devPtr_offset, call_idx, nvtx_color);
   }
 
--- a/cpp/src/utils.cpp
+++ b/cpp/src/utils.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <unistd.h>
 #include <cstring>
 #include <iostream>
 #include <map>
@@ -27,6 +28,12 @@
 #include <kvikio/utils.hpp>
 
 namespace kvikio {
+
+std::size_t get_page_size()
+{
+  static auto const page_size = static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
+  return page_size;
+}
 
 off_t convert_size2off(std::size_t x)
 {
--- a/cpp/include/kvikio/utils.hpp
+++ b/cpp/include/kvikio/utils.hpp
@@ -28,8 +28,7 @@
 
 namespace kvikio {
 
-// cuFile defines a page size to 4 KiB
-inline constexpr std::size_t page_size = 4096;
+std::size_t get_page_size();
 
 [[nodiscard]] off_t convert_size2off(std::size_t x);
 


cpp/src/utils.cpp
```diff
@@ -28,6 +29,12 @@
 
 namespace kvikio {
 
+std::size_t get_page_size()
+{
+  static auto const page_size = static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
```
wence-, line 34: I think there are people building kvikio on non-linux systems. Do we need an alternative on (say) macos📷?

kingcrimsontianyu: Thanks for pointing that out. I was wondering if KvikIO would extend beyond what [RARPIDS system requirements ask ](https://docs.rapids.ai/install/#system-req).

On a separate note, what about compilers? I was expecting gcc/clang, and therefore used the following non-standard extensions in a recent merged PR:
- https://github.com/rapidsai/kvikio/pull/671/files#diff-d22c65abd4ba19a3bc834c80b4d5691223a24eeb3f90f3b7822cfe1096b8df46R53 : `__PRETTY_FUNCTION__ `
- https://github.com/rapidsai/kvikio/pull/671/files#diff-d22c65abd4ba19a3bc834c80b4d5691223a24eeb3f90f3b7822cfe1096b8df46R61: `##__VA_ARGS__`

cc @madsbk

madsbk: > I think there are people building kvikio on non-linux systems. Do we need an alternative on (say) macos📷?

I don't think we officially supports non-linux systems? At least, we only test on linux.
cc. @GregoryKimball, @jakirkham, @quasiben.

jakirkham (commented):
Thanks Tianyu! 🙏

Generally seems reasonable. Had a minor suggestion below

cpp/src/utils.cpp
```diff
@@ -28,6 +29,12 @@
...
+  return page_size;
```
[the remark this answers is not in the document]
kingcrimsontianyu: Is there a reason we do not want to cache the result in a static variable?
On my 7800 X3D processor, 10K calls to `kvikio::get_page_size()` has a median of 61 ns for both the cached and non-cached version. But I'm not quite sure about the overhead of `sysconf` on other systems. None the less, I'd consider it a good practice to cache the result.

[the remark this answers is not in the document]
kingcrimsontianyu: That's interesting. `_SC_PAGESIZE` was available in glibc 1.09.1 (1996) or earlier, and referenced by the Linux kernel 3.0.1 (2011) or earlier. I think it far predates GDS itself. So I believe it's safe to only use it in KvikIO.

PS: glibc 1.09.1 literally has this definition :smile: 
```cpp
#define getpagesize() sysconf(_SC_PAGESIZE)
```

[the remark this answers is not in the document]
jakirkham: Is there any reason the value would change?

Would think they have very thoroughly optimized `sysconf`. Just glancing at [a generic implementation in GLIBC]( https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/posix/sysconf.c;hb=HEAD ), it looks like they have pushed everything into a `switch/case` that should be very well optimized

If we do think this optimization matters, then we should consider using the same approach at the call-site (of which there appears to only be 1) or directly inlining it there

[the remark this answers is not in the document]
kingcrimsontianyu: For page size query, the switch case construct ends up with another internal function call `__getpagesize ()`.

I'm not sure I understand the call-site/inline part you mentioned. If it is about function inlining, which is tangent to the use of `static` for caching, then I'd like to avoid inlining `kvikio::get_page_size()`. The reason is we've been trying to separate interface from implementation since KvikIO became a shared library.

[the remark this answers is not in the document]
jakirkham: Have added some suggestions below to hopefully make this more clear

cpp/include/kvikio/parallel_operation.hpp
```diff
@@ -156,7 +156,7 @@ std::future<std::size_t> parallel_io(F op,
                                       decltype(devPtr_offset)>);
```
jakirkham, line 157: Since we only use page size here, should we just define it here

```suggestion
  static auto const page_size = static_cast<std::size_t>(sysconf(_SC_PAGESIZE));

```

Would also need to add this to the file header

```cpp
#include <unistd.h>
```

cpp/include/kvikio/parallel_operation.hpp
```diff
@@ -156,7 +156,7 @@ std::future<std::size_t> parallel_io(F op,
...
 
   // Single-task guard
-  if (task_size >= size || page_size >= size) {
+  if (task_size >= size || get_page_size() >= size) {
```
jakirkham, line 159:
```suggestion
  if (task_size >= size || page_size >= size) {
```

jakirkham (commented):
Tianyu have added a couple comments below to clarify how we could inline this function

cpp/include/kvikio/parallel_operation.hpp
kingcrimsontianyu: The page size will also be needed in the mmap feature that I'll be working on (https://github.com/rapidsai/kvikio/issues/530), so I believe it is necessary to stay in `util.hpp/cpp`.

cpp/include/kvikio/parallel_operation.hpp
kingcrimsontianyu: I'd disagree.

First, the meaning of `static` will change in this context, and will give `page_size` an internal linkage. There's no benefit of doing so. Imagine, in the future, we have an inline function using `page_size` and that inline function is used in more than one TUs, this will be a violation of ODR and cause undefined behavior.

Second, making `page_size` a static global variable will result in every TU that includes `parallel_operation.hpp` calling `sysconf()` before the main function. With this, we lose the benefit of lazy evaluation from the static variable `page_size` in the function `get_page_size()`.

In short, I think the current implementation in this PR is optimal.

cpp/include/kvikio/parallel_operation.hpp
jakirkham: Ok if we will use this in more places, agree it make sense to have a function

jakirkham (approved):

wence- (approved):

madsbk (approved):

PR #678, opened 2025-04-01
Port all conda recipes to `rattler-build`

gforsyth:
Port all condabuild recipes over to use `rattler-build` instead.

Contributes to rapidsai/build-planning#47

- To satisfy `rattler-build`, this changes all the licenses in the `pyproject.toml` files to the SPDX-compliant `Apache-2.0` instead of `Apache 2.0`

### DIFF
--- a/conda/recipes/libkvikio/recipe.yaml
+++ b/conda/recipes/libkvikio/recipe.yaml
@@ -99,8 +99,8 @@ outputs:
                 - libcufile ${{ cuda11_libcufile_run_version }}
                 - libcufile-dev ${{ cuda11_libcufile_run_version }}
           else:
-            - if: linux64
-              then: libcufile
+            - if: linux
+              then: libcufile-dev
       ignore_run_exports:
         by_name:
           - cuda-version
@@ -154,7 +154,7 @@ outputs:
                 - libcufile =${{ cuda11_libcufile_host_version }}
           else:
             - cuda-cudart
-            - if: linux64
+            - if: linux
               then: libcufile
       ignore_run_exports:
         by_name:


AyodeAwe (approved):

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -0,0 +1,173 @@
...
+        - if: cuda_major == "11"
+          then:
+            - cudatoolkit
+            - if: linux64
+              then:
+                - libcufile ${{ cuda11_libcufile_run_version }}
+                - libcufile-dev ${{ cuda11_libcufile_run_version }}
+          else:
+            - if: linux
+              then: libcufile-dev
```
bdice, line 105:
```suggestion
            - if: linux64
              then: libcufile
```
We probably need to use `libcufile` here. Historically we used the `-dev` package as a runtime dependency because `libkvikio` was header-only, and thus consumers needed `libcufile-dev` any time that `libkvikio` was used (only in host environments). I worked on this in #656, but that PR fell off my radar.

Also, we should only require `libcufile` on linux64, because it did not exist on ARM until CUDA 12.2. We can't easily write "use `libcufile` on ARM if using CUDA 12.2+" as a requirement, afaik. I also fixed in that in #656. There is [a bug that is observable in this PR's CI](https://github.com/rapidsai/kvikio/actions/runs/14206218562/job/39804983799?pr=678#step:9:253) where ARM builds are getting `libcufile 1.9.1.3` from the `nvidia` channel because that channel isn't correctly constrained to `cuda-version` until CUDA 12.5, and the ARM tests are run on `cuda-version==12.0`. Strict priority would've made this more visible.

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -0,0 +1,173 @@
...
+        - if: cuda_major == "11"
+          then:
+            - cudatoolkit
+            - if: linux64
+              then:
+                - libcufile =${{ cuda11_libcufile_host_version }}
+          else:
+            - cuda-cudart
+            - if: linux
+              then: libcufile
```
bdice, line 160:
```suggestion
            - if: linux64
              then: libcufile
```

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -0,0 +1,173 @@
...
+      - if: cuda_major == "11"
+        then:
+          - cudatoolkit
+          - if: linux64
+            then:
+              - libcufile =${{ cuda11_libcufile_host_version }}
+              - libcufile-dev =${{ cuda11_libcufile_host_version }}
+        else:
+          - if: linux
+            then: libcufile-dev
```
bdice, line 71:
```suggestion
          - if: linux64
            then: libcufile-dev
```

bdice (changes requested):
Tacking on my changes from #656 into this PR. Please take these suggestions with a grain of salt -- I haven't fully validated them, and there may be some gray area here where we need to support libcufile on ARM and are willing to accept some minor issues on CUDA 12.0/12.1 in order to get the benefits of libcufile on CUDA 12.2+.

conda/recipes/libkvikio/recipe.yaml
gforsyth: Happy to try it -- I changed it to `linux` (to match the existing recipe) because of the failures here: https://github.com/rapidsai/kvikio/actions/runs/14204454058/job/39799290962?pr=678

but it could be that also switching to `libcufile` fixes those up

gforsyth:
This is the same failure I was seeing before:

```
=========================== short test summary info ============================
FAILED tests/test_cufile_driver.py::test_open_and_close - RuntimeError: KvikIO failure at: /tmp/conda-bld-output/bld/rattler-build_libkvikio/work/cpp/src/shim/cufile.cpp:99: KvikIO not compiled with cuFile.h
FAILED tests/test_cufile_driver.py::test_property_accessor - RuntimeError: KvikIO failure at: /tmp/conda-bld-output/bld/rattler-build_libkvikio/work/cpp/src/cufile/driver.cpp:181: KvikIO not compiled with cuFile.h
= 2 failed, 1350 passed, 105 skipped, 10 xfailed, 2 xpassed in 71.12s (0:01:11) =
```

On the arm job testing against cuda 12.0.1

gforsyth:
Hey @bdice -- I think I've got a reasonable compromise in-place here if you want to take a look

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -0,0 +1,183 @@
...
+              then:
+                - libcufile ${{ cuda11_libcufile_run_version }}
+                - libcufile-dev ${{ cuda11_libcufile_run_version }}
+          else:
+            - if: arm64
+              then:
+                - if: match(cuda_version, ">12.1")
+                  then: libcufile
+                  else: libcufile-dev
+              else: libcufile
```
bdice, line 107: Can you help me understand what this is supposed to do? This will give `libcufile` for all CUDA 12 and linux64 jobs, and CUDA 11 ARM jobs will get `libcufile-dev`. But how does that work? There is no libcufile on ARM for CUDA 11 at all.

bdice: The first builds of `libcufile` for ARM came with CUDA 12.2.

If we could break `libkvikio` into more build variants, this is what would be correct:
- CUDA 11 linux64 (has libcufile-dev at build time, libcufile at runtime)
- CUDA 12 linux64 (has libcufile-dev at build time, libcufile at runtime)
- CUDA 11 arm64 (no libcufile at build time or runtime)
- CUDA 12.0 / 12.1 arm64 (no libcufile at build time or runtime)
- CUDA 12.2+ arm64 (has libcufile-dev at build time, libcufile at runtime)

We've been adding libcufile to all CUDA 12 arm64 builds, so effectively "CUDA 12.0 / 12.1 arm64" has been fetching libcufile from CUDA 12.4 off the `nvidia` channel at runtime, and "CUDA 12.2+ arm64" has been finding the libcufile with a matching `cuda-version` on `conda-forge`.

bdice: If we retain the status quo, where we have a `libcufile` dependency that is being satisfied by the `nvidia` channel, we won't be able to drop the `nvidia` channel in the future when we drop CUDA 11. It's fundamentally incorrect anyway, the only reason that it finds `libcufile` from the `nvidia` channel is because (a) that channel is in the list of channels and (b) the `nvidia` channel packages aren't correctly constrained by `cuda-version` until CUDA 12.5. That's why it gets the `libcufile` build on the `nvidia` channel that shipped with CUDA 12.4.

bdice: We don't want to drop libcufile support on CUDA 12 ARM entirely, which is why we've accepted the status quo thus far. However, we may need to invest in making a second build variant of the CUDA 12 ARM package -- one for CUDA 12.0/12.1 and one for CUDA 12.2+. That's the only way for us to package this "correctly" while still keeping support for libcufile on CUDA 12 ARM platforms.

bdice: The other important consideration is that the `match` syntax doesn't really help us that much because it is evaluated at build time -- we are only doing builds with CUDA 12.8 right now. It might help us if we added a second build that uses CUDA 12.0 or 12.1, but we would need for the `else:` to be empty (no `libcufile` packages on CUDA 12.0/12.1, whereas it would currently use `libcufile-dev`).

bdice: I know this is a huge wall of text, I just needed to write it out for my own sanity (it's hard to communicate the complexity of this problem/solution space). If you want to defer on this until later, I am okay with keeping the status quo where some CUDA 12.0/12.1 ARM configurations end up pulling a library from the `nvidia` channel. It's not desirable, and it won't work with strict priority, but it would let us unblock rattler-build work.

gforsyth: Added a follow-up item to rapidsai/build-planning#47 with a link to this discussion.

bdice (approved):
Approving the specific concerns I had about libcufile on ARM. Sticking to the status quo for now and revisiting later seems like the best we can do.

PR #683, opened 2025-04-04
Segfault repro

kingcrimsontianyu:


jakirkham:
Is it worth turning this into some kind of test?

PR #691, opened 2025-04-10
Fix setting region in C++ constructor instead of via the env var

### LINKED ISSUE #687: Passing region through the API doesn't work
When using this API:
```
S3Endpoint::S3Endpoint(std::string const& bucket_name,
                       std::string const& object_name,
                       std::optional<std::string> aws_region,
                       std::optional<std::string> aws_access_key,
                       std::optional<std::string> aws_secret_access_key,
                       std::optional<std::string> aws_endpoint_url)
```
the region is lost, so it results in an error like `curl_easy_perform() error near /home/joost/workspace/cuda/kvikio/cpp/src/remote_handle.cpp:255(Could not resolve host: bucket.s3..amazonaws.com)` (the region should be between the 2 dots).
The culprit seems to be the `std::move` in line https://github.com/rapidsai/kvikio/blob/branch-25.06/cpp/src/remote_handle.cpp#L256.

joosthooz:
This PR aims to fix #687 by eliminating a `std::move` of something that is used twice in a list of arguments. Because argument evaluation order is unspecified, you don't know which one will be evaluated first.

### DIFF
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -128,7 +128,7 @@ class S3Endpoint : public RemoteEndpoint {
    */
   static std::string url_from_bucket_and_object(std::string const& bucket_name,
                                                 std::string const& object_name,
-                                                std::optional<std::string> const& aws_region,
+                                                std::optional<std::string> aws_region,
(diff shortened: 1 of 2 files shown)


madsbk (approved):
Thanks @joosthooz, good catch!

PR #692, opened 2025-04-10
Remove usage of APIs outside the limited API

vyasr:
I hadn't revisited kvikio's usage of CPython's functions outside the limited API since #504, and while reviving rapidsai/devcontainers#278 I found that these usages persist and are easy enough to remove.


python/kvikio/kvikio/_lib/arr.pyx
```diff
@@ -125,8 +129,7 @@ cdef class Array:
                 self.shape_mv = None
                 self.strides_mv = None
         else:
-            mv = PyMemoryView_FromObject(obj)
-            pybuf = PyMemoryView_GET_BUFFER(mv)
+            PyObject_GetBuffer(obj, &pybuf, PyBUF_FULL_RO)
```
jakirkham, line 132: If we are going to do this, it needs to be part of a `try...finally...` block

python/kvikio/kvikio/_lib/arr.pyx
```diff
@@ -203,7 +207,7 @@ cdef class Array:
         for i in range(self.ndim):
             o = self.shape_mv[i]
             Py_INCREF(o)
-            PyTuple_SET_ITEM(shape, i, o)
+            PyTuple_SetItem(shape, i, o)
```
jakirkham, line 210: Is the incref still needed here?

Same question occurs with similar lines below

jakirkham (dismissed):
Thank Vyas! 🙏

So my understanding is these are all macros. As a result by definition they are not part of the limited API. However they are eliminated as part of the build

So the real question is what do they do under-the-hood. Usually the answer is attribute access or assignment

Given the `struct` types referenced here are part of stable and limited API, I suspect whatever issue we are seeing is a tooling bug and not actually something we need to change. Would be interested in hearing more about how you are checking this and what results the tool is presenting

vyasr:
I'm not sure I follow your logic. Just because something is a macro it doesn't mean that it will always be available in a limited api build. Many macros in cpython are conditionally defined based on whether we are compiling under the limited api or not.

In this case, my test doesn't involve any tooling, I am simply instructing python to be compiled with the limited api. You can see the test in https://github.com/rapidsai/devcontainers/actions/runs/14388179703/job/40348643586. The result is a compilation failure indicating that these functions/macros are not defined when the limited api macro is set.

python/kvikio/kvikio/_lib/arr.pyx
vyasr: It looks like both `PyTuple_SET_ITEM` and `PyTuple_SetItem` [both steal the reference](https://docs.python.org/3/c-api/tuple.html). That means that 1) nothing has changed relative to the old behavior in this PR, and 2) this incref is indeede necessary since we are retaining local references to `o` here and without the incref we will wind up undercounting the number of references alive.

python/kvikio/kvikio/_lib/arr.pyx
vyasr: What would that do? This is a C API function that does not push exceptions onto the stack. If it fails, it will return -1, so we could check the error value, but I chose not to do that in this PR since the same is true of the old functions (they could return NULL references, in which case we would have to pop the last exception).

jakirkham: Sorry for the slow follow up here

Have filed a fix for this in PR: https://github.com/rapidsai/kvikio/pull/728

Hopefully that makes it clearer what I meant

python/kvikio/kvikio/_lib/arr.pyx
jakirkham: Ok great. Thank you for checking! 🙏

Remember there being subtle behavior around stealing with these functions

Agree this seems reasonable then

PR #693, opened 2025-04-14
Add support for temporary AWS credentials to access S3

joosthooz:
This PR adds the `x-amz-security-token` header which is needed when using temporary credentials (These start with `ASIA`). According to the [curl docs](https://curl.se/libcurl/c/CURLOPT_HTTPHEADER.html), the list of headers needs to be freed after use, but `S3Endpoint` currently doesn't have a destructor so I'm not sure what to do with that.
I hope this is useful for others too, looking forward to any comments!

### DIFF
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -30,6 +30,7 @@
 #include <kvikio/error.hpp>
 #include <kvikio/parallel_operation.hpp>
 #include <kvikio/posix_io.hpp>
+#include <kvikio/shim/libcurl.hpp>
 #include <kvikio/utils.hpp>
 
 namespace kvikio {
@@ -92,6 +93,8 @@ class S3Endpoint : public RemoteEndpoint {
   std::string _url;
   std::string _aws_sigv4;
   std::string _aws_userpwd;
+  std::string _aws_token;
+  struct curl_slist *_curl_header_list;
 
   /**
    * @brief Unwrap an optional parameter, obtaining a default from the environment.
--- a/cpp/src/remote_handle.cpp
+++ b/cpp/src/remote_handle.cpp
@@ -29,7 +29,6 @@
 #include <kvikio/parallel_operation.hpp>
 #include <kvikio/posix_io.hpp>
 #include <kvikio/remote_handle.hpp>
-#include <kvikio/shim/libcurl.hpp>
 #include <kvikio/utils.hpp>
 
 namespace kvikio {
@@ -151,6 +150,13 @@ void S3Endpoint::setopt(CurlHandle& curl)
   curl.setopt(CURLOPT_URL, _url.c_str());
   curl.setopt(CURLOPT_AWS_SIGV4, _aws_sigv4.c_str());
   curl.setopt(CURLOPT_USERPWD, _aws_userpwd.c_str());
+  if (!_aws_token.empty()) {
+    _curl_header_list = curl_slist_append(NULL, _aws_token.c_str());
+    if (!_curl_header_list) {
+      throw std::runtime_error("Failed to create curl header for AWS token");
+    }
+    curl.setopt(CURLOPT_HTTPHEADER, _curl_header_list);
+  }
 }
 
 std::string S3Endpoint::unwrap_or_default(std::optional<std::string> aws_arg,
@@ -243,6 +249,16 @@ S3Endpoint::S3Endpoint(std::string url,
     ss << access_key << ":" << secret_access_key;
     _aws_userpwd = ss.str();
   }
+  if (access_key.compare(0, 4, std::string("ASIA")) == 0) {
+    char const* env = std::getenv("AWS_SESSION_TOKEN");
+    if (env == nullptr) {
+      throw std::invalid_argument("When using temporary credentials, AWS_SESSION_TOKEN must be set.");
+    }
+    auto token = std::string(env);
+    std::stringstream ss;
+    ss << "x-amz-security-token: " << token;
+    _aws_token = ss.str();
+  }
 }
 
 S3Endpoint::S3Endpoint(std::string const& bucket_name,
@@ -253,7 +269,7 @@ S3Endpoint::S3Endpoint(std::string const& bucket_name,
                        std::optional<std::string> aws_endpoint_url)
   : S3Endpoint(
       url_from_bucket_and_object(bucket_name, object_name, aws_region, std::move(aws_endpoint_url)),
-      std::move(aws_region),
+      aws_region,
       std::move(aws_access_key),
       std::move(aws_secret_access_key))
 {
@@ -272,7 +288,6 @@ RemoteHandle::RemoteHandle(std::unique_ptr<RemoteEndpoint> endpoint)
 {
   KVIKIO_NVTX_FUNC_RANGE();
   auto curl = create_curl_handle();
-
   endpoint->setopt(curl);
   curl.setopt(CURLOPT_NOBODY, 1L);
   curl.setopt(CURLOPT_FOLLOWLOCATION, 1L);


cpp/src/remote_handle.cpp
```diff
@@ -29,7 +29,6 @@
 #include <kvikio/parallel_operation.hpp>
 #include <kvikio/posix_io.hpp>
 #include <kvikio/remote_handle.hpp>
-#include <kvikio/shim/libcurl.hpp>
```
kingcrimsontianyu, line 32: This include can be kept, as we'd like to avoid transitive includes.

cpp/src/remote_handle.cpp
```diff
@@ -243,23 +244,43 @@ S3Endpoint::S3Endpoint(std::string url,
     ss << access_key << ":" << secret_access_key;
     _aws_userpwd = ss.str();
   }
+  // Create a Custom Curl header for the session token.
+  // The _curl_header_list created by curl_slist_append must be manually freed
+  // (see https://curl.se/libcurl/c/CURLOPT_HTTPHEADER.html)
+  if (access_key.compare(0, 4, std::string("ASIA")) == 0) {
```
kingcrimsontianyu, line 250: Why is there a condition that the access key has to equal `ASIA`?

cpp/src/remote_handle.cpp
```diff
@@ -243,23 +244,43 @@ S3Endpoint::S3Endpoint(std::string url,
...
+    auto session_token =
+      unwrap_or_default(std::move(aws_session_token),
+                        "AWS_SESSION_TOKEN",
+                        "When using temporary credentials, AWS_SESSION_TOKEN must be set.");
+    std::stringstream ss;
+    ss << "x-amz-security-token: " << session_token;
+    auto aws_token_header = ss.str();
+    _curl_header_list     = curl_slist_append(NULL, aws_token_header.c_str());
```
kingcrimsontianyu, line 258:
```suggestion
    std::stringstream ss;
    ss << "x-amz-security-token: " << session_token;
    _curl_header_list = curl_slist_append(NULL, ss.str().c_str());
```

cpp/src/remote_handle.cpp
bdice: See https://docs.aws.amazon.com/STS/latest/APIReference/API_GetAccessKeyInfo.html. The strings are either `AKIA` for long-term credentials or `ASIA` for short-term credentials. It doesn't seem to be a reference to the continent. :)

kingcrimsontianyu:
I think the Python part could be briefly updated in this PR. Currently we only deal with the required parameters `url`, `bucket_name` and `url_name` in Cython and leave all others as default. So the Python update could simply be a docstring update, somewhere around [here](https://github.com/rapidsai/kvikio/blob/branch-25.06/python/kvikio/kvikio/remote_file.py#L83) and [here](https://github.com/rapidsai/kvikio/blob/branch-25.06/python/kvikio/kvikio/remote_file.py#L117).

We may consider the support for programatically setting the optional aws-related parameters in Python in a future PR if there is a need (in addition to the current env var approach).

cpp/src/remote_handle.cpp
kingcrimsontianyu: Could we add this quote from the reference for clarity:
```
  // Access key IDs beginning with ASIA are temporary credentials that are created using AWS STS
  // operations.
```

joosthooz: If the session key starts with `ASIA` (as opposed to `AKIA`), it means it is a temporary key and needs a session token. I should have added a comment about this, I will add it.

cpp/src/remote_handle.cpp
joosthooz: I added that intermediary variable because of this discussion https://stackoverflow.com/questions/21034834/is-there-issue-will-stringstream-str-c-str but I'm not an expert in these C++ subtleties so if it is not needed I'll gladly apply your suggestion!

joosthooz:
Thank you for the comments @kingcrimsontianyu, I will address them tomorrow!

cpp/src/remote_handle.cpp
kingcrimsontianyu: The `ss.str()` method returns a copy of the string, which is a temporary object. It is [destroyed at the end of a full expression](https://eel.is/c++draft/class.temporary#4), i.e. at the end of line 257. If `curl` internally only uses a `const char*` to reference `ss.str().c_str()`, then after line 257 the pointer will become dangling, and any dereference afterwards would result in UB (likely segfault). Fortunately, `curl_slist_append` [doc](https://curl.se/libcurl/c/curl_slist_append.html) states that:

>curl_slist_append copies the string.

In other words, it internally creates a deep copy of the string that is guaranteed to remain valid after line 257, making it safe to pass a temporary as the second argument to `curl_slist_append`.

joosthooz: Great, thanks for the additional insight

kingcrimsontianyu:
@joosthooz I'll add a few more suggestions in a couple of hours to fix the CI issue. Thank you!

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -30,6 +30,7 @@
 #include <kvikio/error.hpp>
 #include <kvikio/parallel_operation.hpp>
 #include <kvikio/posix_io.hpp>
+#include <kvikio/shim/libcurl.hpp>
```
kingcrimsontianyu, line 33: The shim header should be removed to fix the CI.
```suggestion
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -92,6 +93,7 @@ class S3Endpoint : public RemoteEndpoint {
   std::string _url;
   std::string _aws_sigv4;
   std::string _aws_userpwd;
+  struct curl_slist* _curl_header_list;
```
kingcrimsontianyu, line 96: This member needs initialization. Otherwise, if `aws_session_token` is not given, the statement `if (_curl_header_list)` is UB.
```suggestion
  curl_slist* _curl_header_list{};
```

cpp/include/kvikio/remote_handle.hpp
kingcrimsontianyu: This means `curl_slist` needs to be forward declared, e.g.:
```
#include <kvikio/utils.hpp>

struct curl_slist;

namespace kvikio {
```

cpp/src/remote_handle.cpp
```diff
@@ -243,23 +245,44 @@ S3Endpoint::S3Endpoint(std::string url,
...
+    auto session_token =
+      unwrap_or_default(std::move(aws_session_token),
+                        "AWS_SESSION_TOKEN",
+                        "When using temporary credentials, AWS_SESSION_TOKEN must be set.");
+    std::stringstream ss;
+    ss << "x-amz-security-token: " << session_token;
+    _curl_header_list = curl_slist_append(NULL, ss.str().c_str());
+    if (!_curl_header_list) {
+      throw std::runtime_error("Failed to create curl header for AWS token");
+    }
```
kingcrimsontianyu, line 263:
```suggestion
    _curl_header_list = curl_slist_append(nullptr, ss.str().c_str());
    KVIKIO_EXPECT(_curl_header_list != nullptr,
                  "Failed to create curl header for AWS token",
                  std::runtime_error);
```

cpp/include/kvikio/remote_handle.hpp
joosthooz: It's needed for the `struct curl_slist` field, any idea how to get around this?

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -180,11 +185,12 @@ class S3Endpoint : public RemoteEndpoint {
              std::optional<std::string> aws_region            = std::nullopt,
              std::optional<std::string> aws_access_key        = std::nullopt,
              std::optional<std::string> aws_secret_access_key = std::nullopt,
+             std::optional<std::string> aws_session_token     = std::nullopt,
```
bdice, line 188: Can we add the new parameter to the end of the argument list, to avoid an API break? The types of `aws_session_token` and `aws_endpoint_url` are identical, so users of the old API will now be passing an endpoint as if it was a session token.

bdice (changes requested):

cpp/include/kvikio/remote_handle.hpp
joosthooz: Done

cpp/include/kvikio/remote_handle.hpp
joosthooz: Done

cpp/src/remote_handle.cpp
joosthooz: Done

cpp/include/kvikio/remote_handle.hpp
joosthooz: I've made the change. It would be desirable to keep them grouped together but a hidden breaking API change sounds troublesome.

cpp/include/kvikio/remote_handle.hpp
kingcrimsontianyu: Thanks. Could you also remove `#include <kvikio/shim/libcurl.hpp>` from `remote_handle.hpp` as this would cause CI issue? 
`#include <kvikio/shim/libcurl.hpp>` transitively includes `<curl/curl.h>`. We want to avoid exposing `<curl/curl.h>` in KvikIO's headers. That's why we use the shim and also make forward declaration of `curl_slist`.

cpp/include/kvikio/remote_handle.hpp
joosthooz: Turns out I only changed the order in the .cpp file, should be fixed now! Also added missing documentation for the new parameter to that function.

kingcrimsontianyu (changes requested):
Please address https://github.com/rapidsai/kvikio/pull/693/files#r2063590434 so as to fix the CI issue.

joosthooz:
Sorry @kingcrimsontianyu , I was certain I had removed that include but it turns out I didn't! Should be good now.

kingcrimsontianyu (approved):
Thanks for raising the issue and coming up with a solution!

bdice (approved):

davidwendt:
Looks like this may have broken the cudf build. 
@kingcrimsontianyu

kingcrimsontianyu:
Could you elaborate a bit? I'm not able to reproduce the build fail locally with CUDA 12.8 pip cuDF.

PR #697, opened 2025-04-21
Improve unit tests; add a dedicated macro for Linux system calls

kingcrimsontianyu:
- This PR reorganizes the C++ unit tests to put functionally independent tests in separate binaries. Note that the boilerplate `main.cpp` file is removed and its equivalent replacement is a reference to `GTest::gtest_main` in CMake. If in the future we need a main function with custom logic, we will add the file back.
- This PR also adds a dedicated macro `SYSCALL_CHECK(ret)` for Linux system calls as a replacement for the general assertion `KVIKIO_EXPECT(ret != -1, "extra_msg", GenericSystemError);`


cpp/src/error.cpp
```diff
@@ -34,6 +35,20 @@ void log_error(std::string_view err_msg, int line_number, char const* filename)
   std::cerr << "KvikIO error at: " << filename << ":" << line_number << ": " << err_msg << "\n";
 }
 
+void handle_linux_call_error(int line_number, char const* filename, std::string_view extra_msg)
+{
+  auto err_name = strerrorname_np(errno);
```
bdice, line 41: Do we need to check if the compiler is GNU with some kind of macro? This is a GNU extension, according to https://www.gnu.org/software/libc/manual/html_node/Error-Messages.html. Please check what happens with, e.g. clang.

Also we probably need to `#include <string.h>` to use `strerrorname_np`.

cpp/tests/test_error.cpp
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#include <fcntl.h>
+#include <sys/mman.h>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <kvikio/error.hpp>
+#include <kvikio/file_handle.hpp>
+#include "gmock/gmock.h"
```
bdice, line 24: This is included twice, once with "" and once with <>.

cpp/src/error.cpp
kingcrimsontianyu: Good catch. Clang supports `_GNU_SOURCE` but I'm hesitant to continue using this non-POSIX/C++-standard feature now. Scrapped related code.

bdice (approved):

cpp/tests/test_error.cpp
```diff
@@ -0,0 +1,60 @@
...
+    int fd             = -1;
+    off_t offset       = 0;
+    return mmap(nullptr, length, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, fd, offset);
+  };
+
+  // If the mapping fails, mmap() returns (void*)-1, and the error number is EINVAL (invalid
+  // argument).
+  EXPECT_THAT(
+    [=] {
+      SYSCALL_CHECK(map_anonymous_with_0_length(), "mmap failed.", reinterpret_cast<void*>(-1));
```
vuule, line 57: could we just do this?
```suggestion
      SYSCALL_CHECK(map_anonymous_with_0_length(), "mmap failed.", MAP_FAILED);
```

vuule (commented):
looks good, just one possible suggestion

cpp/tests/test_error.cpp
kingcrimsontianyu: Thanks. Done.

vuule (approved):

PR #700, opened 2025-04-22
Fix the S3 endpoint constructor ambiguity

### LINKED ISSUE #688: Ambiguity in the S3Endpoint constructors
The S3Endpoint has the following constructors:

```
  S3Endpoint(std::string url,
             std::optional<std::string> aws_region            = std::nullopt,
             std::optional<std::string> aws_access_key        = std::nullopt,
             std::optional<std::string> aws_secret_access_key = std::nullopt);

  S3Endpoint(std::string const& bucket_name,
             std::string const& object_name,
             std::optional<std::string> aws_region            = std::nullopt,
             std::optional<std::string> aws_access_key        = std::nullopt,
             std::optional<std::string> aws_secret_access_key = std::nullopt,
             std::optional<std::string> aws_endpoint_url      = std::nullopt);
```
But if you try to use the first by just passing strings (and relying on the compiler to create optionals from them), it can happen that it inadvertently uses the second and fills in the "missing" arguments with the default nullopt. So to be sure you're using the first, you need to explicitly create a `std::optional<std::string>` for the `aws_region` argument.

kingcrimsontianyu:
This PR closes #688 .
This PR also introduces to unit tests the utility RAII class `EnvVarContext` to enable temporarily setting environment variables to new values for a test and then restoring to previous states afterwards.


cpp/include/kvikio/remote_handle.hpp
vuule, line 1: Another way to resolve this would be to add a url class that can be constructed from a string (explicitly!). Then the call site becomes url{my_string}, and the other ctor does not need to change.
This is not obviously better to me, just bringing it up for consideration/discussion.

cpp/tests/utils/env.hpp
```diff
@@ -0,0 +1,65 @@
...
+   */
+  ~EnvVarContext();
+
+  EnvVarContext(EnvVarContext const&)            = delete;
+  EnvVarContext(EnvVarContext&&)                 = delete;
+  EnvVarContext& operator=(EnvVarContext const&) = delete;
+  EnvVarContext& operator=(EnvVarContext&&)      = delete;
+
+ private:
+  std::unordered_map<std::string, EnvVarState> _env_var_map;
```
vuule, line 63: we never look up entries, AFAICT this could be a vector of pairs

cpp/tests/utils/env.cpp
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+#include "env.hpp"
+#include <cstdlib>
+#include <unordered_map>
+
+#include <kvikio/error.hpp>
+
+namespace kvikio::test {
+EnvVarContext::EnvVarContext(std::unordered_map<std::string, std::string> const& env_var_entries)
```
vuule, line 24: sane here, no need to take a map

cpp/tests/utils/env.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <string>
+#include <unordered_map>
+
+namespace kvikio::test {
+/**
+ * @brief RAII class used to temporarily set environment variables to new values upon construction,
+ * and restore them to previous values upon destruction.
+ */
+class EnvVarContext {
```
vuule, line 27: neat! I might borrow this for cudf compression tests, they are getting to a point where more than one env var is being set

cpp/tests/utils/env.cpp
```diff
@@ -0,0 +1,48 @@
...
+{
+  for (auto const& [key, current_value] : env_var_entries) {
```
vuule, line 26: do you know we'd ever need to unset an env var instead of setting to the specified value?

vuule (commented):
minor questions/suggestions

cpp/tests/utils/env.hpp
kingcrimsontianyu: Good catch. My initial intention was to check for duplicate keys but I forgot to add the check. Added the logic.
Also to avoid UB from temporaries getting destroyed right after the constructor, I replaced const reference with initializer list as the constructor parameter. The unit test is updated accordingly.

cpp/tests/utils/env.cpp
kingcrimsontianyu: Replaced with an initializer list.

cpp/tests/utils/env.cpp
kingcrimsontianyu: I was thinking of [CUDF_BENCHMARK_DROP_CACHE](https://github.com/rapidsai/cudf/blob/05646df0b8dd4b69f7cfed6fbf9e882df795210c/cpp/benchmarks/io/cuio_common.cpp#L213). No matter what this env var's value is, as long as it exists, the page cache is to be dropped (i.e. `CUDF_BENCHMARK_DROP_CACHE=<nothing>/true/false/123` all have the same effect). In this case to fully revert to a previous state, we need to unset the env var.

cpp/tests/utils/env.hpp
kingcrimsontianyu: PS: My concern over UB is not necessary since the member `_env_var_map` is not a reference to the constructor parameter `env_var_entries`.

cpp/tests/utils/env.cpp
vuule: I'm not referring to reverting the state, but the actual test state requiring an env var to be unset.

cpp/tests/utils/env.hpp
vuule: IMO it's fine in tests, it's not a part of the functionality of the kvikIO library

cpp/tests/utils/env.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <kvikio/error.hpp>
+
+namespace kvikio::test {
+EnvVarContext::EnvVarContext(
+  std::initializer_list<std::pair<std::string, std::string>> env_var_entries)
+{
+  for (auto const& [key, current_value] : env_var_entries) {
+    EnvVarState env_var_state;
+    auto res = std::getenv(key.c_str());
+    if (res) {
```
vuule, line 31:
```suggestion
    if (auto const res = std::getenv(key.c_str()); res != nullptr) {
```

vuule (approved):
looks good; just one small suggestion

cpp/tests/utils/env.cpp
kingcrimsontianyu: Yes. I think a thorough test of S3 endpoint would need this feature. The constructor of S3 endpoint has "optional" parameters, but if they are neither passed as argument or set via the env var, [an exception will be thrown](https://github.com/kingcrimsontianyu/kvikio/blob/branch-25.06/cpp/src/remote_handle.cpp#L217). We may want to check if the constructor can throw properly when some arguments are missing in one test, and check if another combination of arguments are missing in another test. This is where env var context could help.

cpp/tests/utils/env.cpp
kingcrimsontianyu: Thanks! Done.

madsbk (approved):
Looks good!

PR #702, opened 2025-04-25
refactor(wheels): vendor nvcomp as temporary workaround

gforsyth:
xref rapidsai/build-planning#120

RAPIDS is blocked on Python 3.13 support because there is no `nvcomp` release that (yet) supports Python 3.13.

This is a temporary (and wacky) workaround that re-vendors `nvcomp` into
`libkvikio` and removes the runtime dependency on the `nvcomp` wheel, removing
with it the block on Python 3.13.

Once there's an updated `nvcomp` release, we can undo this.

Notes:

I am no kind of CMake expert, and I'm sure this can be done better.

Using `kvikio-exports` as the `BUILD_EXPORT_SET` is no good because it installs
3+ copies of the same 100mb shared object and balloons the size of the wheel.
There are _two_ copies of `libnvcomp.so` as it stands, but they compress down
well enough.  If anyone has any pointers on removing the remaining `so`, I'm
all ears.  I know it's possible because the `nvcomp` wheel only has a single
`libnvcomp.so.4`, but also, this is a temporary patch and it's good enough.

I've added in the `WheelHelpers.cmake` module because it does what I want, but
if there's something simpler, we can try that, too.

### DIFF
--- a/ci/build_wheel_cpp.sh
+++ b/ci/build_wheel_cpp.sh
@@ -25,13 +25,12 @@ rapids-pip-retry install \
 # 0 really means "add --no-build-isolation" (ref: https://github.com/pypa/pip/issues/5735)
 export PIP_NO_BUILD_ISOLATION=0
 
-export SKBUILD_CMAKE_ARGS="-DUSE_NVCOMP_RUNTIME_WHEEL=ON"
+export SKBUILD_CMAKE_ARGS="-DBUILD_SHARED_LIBS=ON;-DUSE_NVCOMP_RUNTIME_WHEEL=OFF;-DCMAKE_MESSAGE_LOG_LEVEL=VERBOSE;-DCUDA_STATIC_RUNTIME=ON"
 ./ci/build_wheel.sh "${package_name}" "${package_dir}"
 
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen "${RAPIDS_CUDA_VERSION}")"
 
 python -m auditwheel repair \
-    --exclude libnvcomp.so.4 \
     -w "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}" \
     ${package_dir}/dist/*
 
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -66,7 +66,9 @@ files:
       table: project
     includes:
       - depends_on_cupy
-      - depends_on_nvcomp
+      # TODO: restore runtime dependency when we no longer vender nvcomp (when
+      # nvcomp supports Python 3.13)
+      # - depends_on_nvcomp
       - depends_on_libkvikio
       - run
   py_rapids_build_libkvikio:
--- a/python/libkvikio/CMakeLists.txt
+++ b/python/libkvikio/CMakeLists.txt
@@ -41,7 +41,17 @@ set(KvikIO_BUILD_EXAMPLES OFF)
 set(KvikIO_BUILD_TESTS OFF)
 if(USE_NVCOMP_RUNTIME_WHEEL)
   set(KvikIO_EXPORT_NVCOMP OFF)
+else()
+  # vendor nvcomp but not the entire kvikio-export set because that's huge
+  include(cmake/thirdparty/get_nvcomp.cmake)
+  include(cmake/Modules/WheelHelpers.cmake)
+  # Set the output directory so libnvcomp.so ends up in the right place
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${SKBUILD_PLATLIB_DIR}/libkvikio/lib64/)
+  install_aliased_imported_targets(
+    TARGETS nvcomp::nvcomp DESTINATION ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+  )
 endif()
+
 set(CUDA_STATIC_RUNTIME ON)
 
 add_subdirectory(../../cpp kvikio-cpp)
--- a/python/libkvikio/cmake/thirdparty/get_nvcomp.cmake
+++ b/python/libkvikio/cmake/thirdparty/get_nvcomp.cmake
@@ -1,0 +1,35 @@
+# =============================================================================
+# Copyright (c) 2021-2024, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+# =============================================================================
+
+set(KVIKIO_USE_PROPRIETARY_BINARY ON)
+
+# This function finds nvcomp and sets any additional necessary environment variables.
+function(find_and_configure_nvcomp)
+
+  include(${rapids-cmake-dir}/cpm/nvcomp.cmake)
+  set(export_args)
+  if(KvikIO_EXPORT_NVCOMP)
+    # We're vendoring nvcomp and we only want `libnvcomp.so.4`
+    set(export_args BUILD_EXPORT_SET nvcomp)
+  endif()
+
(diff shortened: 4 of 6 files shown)


python/libkvikio/cmake/thirdparty/get_nvcomp.cmake
```diff
@@ -0,0 +1,35 @@
...
+set(KVIKIO_USE_PROPRIETARY_BINARY ON)
+
+# This function finds nvcomp and sets any additional necessary environment variables.
+function(find_and_configure_nvcomp)
+
+  include(${rapids-cmake-dir}/cpm/nvcomp.cmake)
+  set(export_args)
+  if(KvikIO_EXPORT_NVCOMP)
+    # We're vendoring nvcomp and we only want `libnvcomp.so.4`
+    set(export_args BUILD_EXPORT_SET nvcomp)
```
bdice, line 24: Should this be `INSTALL_EXPORT_SET`? (I don't know for sure.)
```suggestion
    set(export_args INSTALL_EXPORT_SET nvcomp)
```

gforsyth: I don't think so -- I can't say I really understand the difference, but `INSTALL_EXPORT_SET` balloons the size of the resulting wheel and includes many more copies of the nvcomp SO:

```
checking '/tmp/wheelhouse/libkvikio_cu12-25.6.0a23-py3-none-manylinux_2_28_x86_64.whl'
----- package inspection summary -----
file size
  * compressed size: 0.2G
  * uncompressed size: 0.5G
  * compression space saving: 67.8%
contents
  * directories: 22
  * files: 142 (7 compiled)
size by extension
  * .a - 0.1G (26.2%)
  * .so - 0.1G (25.1%)
  * .4 - 0.1G (24.3%)
  * .11 - 0.1G (24.3%)
  * .h - 0.6M (0.1%)
  * .hpp - 0.4M (0.1%)
  * no-extension - 63.4K (0.0%)
  * .cmake - 63.3K (0.0%)
  * .cuh - 12.4K (0.0%)
  * .py - 4.7K (0.0%)
  * .txt - 38.0B (0.0%)
largest files
  * (0.1G) libkvikio/lib64/libnvcomp_static.a
  * (0.1G) libkvikio/lib64/libnvcomp.so
  * (0.1G) libkvikio/lib64/libnvcomp.so.4
  * (0.1G) libkvikio/lib64/libnvcomp.so.4.2.0.11
  * (4.2M) libkvikio/lib64/libkvikio.so
------------ check results -----------
1. [distro-too-large-compressed] Compressed size 0.2G is larger than the allowed size (90.0M).
```

ci/build_wheel_cpp.sh
```diff
@@ -25,13 +25,12 @@ rapids-pip-retry install \
 # 0 really means "add --no-build-isolation" (ref: https://github.com/pypa/pip/issues/5735)
 export PIP_NO_BUILD_ISOLATION=0
 
-export SKBUILD_CMAKE_ARGS="-DUSE_NVCOMP_RUNTIME_WHEEL=ON"
+export SKBUILD_CMAKE_ARGS="-DBUILD_SHARED_LIBS=ON;-DUSE_NVCOMP_RUNTIME_WHEEL=OFF;-DCMAKE_MESSAGE_LOG_LEVEL=VERBOSE;-DCUDA_STATIC_RUNTIME=ON"
```
bdice, line 28: Do we need to change all of these? Or just `-DUSE_NVCOMP_RUNTIME_WHEEL`? The verbose logging one seems safe to remove.
```suggestion
export SKBUILD_CMAKE_ARGS="-DBUILD_SHARED_LIBS=ON;-DUSE_NVCOMP_RUNTIME_WHEEL=OFF;-DCUDA_STATIC_RUNTIME=ON"
```

dependencies.yaml
```diff
@@ -66,7 +66,9 @@ files:
       table: project
     includes:
       - depends_on_cupy
-      - depends_on_nvcomp
+      # TODO: restore runtime dependency when we no longer vendor nvcomp
+      # (when nvcomp supports Python 3.13)
```
bdice, line 70: We should probably have a tracking issue for this.

python/libkvikio/libkvikio/load.py
```diff
@@ -44,12 +44,17 @@ def _load_wheel_installation(soname: str):
 
 
 def load_library():
+    # TODO: remove this nvcomp load when `nvcomp` is re-de-vendored
```
bdice, line 47: Link to a tracking issue (noted above).

bdice (approved):
This looks fine to me! I would like a second pair of eyes from @jameslamb, @KyleFromNVIDIA, @vyasr, or @robertmaynard.

We should also test this downstream to make sure that libcudf wheels can find this library in the libkvikio wheel, without pulling in nvcomp wheels themselves.

The issue with packaging multiple copies of libnvcomp.so is expected. Wheels don't properly support symlinks, and there is a draft PEP for that. It's partly motivated by SOVERSION symlinks like what we have here. See: https://discuss.python.org/t/pep-778-supporting-symlinks-in-wheels/53824

This is totally fine:
```bash
$ ls -alh libkvikio/lib64/
total 259M
drwxrwxrwx 1 bdice bdice 4.0K Apr 28 15:14 ./
drwxrwxrwx 1 bdice bdice 4.0K Apr 28 15:14 ../
drwxrwxrwx 1 bdice bdice 4.0K Apr 28 15:14 cmake/
-rwxrwxrwx 1 bdice bdice 4.3M Apr 28 15:14 libkvikio.so*
-rwxrwxrwx 1 bdice bdice 128M Apr 28 15:14 libnvcomp.so.4*
-rwxrwxrwx 1 bdice bdice 128M Apr 28 15:14 libnvcomp.so.4.2.0.11*
```

dependencies.yaml
gforsyth: Added!

bdice (approved):
Amazing. Please test downstream in cudf and then this is good to merge.

python/libkvikio/CMakeLists.txt
```diff
@@ -41,7 +41,17 @@ set(KvikIO_BUILD_EXAMPLES OFF)
 set(KvikIO_BUILD_TESTS OFF)
 if(USE_NVCOMP_RUNTIME_WHEEL)
   set(KvikIO_EXPORT_NVCOMP OFF)
+else()
+  # vendor nvcomp but not the entire kvikio-export set because that's huge
+  include(cmake/thirdparty/get_nvcomp.cmake)
+  include(cmake/Modules/WheelHelpers.cmake)
+  # Set the output directory so libnvcomp.so ends up in the right place
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${SKBUILD_PLATLIB_DIR}/libkvikio/lib64/)
```
vyasr, line 49: Since you're manually installing anyway you shouldn't need to use a CMake reserved keyword to specify this directory. Presumably we have disabled normal install rules and are only installing manually via `install_aliased_imported_targets`.

python/libkvikio/cmake/thirdparty/get_nvcomp.cmake
vyasr: Effectively by setting `INSTALL_EXPORT_SET` you are causing all of nvcomp's install rules to be activated, whereas by not setting and using the wheelhelpers function you are just copying in the binaries.

python/libkvikio/cmake/thirdparty/get_nvcomp.cmake
```diff
@@ -0,0 +1,35 @@
...
+  endif()
+
+  rapids_cpm_nvcomp(${export_args} USE_PROPRIETARY_BINARY ${KVIKIO_USE_PROPRIETARY_BINARY})
+
+  # Per-thread default stream
+  if(TARGET nvcomp AND PER_THREAD_DEFAULT_STREAM)
```
vyasr, line 30: I don't see `PER_THREAD_DEFAULT_STREAM` used or documented anywhere else. It's not strictly relevant to this PR, but should we go ahead and add an explicit top-level option for this?

vyasr (approved):
Couple of notes but nothing blocking. I don't need to review again. Thanks!

gforsyth:
Tests are passing downstream in https://github.com/rapidsai/cudf/pull/18593 so I'm going to merge this, then update https://github.com/rapidsai/cudf/pull/18593 to only include the changes needed to point at the vendored `nvcomp` and not the wheel

Issue #701, opened 2025-04-22
Skip benchmark memory checks on MIG


jakirkham:
In supporting [MIG]( https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html ) in UCX-Py previously ( https://github.com/rapidsai/ucx-py/pull/838 ), we skipped NVML memory checks on MIG to avoid errors. UCXX [also implements this behavior]( https://github.com/rapidsai/ucxx/blob/7ebd12007a0c443fb5bbc6c3c49c748fcf365803/python/ucxx/ucxx/__init__.py#L83-L88 )

KvikIO benchmarks also query memory info to display to the user. However this check is not skipped for MIG

https://github.com/rapidsai/kvikio/blob/58e650e3bb1b6e91e376f144d17c648f210db524/python/kvikio/kvikio/benchmarks/utils.py#L42-L44

Would suggest we implement the same MIG check in KvikIO and skip these memory checks on MIG

PR #703, opened 2025-04-26
Fix benchmark code on MIG

jakirkham:
In UCX-Py and UCXX, we skip querying memory info on MIG devices as the approach we use on other hardware doesn't work for MIG. However KvikIO did not skip querying memory info on MIG, which causes this code to fail there.

To fix this issue, we adapt the approach used in UCX-Py and UCXX to KvikIO, which allows this code to run on MIG.

Additionally the code used one `try...except...` to handle everything from `import`ing to querying the device. With the MIG check too, this becomes kind of unwieldy for one `try...except...`. So this moves the `import` to the top-level and breaks up the `try...except...`. This makes the logic of the code and the cases it handles a bit more legible to readers 

Fixes https://github.com/rapidsai/kvikio/issues/701

### DIFF
--- a/python/kvikio/kvikio/benchmarks/utils.py
+++ b/python/kvikio/kvikio/benchmarks/utils.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import argparse
+import contextlib
 import os
 import os.path
 import pathlib
@@ -14,6 +15,12 @@ from dask.utils import format_bytes
 import kvikio
 import kvikio.cufile_driver
 import kvikio.defaults
+
+has_pynvml = False
+with contextlib.suppress(ImportError):
+    import pynvml
+
+    has_pynvml = True
 
 
 def drop_vm_cache() -> None:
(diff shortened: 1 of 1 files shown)


python/kvikio/kvikio/benchmarks/utils.py
```diff
@@ -15,6 +16,12 @@
 import kvikio.cufile_driver
 import kvikio.defaults
 
+has_pynvml = False
+with contextlib.suppres(ImportError):
```
bdice, line 20: Typo?
```suggestion
with contextlib.suppress(ImportError):
```

jakirkham: Yes thanks Bradley! 🙏

python/kvikio/kvikio/benchmarks/utils.py
```diff
@@ -15,6 +16,12 @@
 import kvikio.cufile_driver
 import kvikio.defaults
 
+has_pynvml = False
+with contextlib.suppress(ImportError):
+    import pynvml
+
+    has_pynvml = True
+
```
madsbk, line 24: Consider moving this into `pprint_sys_info()`, as `pprint_sys_info()` isn't performance critical.

python/kvikio/kvikio/benchmarks/utils.py
```diff
@@ -29,19 +36,28 @@ def pprint_sys_info() -> None:
...
+            pynvml.nvmlInit()
+            dev = pynvml.nvmlDeviceGetHandleByIndex(0)
+
+        if dev is not None:
+            is_mig = False
+            with contextlib.suppress(pynvml.NVMLError):
+                pynvml.nvmlDeviceGetMigMode(dev)
+                is_mig = True
+
+            if not is_mig:
```
madsbk, line 53: Are we sure `nvmlDeviceGetMigMode` exist? Maybe to this:
```suggestion
            try:
                pynvml.nvmlDeviceGetMigMode(dev)
            except pynvml.NVMLError, AttributeError:
                pass
            else:
```

jakirkham: Yep the function is [always defined]( https://github.com/gpuopenanalytics/pynvml/blob/82cb5e08f56da105a63ca9fcdeb04335bc90950e/pynvml/nvml.py#L3942-L3948 ). Same is true of `nvidia-ml-py`

python/kvikio/kvikio/benchmarks/utils.py
jakirkham: Had put this here as [top-level `import`s are preferred]( https://peps.python.org/pep-0008/#imports ) and it helps keep the code in `pprint_sys_info` a bit more manageable (given other checks occur there)

jakirkham:
Checked both `pynvml` `11.5.3` and `nvidia-ml-py` `12.570.86`. Both defined `nvmlDeviceGetMigMode`

This appears consistent with the C function [`nvmlDeviceGetMigMode`]( https://docs.nvidia.com/deploy/nvml-api/group__nvmlMultiInstanceGPU.html#group__nvmlMultiInstanceGPU_1g1fa0acf9076404ef28c3c6976eb96f2b ), which returns success or an error. One of the errors being `NVML_ERROR_NOT_SUPPORTED`, which would be used on non-MIG devices

Additionally had tested on a MIG & non-MIG device as part of writing this code and both behaved as expected

python/kvikio/kvikio/benchmarks/utils.py
madsbk: Fair enough, then I would do:
```suggestion
            try:
                pynvml.nvmlDeviceGetMigMode(dev)
            except pynvml.NVMLError:
                pass
            else:
```

madsbk (approved):
Other then the nitpick, I think this looks good. Thanks @jakirkham

python/kvikio/kvikio/benchmarks/utils.py
jakirkham: So I gave this a try. Unfortunately it appears to skip the memory check on non-MIG machines (like a DGX) as well

Looks like we need this change instead

```suggestion
            try:
                pynvml.nvmlDeviceGetMigMode(dev)
            except pynvml.NVMLError:
```

madsbk:
Thanks @jakirkham

jakirkham:
Thanks Mads! 🙏

PR #704, opened 2025-04-26
Add the feature to check page cache residency information for a file

kingcrimsontianyu:
This PR provides a utility function `get_page_cache_info(file)` in C++ and Python to check how many pages of a file are resident in the page cache of the Linux kernel.

This PR is a byproduct of a recent investigation of an issue on GH200 system, where the sudo command that drops the page cache doesn't seem to achieve the expected effect.

This feature is functionally similar to the Linux utility program `fincore`, which is also based on the system call `mincore` used in this PR.

### DIFF
--- a/python/kvikio/kvikio/_lib/file_handle.pyx
+++ b/python/kvikio/kvikio/_lib/file_handle.pyx
@@ -178,7 +178,7 @@ cdef class CuFile:
             stream,
         ))
 
-cdef extern from "<kvikio/file_utils.hpp>" namespace "kvikio" nogil:
+cdef extern from "<kvikio/file_utils.hpp>" nogil:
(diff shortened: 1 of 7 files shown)


python/kvikio/kvikio/_lib/file_handle.pyx
```diff
@@ -175,3 +177,28 @@ cdef class CuFile:
             dev_offset,
             stream,
         ))
+
+cdef extern from "<kvikio/file_utils.hpp>" namespace "kvikio" nogil:
```
madsbk, line 181:
```suggestion
cdef extern from "<kvikio/file_utils.hpp>" nogil:
```

madsbk (approved):

PR #714, opened 2025-05-07
Deprecate nvCOMP python API

### LINKED ISSUE #685: Evaluate proper long-term home for nvcomp bindings
Currently kvikio contains Cython bindings to the nvcomp C++ library. Now that nvcomp has [an official Python API](https://docs.nvidia.com/cuda/nvcomp/py_api.html) we should consider eventually moving to use that if necessary rather than maintaining our own. However, it is also not clear to me if kvikio needs that at all. kvikio was historically the sole place to access nvcomp APIs from Python, so rather than using nvcomp APIs we were primarily a _source_ for those APIs in other libraries. If that is still the case, we could consider deprecating and removing those bits altogether.

kingcrimsontianyu:
Closes #685 

This PR marks nvCOMP python bindings deprecated.

Specifically, this PR adds a free function `kvikio_deprecate_module` used to deprecate modules. This PR also improves the decorator `@kvikio_deprecation_notice`, which now has the following two effects on functions:

- The deprecation notice is shown at runtime when the decorated function is called.
- The deprecation notice is shown on the Sphinx documentation (see examples below).


kingcrimsontianyu:
## Example: deprecation notice on Sphinx doc

image

madsbk (approved):
LGTM, thanks @kingcrimsontianyu

python/kvikio/kvikio/defaults.py
```diff
@@ -181,11 +181,9 @@ def is_compat_mode_preferred() -> bool:
     return kvikio._lib.defaults.is_compat_mode_preferred()
 
 
-@kvikio_deprecation_notice('Use kvikio.defaults.get("compat_mode") instead')
+@kvikio_deprecation_notice("Use :meth:`kvikio.defaults.get()` instead.", "25.04")
```
bdice, line 184: Deprecations would take place in 25.06, followed by removal in 25.08 at the earliest.
```suggestion
@kvikio_deprecation_notice("Use :meth:`kvikio.defaults.get()` instead.", "25.06")
```

python/kvikio/kvikio/utils.py
```diff
@@ -131,8 +133,12 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def kvikio_deprecation_notice(msg: str) -> Callable:
-    """Decorate a function to print the deprecation notice at runtime.
+def kvikio_deprecation_notice(
+    msg: str = "This function is deprecated.",
+    deprecated_since_version: str = __version__,
```
bdice, line 138: Make this a keyword-only argument and simplify the name. It will make the call sites more understandable.
```suggestion
    *, since: str = __version__,
```

python/kvikio/kvikio/utils.py
```diff
@@ -146,6 +152,9 @@ def kvikio_deprecation_notice(msg: str) -> Callable:
     msg: str
         The deprecation notice.
 
+    deprecated_since_version: str
```
bdice, line 155:
```suggestion
    since: str
```

python/kvikio/kvikio/utils.py
```diff
@@ -157,8 +166,29 @@ def wrapper(*args, **kwargs):
             warnings.warn(msg, category=FutureWarning, stacklevel=2)
             return func(*args, **kwargs)
 
-        # Allow the docstring to be corrected generated for the decorated func in Sphinx
-        wrapper.__doc__ = func.__doc__
+        # Allow the docstring to be correctly generated for the decorated func in Sphinx
+        valid_docstring = "" if func.__doc__ is None else func.__doc__
```
bdice, line 170:
```suggestion
        valid_docstring = getattr(func, "__doc__") or ""
```

cpp/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ option(KvikIO_BUILD_EXAMPLES "Configure CMake to build examples" ON)
 option(KvikIO_BUILD_TESTS "Configure CMake to build tests" ON)
 option(KvikIO_REMOTE_SUPPORT "Configure CMake to build with remote IO support" ON)
 option(KvikIO_CUDA_SUPPORT "Configure CMake to build with CUDA support" ON)
-option(KvikIO_EXPORT_NVCOMP "Export NVCOMP as a dependency" ON)
+option(KvikIO_EXPORT_NVCOMP "(Deprecated) Export NVCOMP as a dependency" ON)
```
bdice, line 47: If we deprecate this, will we also stop vendoring libnvcomp inside libkvikio wheels?

bdice (commented):
A few suggestions.

Also do we care how will this impact Python 3.13 users since nvCOMP does not yet provide bindings for Python 3.13?

cpp/CMakeLists.txt
kingcrimsontianyu: Thanks. I marked this option deprecated before realizing that we do wish to keep the use of nvcomp for the Zarr store feature. So I think I should revert this change here.

python/kvikio/kvikio/utils.py
kingcrimsontianyu: Thanks. Good to know the built-in `getattr`. According to this [SO answer](https://stackoverflow.com/a/19213583/1587802) and the referenced [PEP 308](https://peps.python.org/pep-0308), it seems the if-else one liner is still preferred.

python/kvikio/kvikio/defaults.py
kingcrimsontianyu: Actually. This part of the code in `defaults.py` has been deprecated since the `25.04` release.

python/kvikio/pyproject.toml
```diff
@@ -160,6 +160,7 @@ filterwarnings = [
     "error",
     "ignore:Jitify is performing a one-time only warm-up to populate the persistent cache",
     "ignore::DeprecationWarning:botocore.*",
+    "ignore::FutureWarning:.*nvcomp|.*nvcomp_codec",
```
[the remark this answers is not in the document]
kingcrimsontianyu: I set the filter before the import.
According to further discussion, the pytestmark is only able to suppress the warning from the tests but not from import. So the filter is still added to the `pyproject.toml` in this PR.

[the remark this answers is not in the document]
kingcrimsontianyu: Added the unit tests for warnings.

python/kvikio/kvikio/utils.py
kingcrimsontianyu: Done.

TomAugspurger (commented):
Looks good overall, thanks.

To make sure we're on the same page for the Zarr side of things, the `kvikio.zarr` implementation for zarr-python 2.x currently imports `kvikio.nvcomp`: https://github.com/rapidsai/kvikio/blob/362b1d707098080fd77c23856ba4afd9294f97d8/python/kvikio/kvikio/zarr/_zarr_python_2.py#L26-L27. AFAICT, this will mean users of `kvikio.zarr` with zarr-python=2.x in their environment will see this warning, which they can't do anything about.

Are we OK with that? IMO that's not a great user experience.

python/kvikio/kvikio/utils.py
```diff
@@ -157,8 +165,28 @@ def wrapper(*args, **kwargs):
...
+        wrapper.__doc__ = "{:} {:} {:}\n\n    {:}".format(
+            ".. deprecated::", since, msg, valid_docstring
+        )
+
         return wrapper
 
     return decorator
+
+
+def kvikio_deprecate_module(msg: str = "", *, since: str = __version__) -> None:
```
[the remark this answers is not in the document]
kingcrimsontianyu: Done

python/kvikio/kvikio/nvcomp_codec.py
```diff
@@ -9,6 +9,9 @@
 
 from kvikio._lib.libnvcomp_ll import SUPPORTED_ALGORITHMS
 from kvikio.numcodecs import BufferLike, CudaCodec
+from kvikio.utils import kvikio_deprecate_module
+
+kvikio_deprecate_module("Use official nvCOMP Python API instead.", since="25.06")
```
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks. Done.

python/kvikio/kvikio/nvcomp.py
```diff
@@ -8,6 +8,9 @@
 
 import kvikio._lib.libnvcomp as _lib
 from kvikio._lib.arr import asarray
+from kvikio.utils import kvikio_deprecate_module
+
+kvikio_deprecate_module("Use official nvCOMP Python API instead.", since="25.06")
```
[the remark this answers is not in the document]
kingcrimsontianyu: Done.

README.md
```diff
@@ -15,7 +15,7 @@ KvikIO also works efficiently when GDS isn't available and can read/write both h
 * Concurrent reads and writes using an internal thread pool.
 * Non-blocking API.
 * Transparently handles reads and writes to/from memory on both host and device.
-* Provides Python bindings to [nvCOMP](https://github.com/NVIDIA/nvcomp).
+* (Deprecated) Provides Python bindings to [nvCOMP](https://github.com/NVIDIA/nvcomp).
```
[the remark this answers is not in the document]
kingcrimsontianyu: Done. Thanks.

kingcrimsontianyu:
>Move the nvcomp codecs to a private model that kvikio.zarr._zarr_python_2 can import from (without warning). Keep the import path with a warning from the public API.

If I'm not having a misunderstanding of `FutureWarning`, wouldn't the warning always get issued when the module is imported regardless of where the import takes place (unlike `DeprecationWarning`)?

TomAugspurger:
> If I'm not having a misunderstanding of FutureWarning, wouldn't the warning always get issued when the module is imported regardless of where the import takes place (unlike DeprecationWarning)?

That's mostly correct. Whether it's a `FutureWarning` or `DeprecationWarning` doesn't affect whether the warning is emitted, just whether or not it's visible to the user.

If we need to keep around the nvcomp bindings for the zarr-python-2 implementation, but want to remove them from the public API since we want users to migrate to `nvidia.nvcomp` then we can

- move the implementation of `kvikio.nvcomp` to a private module like `kvikio._nvcomp`
- update `kvikio.zarr._zarrpython_2` to import from `kvikio._nvcomp`
- - Update `kvikio.nvcomp` to be something like 

```python
kvikio_deprecate_module("Use nvidia.nvcomp...")
from kvikio._nvcomp import *
```

That way the warning is triggered for users on `import kvikio.nvcomp` but not on `import kvikio.zarr` (with zarr-python 2.x).

kingcrimsontianyu:
@TomAugspurger Thanks for the explanations.
I feel that the initial purpose of using `FutureWarning` is to propagate the warning message to both library developers _and_ application users, so issuing a (annoying) warning about the underlying nvcomp API change when `import kvikio.zarr` is probably what we want? If we prefer not to have this behavior, perhaps `DeprecationWarning` is more appropriate?
There was some discussion related to this earlier: https://github.com/rapidsai/kvikio/pull/644#pullrequestreview-2653257332 involving @madsbk @TomAugspurger @bdice 
I'm open to any further changes, including moving the current nvcomp to a private module as Tom proposed.

madsbk:
> Based on @TomAugspurger's comment I'd like to check with @madsbk if we plan to continue the support for Zarr-Python version 2.

Yes, I think we are in the short to medium term.

madsbk:
I like @TomAugspurger idea of only warning when importing `kvikio.nvcomp` directly.

TomAugspurger:
> I feel that the initial purpose of using FutureWarning is to propagate the warning message to both library developers and application users, so issuing a (annoying) warning about the underlying nvcomp API change when import kvikio.zarr is probably what we want?

I don't think users should see warnings they can't do anything about. In the case of `import kvikio.zarr` (with zarr-python 2.x installed), emitting a warning that `kvikio.nvcomp` is being removed in the future doesn't leave them with any action to take to avoid the warning, and so is just noise.

kingcrimsontianyu:
I tuck the nvcomp contents in a private module. `pre-commit run` lists the following issues that need further investigation:
```
python/kvikio/kvikio/nvcomp.py:5:1: F403 'from kvikio._nvcomp import *' used; unable to detect undefined names
python/kvikio/kvikio/nvcomp.py:5:1: F401 'kvikio._nvcomp.*' imported but unused
python/kvikio/kvikio/nvcomp_codec.py:4:1: F403 'from kvikio._nvcomp_codec import *' used; unable to detect undefined names
python/kvikio/kvikio/nvcomp_codec.py:4:1: F401 'kvikio._nvcomp_codec.*' imported but unused
```
Perhaps somewhere the `__all__` property needs modifications. I'm not familiar with this part of Python. Need some studying.

```
python/kvikio/kvikio/zarr/_zarr_python_2.py:341: error: Unexpected keyword argument "path" for "GDSStore"  [call-arg]
Found 1 error in 1 file (checked 42 source files)
```
Not sure why.

```
In file python/kvikio/kvikio/_nvcomp_codec.py:1:3:
-# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
note: suggested fix applied
```
It looks like `verify-copyright` thinks it is 2023 now.

TomAugspurger:
For the `__all__` issue, probably easiest to have the new `kvikio.nvcomp` be something like

```
from kvikio._nvcomp import cp_to_nvcomp_dtype, nvCompManager, ... # repeat for all public classes / functions

__all__ = [
    "cp_to_nvcomp_dtype",
    "nvCompManager",
    ...
]
```

python/kvikio/kvikio/zarr/_zarr_python_2.py
```diff
@@ -337,7 +337,8 @@ def open_cupy_array(
         # In order to handle "a", we start by trying to open the file in read mode.
         try:
             ret = zarr.open_array(
-                store=kvikio.zarr.GDSStore(path=store),  # type: ignore[call-arg]
+                # type: ignore[call-arg]
+                store=kvikio.zarr.GDSStore(path=store),
```
TomAugspurger, line 341: The `type: ignore[call-arg]` needs to be on the same line as the `kvikio.zarr.GDSStore(path=store)`.

If this needs to be reformatted, I think that you can move this outside of the `zarr.open_array`
```
store = store=kvikio.zarr.GDSStore(path=store)  # type: ignore[call-arg]
ret = zarr.open_array(
    store=store,
    ...
)
```

kingcrimsontianyu: Thanks!
Looks like the Python reformatter in my editor did the change automatically.

TomAugspurger (approved):
That nvcomp -> _nvcomp change looks good now, thanks.

kingcrimsontianyu:
Thanks for the review!

bdice (approved):
Seems fine -- I skimmed the diff but agreed with @TomAugspurger's earlier comments. It looks like this is in a relatively good place.

Issue #720, opened 2025-05-16
Add the feature to detect NVLink C2C platform


kingcrimsontianyu:
Recent benchmark shows that the best cold cache read performance on GH200 requires a different number of threads from that on x86+PCIe platforms, making it necessary to have an API that can automatically determine a platform-dependent, optimal thread count setting. Internal discussion about https://github.com/rapidsai/cudf/issues/18272 suggests that this feature should be part of KvikIO instead of cuDF.

The first step is to introduce the capability to detect C2C platforms. Although this feature can be made internal to KvikIO, it won't hurt to expose it as a public C++ API.

PR #721, opened 2025-05-16
Provide API to detect C2C platform and add NVML shim

kingcrimsontianyu:
Closes #https://github.com/rapidsai/kvikio/issues/720

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -151,6 +151,7 @@ set(SOURCES
     "src/posix_io.cpp"
     "src/shim/cuda.cpp"
     "src/shim/cufile.cpp"
+    "src/shim/nvml.cpp"
     "src/shim/utils.cpp"
     "src/stream.cpp"
     "src/utils.cpp"
--- a/cpp/src/shim/cuda.cpp
+++ b/cpp/src/shim/cuda.cpp
@@ -38,6 +38,7 @@ cudaAPI::cudaAPI()
   get_symbol(CtxPushCurrent, lib, KVIKIO_STRINGIFY(cuCtxPushCurrent));
   get_symbol(CtxPopCurrent, lib, KVIKIO_STRINGIFY(cuCtxPopCurrent));
   get_symbol(CtxGetCurrent, lib, KVIKIO_STRINGIFY(cuCtxGetCurrent));
+  get_symbol(CtxGetDevice, lib, KVIKIO_STRINGIFY(cuCtxGetDevice));
   get_symbol(MemGetAddressRange, lib, KVIKIO_STRINGIFY(cuMemGetAddressRange));
   get_symbol(GetErrorName, lib, KVIKIO_STRINGIFY(cuGetErrorName));
   get_symbol(GetErrorString, lib, KVIKIO_STRINGIFY(cuGetErrorString));
--- a/cpp/src/shim/nvml.cpp
+++ b/cpp/src/shim/nvml.cpp
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "nvml.hpp"
+#include <kvikio/error.hpp>
+#include "kvikio/shim/cuda.hpp"
+
+namespace kvikio {
+
+#ifdef KVIKIO_CUDA_FOUND
+NvmlAPI::NvmlAPI()
+{
+  auto lib_handle = load_library("libnvidia-ml.so.1");
+  get_symbol(Init, lib_handle, "nvmlInit_v2");
+  get_symbol(Shutdown, lib_handle, "nvmlShutdown");
+  get_symbol(ErrorString, lib_handle, "nvmlErrorString");
+  get_symbol(DeviceGetHandleByIndex, lib_handle, "nvmlDeviceGetHandleByIndex_v2");
+  get_symbol(DeviceGetFieldValues, lib_handle, "nvmlDeviceGetFieldValues");
+  CHECK_NVML(Init());
+}
+#else
+NvmlAPI::NvmlAPI() { KVIKIO_FAIL("KvikIO not compiled with CUDA support.", std::runtime_error); }
+#endif
+
+NvmlAPI& NvmlAPI::instance()
+{
+  static NvmlAPI instance;
+  return instance;
+}
+
+#ifdef KVIKIO_CUDA_FOUND
+bool is_nvml_available()
+{
+  try {
+    NvmlAPI::instance();
+  } catch (std::runtime_error const&) {
+    return false;
+  }
+  return true;
+}
+#endif
+
+#ifdef KVIKIO_CUDA_FOUND
+bool is_c2c_available()
+{
+  // todo: remove this once CUDA 11 support is dropped
+#if CUDA_VERSION < 12000
+  return false;
+#else
+  nvmlDevice_t device_handle_nvml{};
+  CUdevice device_handle_cuda{};
+  cudaAPI::instance().CtxGetDevice(&device_handle_cuda);
+  CHECK_NVML(NvmlAPI::instance().DeviceGetHandleByIndex(device_handle_cuda, &device_handle_nvml));
+
+  nvmlFieldValue_t field{};
+  field.fieldId = NVML_FI_DEV_C2C_LINK_COUNT;
+  CHECK_NVML(NvmlAPI::instance().DeviceGetFieldValues(device_handle_nvml, 1, &field));
+
+  return (field.nvmlReturn == nvmlReturn_t::NVML_SUCCESS) && (field.value.uiVal > 0);
+#endif
+}
+#endif
+
+}  // namespace kvikio
--- a/cpp/src/shim/nvml.hpp
+++ b/cpp/src/shim/nvml.hpp
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <sstream>
+
+#include "nvml_h_wrapper.hpp"
+
+/**
+ * @brief Macro for checking the error code of the NVML API call.
+ *
+ * @param err_code The error code of the NVML API call.
+ *
+ * @throws std::runtime_error if the NVML API call fails.
+ */
+#define CHECK_NVML(err_code)                          \
+  do {                                                \
+    kvikio::check_nvml(err_code, __FILE__, __LINE__); \
+  } while (0)
+
+namespace kvikio {
+
+/**
+ * @brief Singleton class to manage dynamic loading of the NVML library.
+ *
+ * NVML initialization is costly. Using this singleton ensures it is performed only once per
+ * process. NVML shutdown is not performed in the destructor, but a wrapper is provided anyway to be
+ * explicitly called on users' discretion.
+ */
+class NvmlAPI {
+ public:
+  NvmlAPI(NvmlAPI const&)            = delete;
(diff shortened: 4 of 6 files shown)


cpp/CMakeLists.txt
```diff
@@ -77,6 +77,12 @@ if(KvikIO_CUDA_SUPPORT)
   )
   include(cmake/thirdparty/get_nvtx.cmake)
 
+  if(NOT TARGET CUDA::nvml)
```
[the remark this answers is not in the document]
vyasr: Instead of doing this could we instead use [`BUILD_LOCAL_INTERFACE`](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:BUILD_LOCAL_INTERFACE) and link properly below?

cpp/CMakeLists.txt
```diff
@@ -77,6 +80,18 @@ if(KvikIO_CUDA_SUPPORT)
   )
   include(cmake/thirdparty/get_nvtx.cmake)
 
+  if(CUDAToolkit_VERSION_MAJOR VERSION_GREATER_EQUAL 12)
```
kingcrimsontianyu, line 83: Per @bdice 's comment here https://github.com/rapidsai/cudf/pull/18447#discussion_r2062699204, the NVML dependency (specified via several `.yaml` files) is not added to CUDA 11. `is_c2c_available()` is expected to return false if CUDA 11 is detected at build time or at runtime.

cpp/CMakeLists.txt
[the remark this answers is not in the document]
kingcrimsontianyu: @vyasr Thanks. Good to know of this expression. When I switched to
```
target_link_libraries(
  kvikio
  PUBLIC ...
  PRIVATE ... $<BUILD_LOCAL_INTERFACE:CUDA::nvml>
)
```
this CI error followed, and I'm not sure about the cause. https://github.com/rapidsai/kvikio/actions/runs/15138674767/job/42557566983#step:11:992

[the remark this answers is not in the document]
robertmaynard: `BUILD_LOCAL_INTERFACE` will still cause `nvml` to be on the link line and therefore cause a DT_NEEDED entry.

I am pretty sure that we can do:
```
target_link_libraries(
  kvikio
  PUBLIC ...
  PRIVATE ... CUDA::toolkit
)
```
As the `CUDA::toolkit` target holds all the include directories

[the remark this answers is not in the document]
kingcrimsontianyu: @robertmaynard Great, thanks! I'll try that.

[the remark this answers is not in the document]
kingcrimsontianyu: Cool. CI has passed.

cpp/src/shim/nvml_h_wrapper.hpp
```diff
@@ -0,0 +1,31 @@
...
+enum nvmlReturn_t {
+  NVML_SUCCESS,
+};
+using nvmlDevice_t = int;
+nvmlReturn_t nvmlInit_v2(...);
+nvmlReturn_t nvmlShutdown(...);
+nvmlReturn_t nvmlErrorString(...);
+nvmlReturn_t nvmlDeviceGetHandleByIndex_v2(...);
+nvmlReturn_t nvmlDeviceGetFieldValues(...);
+nvmlReturn_t nvmlDeviceGetHandleByUUID(...);
```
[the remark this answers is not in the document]
kingcrimsontianyu: All credits to @madsbk for using this neat C++ feature on cuFile and CUDA shims.

cpp/src/shim/nvml.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+#ifdef KVIKIO_NVML_FOUND
+bool is_nvml_available()
+{
+  try {
+    NvmlAPI::instance();
+  } catch (std::runtime_error const&) {
+    return false;
+  }
+  return true;
```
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks! Added the result caching for NVML, CUDA, cuFile shims.

vuule (approved):

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -99,6 +101,7 @@ outputs:
                 - libcufile ${{ cuda11_libcufile_run_version }}
                 - libcufile-dev ${{ cuda11_libcufile_run_version }}
           else:
+            - cuda-nvml-dev
             - if: linux
```
bdice, line 105: This line is a little odd. I don't think we ship kvikio on non-Linux platforms. However it's not changed in this PR so let's leave it alone. I am fine with adding `cuda-nvml-dev` outside this `if` block as it currently is.

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -109,6 +112,8 @@ outputs:
           - if: cuda_major == "11"
             then:
               - cudatoolkit
+            else:
+              - cuda-nvml-dev
```
bdice, line 116: There are no run exports in https://github.com/conda-forge/cuda-nvml-dev-feedstock/blob/main/recipe/meta.yaml.
```suggestion
```

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -165,6 +172,8 @@ outputs:
...
```
bdice, line 176:
```suggestion
```

raydouglass (approved):

cpp/src/shim/nvml.cpp
```diff
@@ -0,0 +1,102 @@
...
+    std::string const mig_uuid = "MIG-" + ss.str();
+    CHECK_NVML(NvmlAPI::instance().DeviceGetHandleByUUID(mig_uuid.c_str(), &nvml_device_handle));
+  }
+
+  return nvml_device_handle;
+}
+#else
+nvmlDevice_t convert_device_handle_from_cuda_to_nvml(CUdevice cuda_device_handle)
+{
+  KVIKIO_FAIL("KvikIO not compiled with CUDA support.", std::runtime_error);
```
bdice, line 97:
```suggestion
  KVIKIO_FAIL("KvikIO not compiled with NVML support.", std::runtime_error);
```

robertmaynard (approved):

cpp/src/shim/nvml.hpp
```diff
@@ -0,0 +1,110 @@
...
+  } while (0)
+
+namespace kvikio {
+
+/**
+ * @brief Singleton class to manage dynamic loading of the NVML library.
+ *
+ * @note NVML initialization is costly, and is performed in the constructor of this singleton class.
+ * NVML shutdown is not performed in the destructor, but a wrapper is provided anyway to be
+ * explicitly called on users' discretion.
```
bdice, line 43:
```suggestion
 * explicitly called at a user's discretion.
```

cpp/src/utils.cpp
```diff
@@ -180,4 +181,39 @@ std::tuple<void*, std::size_t, std::size_t> get_alloc_info(void const* devPtr, C
...
 }
 
+#ifdef KVIKIO_NVML_FOUND
+bool is_c2c_available(int device_idx, DeviceIdType device_id_type)
+{
+  // todo: Remove the version checking once CUDA 11 support is dropped.
+  // Version format: 1000 * major + 10 * minor
+  int cuda_driver_version{};
+  cudaAPI::instance().DriverGetVersion(&cuda_driver_version);
+  if (cuda_driver_version <= 12000) { return false; }
```
bdice, line 191: We know the CUDA version at build time and only enable NVML for CUDA >=12. Do we need these runtime checks at all? Can we instead use something like

```cpp
#ifndef KVIKIO_NVML_FOUND
return false;
#else
.... the rest of the logic ....
#endif
```
?

That should eliminate the compile time issue with CUDA 11, since CUDA 11 would never try to use a pre-defined NVML macro.

conda/recipes/libkvikio/recipe.yaml
kingcrimsontianyu: Thanks. Done.

conda/recipes/libkvikio/recipe.yaml
kingcrimsontianyu: Done.

cpp/src/shim/nvml.cpp
kingcrimsontianyu: Done. Made the same change to another place as well in the same file.

cpp/src/shim/nvml.hpp
kingcrimsontianyu: Done. Thanks.

kingcrimsontianyu:
Trivial addition: Added the missing free function `kvikio::nvml_shutdown` in the utils, that allows users to explicitly clean up NVML before main returns (the shim singleton doesn't do this in the destructor).

cpp/src/utils.cpp
kingcrimsontianyu: I was thinking of a special case where a program using libkvikio is built with CUDA 12, but deployed on a system with CUDA 11 (with the forward compatibility support). But on second thought, this does not apply to CUDA driver API. So this check should be removed.

PS: The suggested change that `is_c2c_available` returns false when `KVIKIO_NVML_FOUND` is not defined is already [implemented in the header file](https://github.com/rapidsai/kvikio/pull/721/files#diff-eecafe805328a42d93427690a3e0e9aa55c8f6680ad6f5a839a06f010c850df6R222).

cpp/src/utils.cpp
```diff
@@ -180,4 +181,40 @@ std::tuple<void*, std::size_t, std::size_t> get_alloc_info(void const* devPtr, C
...
+    CHECK_NVML(NvmlAPI::instance().DeviceGetHandleByIndex(device_idx, &device_handle_nvml));
+  }
+
+  nvmlFieldValue_t field{};
+
+  // todo: This is to fix the compile error for CUDA 11. Remove this part once CUDA 11 support is
+  // dropped.
+#ifndef NVML_FI_DEV_C2C_LINK_COUNT
+#define NVML_FI_DEV_C2C_LINK_COUNT 0
+#endif
```
bdice, line 202: This code path should not be reachable on CUDA 11 systems. Can this be removed?

bdice (approved):

cpp/src/utils.cpp
kingcrimsontianyu: Thanks. Good catch. Done!

cpp/CMakeLists.txt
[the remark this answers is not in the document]
vyasr: > `BUILD_LOCAL_INTERFACE` will still cause `nvml` to be on the link line and therefore cause a DT_NEEDED entry.
> 
> I am pretty sure that we can do:
> 
> ```
> target_link_libraries(
>   kvikio
>   PUBLIC ...
>   PRIVATE ... CUDA::toolkit
> )
> ```
> 
> As the `CUDA::toolkit` target holds all the include directories

Good call, I was thinking of just trying to isolate the dependency to be internal but you're right that's not sufficient here.

cpp/src/shim/nvml.cpp
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+#include "nvml.hpp"
+
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+
+#include <kvikio/error.hpp>
+#include <kvikio/shim/cuda.hpp>
```
vyasr, line 24:
```suggestion
#include <kvikio/shim/cuda.hpp>
#include <kvikio/shim/utils.hpp>
```

vyasr (changes requested):

cpp/src/shim/nvml.cpp
```diff
@@ -0,0 +1,104 @@
...
+NvmlAPI& NvmlAPI::instance()
+{
+  static NvmlAPI instance;
+  return instance;
+}
+
+void NvmlAPI::shutdown() { CHECK_NVML(instance().Shutdown()); }
+
+#ifdef KVIKIO_NVML_FOUND
+bool is_nvml_available()
```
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks. Great idea. Implemented.

conda/recipes/kvikio/recipe.yaml
```diff
@@ -78,6 +78,7 @@ requirements:
           then: libcufile-dev =${{ cuda11_libcufile_host_version }}
       else:
         - cuda-cudart-dev
+        - cuda-nvml-dev
```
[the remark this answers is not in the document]
kingcrimsontianyu: Thanks. Removed.

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -142,6 +145,7 @@ outputs:
                 - libcufile-dev =${{ cuda11_libcufile_host_version }}
           else:
             - cuda-cudart-dev
+            - cuda-nvml-dev
```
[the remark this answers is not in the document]
kingcrimsontianyu: Deleted.

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -154,6 +158,7 @@ outputs:
                 - libcufile =${{ cuda11_libcufile_host_version }}
           else:
             - cuda-cudart
+            - cuda-nvml-dev
```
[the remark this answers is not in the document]
kingcrimsontianyu: Erased.

cpp/CMakeLists.txt
```diff
@@ -177,17 +189,21 @@ target_include_directories(
   INTERFACE "$<INSTALL_INTERFACE:include>"
 )
 
-# Notice, we do not link to cuda or cufile since KvikIO opens them manually using `dlopen()`.
+# Notice, we do not link to cuda or cufile since KvikIO opens them manually using `dlopen()`. Also
+# note that KvikIO uses an NVML shim that dynamically loads the NVML shared library at runtime. At
+# build time only the header location is needed. The shared library is not linked against KvikIO.
```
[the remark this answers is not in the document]
kingcrimsontianyu: Done.

cpp/src/shim/cuda.cpp
```diff
@@ -61,12 +64,15 @@ cudaAPI& cudaAPI::instance()
 #ifdef KVIKIO_CUDA_FOUND
 bool is_cuda_available()
 {
-  try {
-    cudaAPI::instance();
-  } catch (std::runtime_error const&) {
-    return false;
-  }
-  return true;
+  static auto result = []() -> bool {
```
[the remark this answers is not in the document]
kingcrimsontianyu: Yes!

cpp/src/shim/nvml.cpp
kingcrimsontianyu: Thanks. Done.

vyasr (approved):

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -99,6 +101,7 @@ outputs:
                 - libcufile ${{ cuda11_libcufile_run_version }}
                 - libcufile-dev ${{ cuda11_libcufile_run_version }}
           else:
+            - cuda-nvml-dev
```
jakirkham, line 104: So this NVML package only has headers and a stub library

The driver has a library, but it is not in this package

If we are building a KvikIO library, there should be no need for this dependency at runtime

PR #729, opened 2025-05-22
Check whether memory resource/BAR1 are supported

pentschev:
After further discussions with @jakirkham, we believe this is a more appropriate way to determine whether the device has a memory resource (i.e., dedicated memory) than relying on MIG-related checks as done previously in https://github.com/rapidsai/kvikio/pull/703, this will also align more to https://github.com/rapidsai/ucxx/pull/424. Additionally, check that BAR1 is supported and report appropriately.

### DIFF
--- a/python/kvikio/kvikio/benchmarks/utils.py
+++ b/python/kvikio/kvikio/benchmarks/utils.py
@@ -45,15 +45,15 @@ def pprint_sys_info() -> None:
             dev = pynvml.nvmlDeviceGetHandleByIndex(0)
 
         if dev is not None:
+            gpu_name = f"{pynvml.nvmlDeviceGetName(dev)} (dev #0)"
             try:
-                pynvml.nvmlDeviceGetMigMode(dev)
-            except pynvml.NVMLError:
-                with contextlib.suppress(pynvml.NVMLError):
-                    gpu_name = f"{pynvml.nvmlDeviceGetName(dev)} (dev #0)"
-                    mem_total = format_bytes(pynvml.nvmlDeviceGetMemoryInfo(dev).total)
-                    bar1_total = format_bytes(
-                        pynvml.nvmlDeviceGetBAR1MemoryInfo(dev).bar1Total
-                    )
+                mem_total = format_bytes(pynvml.nvmlDeviceGetMemoryInfo(dev).total)
+            except pynvml.NVMLError_NotSupported:
+                mem_total = "Device has no memory resource"
+            try:
+                bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(handle).bar1Total
+            except pynvml.NVMLError_NotSupported:
+                bar1_total = "Device has no BAR1 support"
 
     if version == (0, 0):
         libcufile_version = "unknown (earlier than cuFile 1.8)"


bdice (approved):

python/kvikio/kvikio/benchmarks/utils.py
```diff
@@ -45,15 +45,15 @@ def pprint_sys_info() -> None:
...
-                    bar1_total = format_bytes(
-                        pynvml.nvmlDeviceGetBAR1MemoryInfo(dev).bar1Total
-                    )
+                mem_total = format_bytes(pynvml.nvmlDeviceGetMemoryInfo(dev).total)
+            except pynvml.NVMLError_NotSupported:
+                mem_total = "Device has no memory resource"
+            try:
+                bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(handle).bar1Total
+            except pynvml.NVMLError_NotSupported:
+                bar1_total = "Device has no BAR1 support"
```
jakirkham, line 56: Is it possible for the attribute access to fail? IOW should we add `AttributeError` to the `except`s?

Or can we count on the attributes to be defined assuming the functions return without error?

jakirkham (commented):
Thanks Peter! 🙏

Had one question below

python/kvikio/kvikio/benchmarks/utils.py
```diff
@@ -45,15 +45,15 @@ def pprint_sys_info() -> None:
...
-                    gpu_name = f"{pynvml.nvmlDeviceGetName(dev)} (dev #0)"
-                    mem_total = format_bytes(pynvml.nvmlDeviceGetMemoryInfo(dev).total)
-                    bar1_total = format_bytes(
-                        pynvml.nvmlDeviceGetBAR1MemoryInfo(dev).bar1Total
-                    )
+                mem_total = format_bytes(pynvml.nvmlDeviceGetMemoryInfo(dev).total)
+            except pynvml.NVMLError_NotSupported:
+                mem_total = "Device has no memory resource"
+            try:
+                bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(handle).bar1Total
```
jakirkham, line 54: This gets flagged by the style check [on CI]( https://github.com/rapidsai/kvikio/actions/runs/15196629457/job/42742221569?pr=729#step:7:227 )

```
python/kvikio/kvikio/benchmarks/utils.py:54:65: F821 undefined name 'handle'
```

Guessing we want to use `dev` here as used in the rest of the code

```suggestion
                bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(dev).bar1Total
```

python/kvikio/kvikio/benchmarks/utils.py
pentschev: No, I don't think so, those are fixed attributes of the [nvmlBAR1Memory_t struct](https://docs.nvidia.com/deploy/nvml-api/structnvmlBAR1Memory__t.html#structnvmlBAR1Memory__t), the same applies to [nvmlMemory_t](https://docs.nvidia.com/deploy/nvml-api/structnvmlMemory__t.html#structnvmlMemory__t) and [nvmlMemory_v2_t struct](https://docs.nvidia.com/deploy/nvml-api/structnvmlMemory__v2__t.html#structnvmlMemory__v2__t).

python/kvikio/kvikio/benchmarks/utils.py
pentschev: Good catch, thanks @jakirkham !

jakirkham (approved):

PR #731, opened 2025-05-22
Revert "Provide API to detect C2C platform and add NVML shim"

vyasr:
Reverts rapidsai/kvikio#721

We're seeing some downstream build breakages in cudf 25.08 runs. It looks like this PR is requiring that NVML be installed in the build environment, which was not previously the case, and since cudf is now frozen for 25.06 we don't want to have to modify its build setup right now. We can reopen this PR for 25.08 and then get the issues fixed there.


bdice (approved):

jakirkham:
Would be interested to know if this comes down to a packaging issue

xref: https://github.com/rapidsai/kvikio/pull/721#discussion_r2103102057

In any event agree retargeting to 25.08 seems appropriate

PR #735, opened 2025-05-27
Add an overload getenv_or that supports env var alias. Add new env var KVIKIO_NUM_THREADS. Fix UB.

kingcrimsontianyu:
This PR performs the following items:
- Add a new overload `kvikio::getenv_or` that supports env var alias, i.e. multiple different env vars referring to the same property.
- Add new env var `KVIKIO_NUM_THREADS` as the alias of `KVIKIO_NTHREADS`, and add new C++ API `num_threads`/`set_thread_pool_nthreads` as the alias of existing `thread_pool_nthreads`/`set_thread_pool_nthreads`. This makes C++ and Python API (i.e. `kvikio.defaults.get("num_threads")`) consistent.
- Fix a well-hidden UB related to string temporaries and `std::initializer_list`.

### DIFF
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -62,7 +62,7 @@ std::vector<int> getenv_or(std::string_view env_var_name, std::vector<int> defau
  * @brief Get the environment variable value from a candidate list
  *
  * @tparam T Type of the environment variable value
- * @param env_var_names Name of the environment variable
+ * @param env_var_names Candidate list containing the names of environment variable
  * @param default_val Default value of the environment variable, if none of the candidates has been
  * found
  * @return A tuple of (`env_var_name`, `result`, `has_found`), where:
@@ -73,10 +73,8 @@ std::vector<int> getenv_or(std::string_view env_var_name, std::vector<int> defau
  *
  * @throws std::invalid_argument if:
  *   - `env_var_names` is empty
- *   - An environment variable from `env_var_names` has an empty value, e.g. by setting `export
- * KVIKIO_NTHREADS=`
  *   - Multiple candidates have been set at the same time
- *   - An invalid value is given
+ *   - An invalid value is given, e.g. value that cannot be converted to type T
  */
 template <typename T>
 std::tuple<std::string_view, T, bool> getenv_or(
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -92,9 +92,6 @@ std::tuple<std::string_view, T, bool> getenv_or(
   for (auto const& env_var_name : env_var_names) {
     auto const* env_val = std::getenv(env_var_name.data());
     if (env_val == nullptr) { continue; }
-    KVIKIO_EXPECT(std::strlen(env_val) != 0,
-                  std::string{env_var_name} + " must not have an empty value.",
-                  std::invalid_argument);
     KVIKIO_EXPECT(
       has_already_been_set == false,
       "Environment variable " + std::string{env_var_name} + " has already been set by its alias.",
@@ -104,15 +101,9 @@ std::tuple<std::string_view, T, bool> getenv_or(
     env_val_target       = env_val;
   }
 
-  if (env_val_target.empty()) { return {env_val_target, default_val, false}; }
+  if (env_name_target.empty()) { return {env_name_target, default_val, false}; }
 
-  std::stringstream sstream(env_val_target.data());
-  T res;
-  sstream >> res;
-  KVIKIO_EXPECT(
-    !sstream.fail(),
-    std::string{"Unknown config value "} + env_name_target.data() + "=" + env_val_target.data(),
-    std::invalid_argument);
+  auto res = getenv_or<T>(env_name_target, default_val);
   return {env_name_target, res, true};
 }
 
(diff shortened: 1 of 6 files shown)


cpp/include/kvikio/defaults.hpp
```diff
@@ -56,6 +58,64 @@ CompatMode getenv_or(std::string_view env_var_name, CompatMode default_val);
...
+  std::string_view env_val_target;
+  bool has_already_been_set{false};
+  for (auto const& env_var_name : env_var_names) {
+    auto const* env_val = std::getenv(env_var_name.data());
+    if (env_val == nullptr) { continue; }
+    KVIKIO_EXPECT(std::strlen(env_val) != 0,
+                  std::string{env_var_name} + " must not have an empty value.",
+                  std::invalid_argument);
+    KVIKIO_EXPECT(
+      has_already_been_set == false,
```
vuule, line 99:
```suggestion
      !has_already_been_set,
```

cpp/include/kvikio/defaults.hpp
```diff
@@ -56,6 +58,64 @@ CompatMode getenv_or(std::string_view env_var_name, CompatMode default_val);
...
+                "`env_var_names` must contain at least one environment variable name.",
+                std::invalid_argument);
+  std::string_view env_name_target;
+  std::string_view env_val_target;
+  bool has_already_been_set{false};
+  for (auto const& env_var_name : env_var_names) {
+    auto const* env_val = std::getenv(env_var_name.data());
+    if (env_val == nullptr) { continue; }
+    KVIKIO_EXPECT(std::strlen(env_val) != 0,
+                  std::string{env_var_name} + " must not have an empty value.",
```
vuule, line 96: why do we have this limitation?

cpp/include/kvikio/defaults.hpp
```diff
@@ -56,6 +58,64 @@ CompatMode getenv_or(std::string_view env_var_name, CompatMode default_val);
...
+ *   - An environment variable from `env_var_names` has an empty value, e.g. by setting `export
+ * KVIKIO_NTHREADS=`
+ *   - Multiple candidates have been set at the same time
+ *   - An invalid value is given
+ */
+template <typename T>
+std::tuple<std::string_view, T, bool> getenv_or(
+  std::initializer_list<std::string_view> env_var_names, T default_val)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
```
vuule, line 85: Do we need this range? reading env vars is usually very fast, I don't see when having this range would help with profiling.

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: I think existing implementation (such as the [primary template](https://github.com/rapidsai/kvikio/blob/branch-25.06/cpp/include/kvikio/defaults.hpp#L41)) has already prohibited use of empty values. In such cases, the string stream would have a botched extraction and `fail()` become true, causing `KVIKIO_EXPECT` to fail.
I've removed this eager checking.

cpp/include/kvikio/defaults.hpp
```diff
@@ -56,6 +58,54 @@ CompatMode getenv_or(std::string_view env_var_name, CompatMode default_val);
...
+      "Environment variable " + std::string{env_var_name} + " has already been set by its alias.",
+      std::invalid_argument);
+    has_already_been_set = true;
+    env_name_target      = env_var_name;
+    env_val_target       = env_val;
+  }
+
+  if (env_name_target.empty()) { return {env_name_target, default_val, false}; }
+
+  auto res = getenv_or<T>(env_name_target, default_val);
```
kingcrimsontianyu, line 105: Simplified the logic so that each type's own version of `getenv_or<T>` is in charge of value extraction.

cpp/include/kvikio/defaults.hpp
```diff
@@ -56,6 +58,52 @@ CompatMode getenv_or(std::string_view env_var_name, CompatMode default_val);
...
+  std::string_view env_name_target;
+  std::string_view env_val_target;
+  bool has_already_been_set{false};
+  for (auto const& env_var_name : env_var_names) {
+    auto const* env_val = std::getenv(env_var_name.data());
+    if (env_val == nullptr) { continue; }
+    KVIKIO_EXPECT(
+      !has_already_been_set,
+      "Environment variable " + std::string{env_var_name} + " has already been set by its alias.",
+      std::invalid_argument);
```
vuule, line 95: how messy would it be to allow multiple aliases to be set, as long as they have the same value?

cpp/include/kvikio/defaults.hpp
vuule: adds an extra call to `getenv`, but AFAIK this is fine - repeated calls are VERY fast.

cpp/tests/test_defaults.cpp
```diff
@@ -72,3 +80,120 @@ TEST(DefaultsTest, parse_http_status_codes)
...
+                  "`env_var_names` must contain at least one environment variable name")));
+  }
+
+  // Env var has an empty value
+  {
+    kvikio::test::EnvVarContext env_var_ctx{{{"KVIKIO_TEST_ALIAS", ""}}};
+    EXPECT_THAT(
+      [=] { kvikio::getenv_or({"KVIKIO_TEST_ALIAS"}, 123); },
+      ThrowsMessage<std::invalid_argument>(HasSubstr("unknown config value KVIKIO_TEST_ALIAS=")));
+  }
```
vuule, line 99: How about a test where the empty variable is read as a string, so the value is valid?

vuule (commented):
looks good, just a few ideas

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: Implemented, in a not so messy way.

madsbk (approved):
Looks great!

PR #740, opened 2025-05-29
Support file-backed mapping (1/n): host/device read C++ implementation

kingcrimsontianyu:
This PR introduces memory-mapped I/O (`MmapHandle`) as an alternative to the standard I/O (`FileHandle`).

The benchmark results are at https://github.com/rapidsai/kvikio/issues/530#issuecomment-2994871923

Partially addresses #530


cpp/include/kvikio/shim/cuda.hpp
```diff
@@ -15,11 +15,67 @@
...
+  }
+
+  /**
+   * @brief Check if the object holds a callable
+   */
+  operator bool() const { return _callable.has_value(); }
+};
+
+}  // namespace detail
+
```
kingcrimsontianyu, line 81: The `AnyCallable` class delays type specification of `MemcpyBatchAsync` to the symbol loading time in the constructor. This is an effort to get the best of the two worlds as per the team discussion:
- We want to have a wrapper for `MemcpyBatchAsync` that is capable of avoiding the conditional compilation (`CUDA_VERSION`) in the header to maintain type consistency for library users.
- We want to  have a wrapper that fully preserves the function signature of the batch copy API, so that at the call site, callers still have the freedom of having a custom `CUmemcpyAttributes` parameter.

All the hassle arises from the fact that the cuda shim header is currently publicly exposed. Ideally, it (and the cufile shim) should be moved to the implementation directory. There's still quite a few knots to untie to make that happen.

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,438 @@
...
+ *
+ * @tparam Integer Signed integer type
+ * @param p An address
+ * @param v Change of `p` in bytes
+ * @return A new address as a result of applying `v` on `p`
+ *
+ * @note It is UB in C++ when the initial pointer and the resulting pointer do not point to the
+ * elements from the same array. This UB is considered acceptable here due to lack of a better
+ * alternative.
+ */
```
kingcrimsontianyu, line 62: Further reading: [How to check if a pointer is in a range of memory](https://devblogs.microsoft.com/oldnewthing/20170927-00/?p=97095) by Raymond Chen, Microsoft

cpp/include/kvikio/shim/cuda.hpp
```diff
@@ -15,11 +15,67 @@
...
+   * @param args Arguments to be passed
+   * @return CUDA driver API error code
+   * @exception std::bad_any_cast if any argument type does not exactly match the parameter type of
+   * the contained callable.
+   */
+  template <typename... Args>
+  CUresult operator()(Args... args)
+  {
+    using T = std::function<CUresult(Args...)>;
+    return std::any_cast<T&>(_callable)(args...);
```
jakirkham, line 71: Should we check `_callable.has_value()` before casting?

cpp/src/file_utils.cpp
```diff
@@ -169,6 +169,18 @@ int open_fd(std::string const& file_path, std::string const& flags, bool o_direc
...
 }
 
+[[nodiscard]] std::size_t get_file_size(std::string const& file_path)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  std::string const flags{"r"};
+  bool const o_direct{false};
+  mode_t const mode{FileHandle::m644};
+  auto fd     = open_fd(file_path, flags, o_direct, mode);
+  auto result = get_file_size(fd);
```
jakirkham, line 179: Can we just use [`std::filesystem::file_size`]( https://en.cppreference.com/w/cpp/filesystem/file_size.html )?

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+  while (size > 0) {
+    detail::do_not_optimize_away_read(aligned_addr);
+    if (size >= page_size) {
+      aligned_addr = detail::pointer_add(aligned_addr, page_size);
+      size -= page_size;
+      touched_bytes += page_size;
+    } else {
+      touched_bytes += size;
+      break;
```
jakirkham: Assuming this is to prefetch the data

First would recommend using `__builtin_prefetch` instead of rolling our own

Second unless we can demonstrate this is needed (benchmark?), usually the compiler and processor will do reasonable things. If they don't, it's possible there are other things we should consider in our code before reaching to prefetching

Would recommend reading [this blogpost on prefetching]( https://lemire.me/blog/2018/04/30/is-software-prefetching-__builtin_prefetch-useful-for-performance/ )

jakirkham (commented):
Thanks Tianyu! 🙏

Had a few comments below

cpp/include/kvikio/shim/cuda.hpp
kingcrimsontianyu: Yes. Thanks. Done.
`std::any_cast()` would throw "bad any_cast" exception on an empty `std::any` object, but empty check is a good defensive practice. Ideally `KVIKIO_EXPECT(_callable.has_value(), ...)` macro could be used here, but there's some dependency issue between the CUDA shim header and the error handling header that I'd like to untangle in a separate PR in the future. So a simple `throw` is used here.

cpp/src/file_utils.cpp
kingcrimsontianyu: The POSIX API used in `get_file_size(fd)` and the C++ API `file_size(path)` should be functionally equivalent here, since the C++ API also checks the field `st_size` as mentioned in the reference. Either will do.

cpp/src/mmap.cpp
kingcrimsontianyu: Software prefetch with `__builtin_prefetch` and prefault serve different purposes.

`__builtin_prefetch` is to bring data from memory to cache ahead of time to reduce processor stall and hide memory access latency. I did [quite a bit of tests on prefetch years ago](https://tianyuliukingcrimson.com/wp-content/uploads/2014/08/2017_nuclear-science-and-engineering.pdf) on different processors.

Prefault exists in the context of mapped memory, when access to a virtual memory address is made, but the data are still on the disk only, causing a page fault to bring the data to the physical memory. This prefault function is an effective optimization that the compiler cannot do automatically. Without it, when the GPU copy engine (which is a limited resource) copies data from the host to device, it will absorb all the cost of page fault, causing overall slowdown. This was verified on GH200.

I think the two techniques do have overlap, though. Prefetching a mapped region should still cause page faults; and prefaulting a mapped region using the approach in this PR brings the data to the register where the cache will be used. The problem is `__builtin_prefetch` is a hint and does not guarantee data availability in physical host memory when the H2D copy starts, but prefault does.

cpp/include/kvikio/defaults.hpp
```diff
@@ -367,6 +368,23 @@ class defaults {
    * @param status_codes The HTTP status codes to retry.
    */
   static void set_http_status_codes(std::vector<int> status_codes);
+
+  /**
+   * @brief Get the default mmap task size used for parallel memory-mapped I/O.
+   *
+   * Set the default value using `kvikio::default::set_mmap_task_size()` or by setting
+   * the `KVIKIO_MMAP_TASK_SIZE` environment variable. If not set, the default value is 4 MiB.
```
vuule: I prefer not to include the exact value in comments. Very easy to become outdated.
```suggestion
   * the `KVIKIO_MMAP_TASK_SIZE` environment variable.
```

cpp/tests/test_mmap.cpp
```diff
@@ -0,0 +1,335 @@
...
+        [[maybe_unused]] auto mmap_handle = kvikio::MmapHandle(_filepath, "");
+      }
+    },
+    ThrowsMessage<std::invalid_argument>(HasSubstr("Unknown file open flag")));
+
+  // Invalid file open flag
+  EXPECT_THAT(
+    [=] {
+      {
+        [[maybe_unused]] auto mmap_handle = kvikio::MmapHandle(_filepath, "z");
```
vuule:
```suggestion
        kvikio::MmapHandle(_filepath, "z");
```

cpp/tests/test_mmap.cpp
```diff
@@ -0,0 +1,335 @@
...
+  using value_type = decltype(_host_buf)::value_type;
+};
+
+TEST_F(MmapTest, invalid_file_open_flag)
+{
+  // Empty file open flag
+  EXPECT_THAT(
+    [=] {
+      {
+        [[maybe_unused]] auto mmap_handle = kvikio::MmapHandle(_filepath, "");
```
vuule:
```suggestion
        kvikio::MmapHandle(_filepath, "");
```

cpp/tests/test_mmap.cpp
```diff
@@ -0,0 +1,335 @@
...
+    mmap_handle = kvikio::MmapHandle(_filepath, "r");
+    EXPECT_TRUE(!mmap_handle.closed());
+    do_test(mmap_handle);
+  }
+
+  {
+    kvikio::MmapHandle mmap_handle_1(_filepath, "r");
+    kvikio::MmapHandle mmap_handle_2{std::move(mmap_handle_1)};
+    EXPECT_TRUE(mmap_handle_1.closed());
+    EXPECT_TRUE(!mmap_handle_2.closed());
```
vuule:
```suggestion
    EXPECT_FALSE(mmap_handle_2.closed());
```

cpp/tests/test_mmap.cpp
```diff
@@ -0,0 +1,335 @@
...
+      EXPECT_EQ(_host_buf[i], out_host_buf[i]);
+    }
+    EXPECT_EQ(read_size, _file_size);
+  };
+
+  {
+    kvikio::MmapHandle mmap_handle{};
+    EXPECT_TRUE(mmap_handle.closed());
+    mmap_handle = kvikio::MmapHandle(_filepath, "r");
+    EXPECT_TRUE(!mmap_handle.closed());
```
vuule:
```suggestion
    EXPECT_FALSE(mmap_handle.closed());
```

cpp/tests/utils/utils.hpp
```diff
@@ -119,6 +119,7 @@ class DevBuffer {
   DevBuffer(std::size_t nelem) : nelem{nelem}, nbytes{nelem * sizeof(std::int64_t)}
   {
     KVIKIO_CHECK_CUDA(cudaMalloc(&ptr, nbytes));
+    KVIKIO_CHECK_CUDA(cudaMemset(ptr, 0, nbytes));
```
vuule, line 122: why is this needed now?

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,211 @@
...
+   * @param initial_file_offset File offset of the mapped region
+   * @param mode Access mode
+   * @param map_flags Flags to be passed to the system call `mmap`. See `mmap(2)` for details
+   */
+  MmapHandle(std::string const& file_path,
+             std::string const& flags                = "r",
+             std::optional<std::size_t> initial_size = std::nullopt,
+             std::size_t initial_file_offset         = 0,
+             mode_t mode                             = FileHandle::m644,
+             std::optional<int> map_flags            = std::nullopt);
```
vuule: are there tests for this parameter?

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,211 @@
...
+  MmapHandle(MmapHandle&& o) noexcept;
+  MmapHandle& operator=(MmapHandle&& o) noexcept;
+  ~MmapHandle() noexcept;
+
+  /**
+   * @brief Size in bytes of the mapped region when the mapping handle was constructed
+   *
+   * @return Initial size of the mapped region
+   */
+  std::size_t initial_size() const noexcept;
```
vuule:
```suggestion
  [[nodiscard]] std::size_t initial_size() const noexcept;
```

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * @brief File offset of the mapped region when the mapping handle was constructed
+   *
+   * @return Initial file offset of the mapped region
+   */
+  std::size_t initial_file_offset() const noexcept;
```
vuule:
```suggestion
  [[nodiscard]] std::size_t initial_file_offset() const noexcept;
```

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,211 @@
...
+   * @exception std::invalid_argument if the size is given but is 0
+   * @exception std::runtime_error if the mapping handle is closed
+   *
+   * @note The `std::future` object's `wait()` or `get()` should not be called after the lifetime of
+   * the MmapHandle object ends. Otherwise, the behavior is undefined.
+   */
+  std::future<std::size_t> pread(void* buf,
+                                 std::optional<std::size_t> size = std::nullopt,
+                                 std::size_t file_offset         = 0,
+                                 std::size_t mmap_task_size      = defaults::mmap_task_size());
```
vuule: I think we should use the same naming as the other handles.
```suggestion
                                 std::size_t task_size      = defaults::mmap_task_size());
```

vuule (commented):
partial review - just some minor comments for now

cpp/tests/utils/utils.hpp
kingcrimsontianyu: There's `to_vector()` method that copies data to a host vector. So purely from the language perspective, I hope to keep device data cleanly initialized.

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: I missed that. Done!

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: Done. Also applied to `mmap.cpp` except for the exception message.

cpp/include/kvikio/defaults.hpp
```diff
@@ -367,6 +368,23 @@ class defaults {
    * @param status_codes The HTTP status codes to retry.
    */
   static void set_http_status_codes(std::vector<int> status_codes);
+
+  /**
+   * @brief Get the default mmap task size used for parallel memory-mapped I/O.
+   *
+   * Set the default value using `kvikio::default::set_mmap_task_size()` or by setting
+   * the `KVIKIO_MMAP_TASK_SIZE` environment variable.
```
vuule: thoughts on having a single task size env var? do you think there's a scenario where different files are read using different approaches in the same process?

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+namespace kvikio {
+
+namespace detail {
+/**
+ * @brief Prevent the compiler from optimizing away the read of a byte from a given address
+ *
+ * @param addr The address to read from
+ */
+void do_not_optimize_away_read(void* addr)
```
vuule: Optional: How about `unoptimizable_read_byte` or something like that? I slightly don't like that read is not the verb in the current name.

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+ * @note It is UB in C++ when the initial pointer and the resulting pointer do not point to the
+ * elements from the same array. This UB is considered acceptable here due to lack of a better
+ * alternative.
+ */
+template <typename Integer>
+void* pointer_add(void* p, Integer v)
+{
+  static_assert(std::is_integral_v<Integer>);
+  return static_cast<std::byte*>(p) + v;
+}
```
vuule, line 68:
```suggestion
template <std::integral Integer>
void* pointer_add(void* p, Integer v)
{
  return static_cast<std::byte*>(p) + v;
}
```

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+/**
+ * @brief Whether the current device supports address translation service (ATS), whereby the CPU and
+ * GPU share a single page table.
+ *
+ * @return Boolean answer
+ */
+bool is_ats_available()
+{
+  // Memoize the ATS availability record of all devices
+  static auto ats_availability = []() -> auto {
```
vuule:
```suggestion
  static auto const ats_availability = []() -> auto {
```

vuule (commented):
few more small comments

cpp/src/mmap.cpp
kingcrimsontianyu: Currently KvikIO still uses C++17. There's this PR that changes it to C++20 (https://github.com/rapidsai/kvikio/pull/749), but there was CI failures to be looked into. Just ran the CI again and we'll see where it leads.

cpp/src/mmap.cpp
kingcrimsontianyu: Changed to `disable_read_optimization`

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: Done. Since remote I/O also uses `defaults::task_size`, I've used it for mmap for consistency.
In a special scenario where different handles are used, each requiring a different task size, users should explicitly pass it to the `pread` method.

cpp/src/mmap.cpp
vuule: My bad, thought that the switch to 20 has already been made. Please ignore the suggestion.

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,212 @@
...
+namespace kvikio {
+
+/**
+ * @brief Handle of a memory-mapped file
+ */
+class MmapHandle {
+ private:
+  void* _buf{};
+  std::size_t _initial_size{};
+  std::size_t _initial_file_offset{};
```
vuule: should these be _initial_map_size and _initial_map_offset?

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,212 @@
...
+   * @param initial_size Size in bytes of the mapped region. If not specified, map the region
+   * starting from `initial_file_offset` to the end of file
+   * @param initial_file_offset File offset of the mapped region
+   * @param mode Access mode
+   * @param map_flags Flags to be passed to the system call `mmap`. See `mmap(2)` for details
+   */
+  MmapHandle(std::string const& file_path,
+             std::string const& flags                = "r",
+             std::optional<std::size_t> initial_size = std::nullopt,
+             std::size_t initial_file_offset         = 0,
```
vuule: these can be map_size and map_offset. The fact that we might map a slightly different region is an implementation detail that is not exposed, AFAICT.

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,213 @@
...
+   *
+   * @param size Size in bytes to read. If not specified, set it to the bytes from `file_offset` to
+   * the end of file
+   * @param file_offset File offset
+   * @return Adjusted size in bytes to read
+   *
+   * @exception std::out_of_range if the read region specified by `file_offset` and `size` is
+   * outside the initial region specified when the mapping handle was constructed
+   * @exception std::invalid_argument if the size is given but is 0
+   * @exception std::runtime_error if the mapping handle is closed
```
madsbk:
```suggestion
   * @throw std::out_of_range if the read region specified by `file_offset` and `size` is
   * outside the initial region specified when the mapping handle was constructed
   * @throw std::invalid_argument if the size is given but is 0
   * @throw std::runtime_error if the mapping handle is closed
```

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,213 @@
...
+   * @param buf Address of the host or device memory (destination buffer)
+   * @param size Size in bytes to read. If not specified, read starts from `file_offset` to the end
+   * of file
+   * @param file_offset File offset
+   * @return Number of bytes that have been read
+   *
+   * @exception std::out_of_range if the read region specified by `file_offset` and `size` is
+   * outside the initial region specified when the mapping handle was constructed
+   * @exception std::invalid_argument if the size is given but is 0
+   * @exception std::runtime_error if the mapping handle is closed
```
madsbk:
```suggestion
   * @throw std::out_of_range if the read region specified by `file_offset` and `size` is
   * outside the initial region specified when the mapping handle was constructed
   * @throw std::invalid_argument if the size is given but is 0
   * @throw std::runtime_error if the mapping handle is closed
```

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,213 @@
...
+   * @param size Size in bytes to read. If not specified, read starts from `file_offset` to the end
+   * of file
+   * @param file_offset File offset
+   * @param task_size Size of each task in bytes
+   * @return Future that on completion returns the size of bytes that were successfully read.
+   *
+   * @exception std::out_of_range if the read region specified by `file_offset` and `size` is
+   * outside the initial region specified when the mapping handle was constructed
+   * @exception std::invalid_argument if the size is given but is 0
+   * @exception std::runtime_error if the mapping handle is closed
```
madsbk:
```suggestion
   * @throw std::out_of_range if the read region specified by `file_offset` and `size` is
   * outside the initial region specified when the mapping handle was constructed
   * @throw std::invalid_argument if the size is given but is 0
   * @throw std::runtime_error if the mapping handle is closed
```

cpp/include/kvikio/shim/cuda.hpp
```diff
@@ -15,11 +15,70 @@
...
+
+  /**
+   * @brief Invoke the container callable
+   *
+   * @tparam Args Types of the argument. Must exactly match the parameter types of the contained
+   * callable. For example, if the parameter is `std::size_t*`, an argument of `nullptr` must be
+   * explicitly cast to `std::size_t*`.
+   * @param args Arguments to be passed
+   * @return CUDA driver API error code
+   * @exception std::bad_any_cast if any argument type does not exactly match the parameter type of
```
madsbk, line 61:
```suggestion
   * @throw std::bad_any_cast if any argument type does not exactly match the parameter type of
```

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,213 @@
...
+#include <future>
+
+#include <kvikio/defaults.hpp>
+#include <kvikio/file_handle.hpp>
+#include <optional>
+
+namespace kvikio {
+
+/**
+ * @brief Handle of a memory-mapped file
```
madsbk, line 28: It would be good some more details and when it use full.

cpp/src/utils.cpp
```diff
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
+#include "utils.hpp"
```
madsbk:
```suggestion
#include <kvikio/utils.hpp>
```

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+ * @note It is UB in C++ when the initial pointer and the resulting pointer do not point to the
+ * elements from the same array. This UB is considered acceptable here due to lack of a better
+ * alternative.
+ */
+template <typename Integer>
+void* pointer_add(void* p, Integer v)
+{
+  static_assert(std::is_integral_v<Integer>);
+  return static_cast<std::byte*>(p) + v;
+}
```
madsbk, line 68: What about something like the following, I think that is implementation-defined, but not undefined?

```c++
template <typename Integer>
void* pointer_add(void* p, Integer v)
{
  static_assert(std::is_integral_v<Integer>);

  auto addr = reinterpret_cast<std::uintptr_t>(p);
  addr += static_cast<std::uintptr_t>(v);
  return reinterpret_cast<void*>(addr);
}
```

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+//                     |--> _map_addr
+//
+// (3) |<---------------------file_offset--------------------->|<--size-->|
+//                                |--> _buf
+//                                                             |--> start_addr
+//                                                     |--> start_aligned_addr
+// (0): Layout of the file-backed memory mapping if the whole file were mapped
+// (1): At mapping handle construction time, the member `_initial_file_offset` and `_initial_size`
+// determine the mapped region (2): `_map_addr` is the page aligned address returned by `mmap`.
+// `_map_offset` is the adjusted offset
```
madsbk:
```suggestion
// `_map_offset` is the adjusted offset.
```

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+// (3): At read time, the argument `file_offset` and `size` determine the region to be read. This
+// region must be a subset of the one defined at mapping handle construction time
```
madsbk:
```suggestion
// region must be a subset of the one defined at mapping handle construction time.
```

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+                       std::size_t initial_file_offset,
+                       mode_t mode,
+                       std::optional<int> map_flags)
+  : _initial_file_offset(initial_file_offset),
+    _initialized{true},
+    _file_wrapper(file_path, flags, false /* o_direct */, mode)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+
+  _file_size = get_file_size(_file_wrapper.fd());
```
madsbk, line 282: You don't want this  is the initializer list?

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+  _initial_file_offset = std::exchange(o._initial_file_offset, {});
+  _file_size           = std::exchange(o._file_size, {});
+  _map_offset          = std::exchange(o._map_offset, {});
+  _map_size            = std::exchange(o._map_size, {});
+  _map_addr            = std::exchange(o._map_addr, {});
+  _initialized         = std::exchange(o._initialized, {});
+  _map_protection      = std::exchange(o._map_protection, {});
+  _map_flags           = std::exchange(o._map_flags, {});
+  _file_wrapper        = std::exchange(o._file_wrapper, {});
+
```
madsbk:
```suggestion
```

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,443 @@
...
+                     actual_size,
+                     file_offset,
+                     task_size,
+                     0,  // dst buffer offset initial value
+                     call_idx,
+                     nvtx_color);
+}
+
+std::size_t MmapHandle::validate_and_adjust_read_args(std::optional<std::size_t> const& size,
+                                                      std::size_t& file_offset)
```
madsbk:
```suggestion
                                                      std::size_t file_offset)
```

cpp/include/kvikio/mmap.hpp
```diff
@@ -0,0 +1,213 @@
...
+   * `global_dst_buf + buf_offset`.
+   *
+   * @param global_dst_buf Address of the host or device memory (destination buffer)
+   * @param global_src_buf Address of the host memory (source buffer)
+   * @param size Size in bytes to read
+   * @param buf_offset Offset for both `global_dst_buf` and `global_src_buf`
+   * @param is_dst_buf_host_mem Whether the destination buffer is host memory or not
+   * @param ctx CUDA context when the destination buffer is not host memory
+   */
+  void read_impl(void* global_dst_buf,
```
madsbk: Can we make this a function and move it to .cpp or would that require too many arguments?

madsbk (commented):
Overall looks great!

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: The Doxygen doc lists [throw](https://www.doxygen.nl/manual/commands.html#cmdthrow) as the synonym of [exception](https://www.doxygen.nl/manual/commands.html#cmdexception), and [throws](https://www.doxygen.nl/manual/commands.html#cmdthrows) as the equivalent to `throw`, so I think using `exception` may be the present-day norm, but it may also simply because the doc is using the alphabetical order and provides details only for the first command among its equivalent forms.
On a side note, we have used both `throw` and `throws`, and we may stick to `throw` for better consistency.

madsbk: I am fine with `@exception` as long as we use it everywhere.

kingcrimsontianyu: Done. `@throw` and `@throws` have been replaced by `@exception` throughout KvikIO.

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: Thanks. Agreed. Done.

cpp/src/utils.cpp
kingcrimsontianyu: Actually, there is a `#include <kvikio/utils.hpp>` down below. `#include "utils.hpp"` is a private header I've added in this PR to the `src` directory. I'd like to minimize the symbols and utility functions we expose publicly.
I'm open to any suggesting renaming (`utils_private.hpp`) or relocating (`detail/utils.hpp` or `internal/utils.hpp`) the private header `utils.hpp`.

cpp/src/utils.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cstring>
+
+namespace kvikio {
```
madsbk: Move to `kvikio::detail`?

cpp/src/utils.cpp
madsbk: I think we should move them to `include` -- maybe `detail/utils.hpp`. As long as we use the `kvikio::detail` namespace, the headers are _marked_ private.

kingcrimsontianyu: Thanks. I've moved private header contents to `include/detail/utils.hpp`.

cpp/src/mmap.cpp
kingcrimsontianyu: I think either is acceptable. The UB won't cause trouble in practice. Also, I'm not sure for pointer-to-int and int-to-pointer, whether it constitutes a UB if the conversion is not a round-trip to a fixed object in the `uintptr_t` approach.

cpp/src/utils.hpp
kingcrimsontianyu: Done.

kingcrimsontianyu: Do you have any thought on whether to place this private header to `include/detail/utils.hpp` or `src/detail/utils.hpp`? I'm interpreting the `include` directory as a place for public headers so leaning towards the 2nd option. Ideally, in a long run, I'd like to separate the shim libraries, KvikIO-specific macros, and all the private contents from the public interface and put them into `src/detail`. But I'm also fine putting all the internal/detail header contents to `include/detail`. Please let me know your thoughts. @madsbk @vuule

cpp/src/mmap.cpp
kingcrimsontianyu: I'm hesitant to place it in the member initializer because that would require `_file_wrapper` be declared prior to `_file_size` in the class, and I'd like to avoid this implicit ordering requirement.

cpp/src/mmap.cpp
kingcrimsontianyu: Thanks. Done.

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: Thanks. Done. I've made `read_impl` and `perform_prefault` free functions inside `.cpp`.

cpp/src/mmap.cpp
madsbk: > whether it constitutes a UB if the conversion is not a round-trip to a fixed object in the uintptr_t approach.

Ahh, you are right. 
The following is legal if the returned pointer remain within the bounds of the same allocated object (or one-past-the-end) as `p`:

```c++
void* pointer_add(void* p, std::ptrdiff_t v)
{
    return static_cast<std::byte*>(p) + v;
}
```

cpp/src/utils.hpp
madsbk: Personally, I find it easier to read and understand libraries that include everything, incl. private headers, under the public `include` directory, using namespaces to indicate internal/private status. Being able to read and easily find the private code can be very helpful for understanding how the library works.

kingcrimsontianyu: @madsbk Thanks! I'll use this convention when reorganizing headers.

kingcrimsontianyu:
The comments for `pointer_add` have been modified to clarify.

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,466 @@
...
+/**
+ * @brief The distance in bytes between pointer `p1` and `p2`
+ *
+ * @param p1 The first pointer
+ * @param p2 The second pointer
+ * @return Signed result of (`p1` - `p2`). Both pointers are cast to std::byte* before subtraction.
+ *
+ * @note It is UB in C++ when the two pointers engaged in subtraction do not point to the elements
+ * from the same array. This UB is considered acceptable here due to lack of a better alternative.
+ */
```
madsbk: Also update the note here?

madsbk (approved):
I am still not convinced :) Surely, we are allowed to use pointer arithmetic to access a memory allocation!?!?

But in any case, the implementation looks good!

cpp/src/mmap.cpp
kingcrimsontianyu: Thanks. Updated, by referencing another passage from the standard.

cpp/src/mmap.cpp
```diff
@@ -0,0 +1,467 @@
...
+ *
+ * @param dst_buf Address of the host or device memory (destination buffer)
+ * @param src_mapped_buf Address of the host memory (source buffer)
+ * @param size Size in bytes to read
+ * @param buf_offset Offset for both `dst_buf` and `src_mapped_buf`
+ * @param is_dst_buf_host_mem Whether the destination buffer is host memory or not
+ * @param ctx CUDA context when the destination buffer is not host memory
+ */
+void read_impl(void* dst_buf,
+               void* src_mapped_buf,
```
vuule, line 177:
```suggestion
               void const* src_mapped_buf,
```

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: All `initial_size` and `initial_file_offset` have been replaced by `initial_map_size` and `initial_map_offset`. To maintain naming consistency, for `read/pread`, all `file_offset` have been replaced by `offset`. Comments and test updated as well.

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: Done. The prefix `initial` is kept to distinguish from the size and offset in the `read/pread` functions.

cpp/src/mmap.cpp
kingcrimsontianyu: This would require the pointer manipulation utility `void* pointer_add(void* p, Integer v)` to change accordingly (compile error otherwise), and I think it's a bit too much hassle. Since `read_impl` is not in the public header, I think we can skip the good constness practice for convenience.

vuule (approved):

PR #741, opened 2025-06-01
Add the utility function to clear page cache

kingcrimsontianyu:
This PR introduces a utility function to clear page cache in C++ and Python.


cpp/src/file_utils.cpp
```diff
@@ -209,4 +212,16 @@ std::pair<std::size_t, std::size_t> get_page_cache_info(int fd)
...
   return {num_pages_in_page_cache, num_pages};
 }
+
+bool clear_page_cache(bool reclaim_dentries_and_inodes, bool clear_dirty_pages)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  if (clear_dirty_pages) { sync(); }
+  std::string param = reclaim_dentries_and_inodes ? "3" : "1";
+  std::stringstream ss;
+  ss << "echo " << param << " | sudo tee /proc/sys/vm/drop_caches 1>/dev/null";
```
madsbk: Can we also try without sudo? Some like: https://github.com/rapidsai/cudf/blob/6bc515d8219538b104861b48f2b8822a53be841f/cpp/benchmarks/io/cuio_common.cpp#L230

kingcrimsontianyu: I think cuDF's implementation still relies on sudo or superuser privilege. The logic is:
1. run the cache clearing command in a shell
1. run the sudo-ed command in a shell
1. if both of them fail, throw an exception

Consequently:
- General case
  - For superuser, 1 succeeds.
  - For non-superuser, 1 fails, 2 succeeds once providing the right password.
- Docker container
  - Without `--priviledged` argument, both fail regardless of superuser or not.
  - With `--priviledged`, the same with the general case (except that no password prompt for non-superuser).

So this PR simplifies 1 and 2 to a single `sudo` command, and also returns boolean to indicate success/failure instead of throwing on failure.

PS: I actually had another implementation earlier that does not use `sudo` at all: https://github.com/rapidsai/kvikio/pull/741/commits/606b233a32abcf4b5054cff03870a74aa756a1ae#diff-e00c8157cb029cd5327cfe9b8b834601b40dcfef0b847bc3d6272e3c7b5c3c1cR216
This method does not have the problem in `system()` or `popen()` (they create a separate process and run the shell command, which adds to security vulnerability). But it also lacks the convenience: users need to run the entire process with `sudo` in order to clear the cache, as opposed to only `sudo` the cache clearing part in the shell process.

PSS: I also had an experimental snippet that does something like below, with the hope that I can achieve temporary privilege elevation for non-superusers:
```
perform tasks
{
    // temporarily elevate privilege
    // stash the real uid and real gid
    SuperUserRaiiContext ctx;
    perform tasks
} //  set to the original real uid and real gid
perform tasks
```
But I scratched this idea too, because it still requires users to run the process with elevated privilege, and as a result the code prior to `SuperUserRaiiContext` is still privileged.

cpp/src/file_utils.cpp
```diff
@@ -209,4 +212,16 @@ std::pair<std::size_t, std::size_t> get_page_cache_info(int fd)
   SYSCALL_CHECK(munmap(addr, file_size));
   return {num_pages_in_page_cache, num_pages};
 }
+
+bool clear_page_cache(bool reclaim_dentries_and_inodes, bool clear_dirty_pages)
```
vuule, line 217: when would we want to call this without clearing the dirty pages?

kingcrimsontianyu: I'd say these options are for curious minds who wonder the respective effect on page cache :smiley:

cpp/src/file_utils.cpp
kingcrimsontianyu: @vuule explained that the non-sudo command is used on specially configured machines where unprivileged users have no access to `sudo` executable but can run `sysctl`. This complicates things, as the sysctl command does not return non-zero code on insufficient permission failure, unlike the other approach `echo 3 | tee /proc/sys/vm/drop_caches` which does. So I changed to cuDF's implementation that uses stderr to check failure instead.

python/kvikio/kvikio/__init__.py
```diff
@@ -14,14 +14,15 @@
...
-from kvikio.cufile import CuFile, get_page_cache_info
+from kvikio.cufile import CuFile, clear_page_cache, get_page_cache_info
 from kvikio.remote_file import RemoteFile, is_remote_file_available
 
 __all__ = [
     "__git_commit__",
     "__version__",
     "CuFile",
     "get_page_cache_info",
+    "clear_page_cache",
```
bdice: Typically we try to keep `__all__` alphabetized. Could you sort these lines?

kingcrimsontianyu: Thanks. Sorted.

cpp/include/kvikio/file_utils.hpp
```diff
@@ -180,4 +180,23 @@ std::pair<std::size_t, std::size_t> get_page_cache_info(std::string const& file_
...
+ * @param reclaim_dentries_and_inodes Whether to free reclaimable slab objects which include
+ * dentries and inodes.
+ * - If `true`, equivalent to executing `echo 3 > /proc/sys/vm/drop_caches`;
+ * - If `false`, equivalent to executing `echo 1 > /proc/sys/vm/drop_caches`.
+ * @param clear_dirty_pages Whether to trigger the writeback process to clear the dirty pages. If
+ * `true`, `sync` will be called prior to cache clearing.
+ * @return Whether the page cache has been successfully cleared
+ *
+ * @note This function creates a child process and executes the cache clearing shell command with
+ * `sudo`. Superuser privilege is therefore generally needed for the function to return `true`.
```
madsbk: Update to say it first try without sudo?

kingcrimsontianyu: Thanks. The comments have been updated.

madsbk (approved):

PR #742, opened 2025-06-02
Support file-backed mapping (2/n): host side read Python binding

kingcrimsontianyu:
# work in progress

Partially addresses https://github.com/rapidsai/kvikio/issues/530


python/kvikio/kvikio/mmap.py
```diff
@@ -0,0 +1,171 @@
...
+        """Whether the mapping handle is closed
+
+        Returns
+        -------
+        bool
+            Boolean answer.
+        """
+        return self._handle.closed()
+
+    def read(self, buf, size: Optional[int] = None, offset: int = 0) -> int:
```
kingcrimsontianyu: Need to figure out an accurate type hint for `buf`, given this: https://github.com/rapidsai/kvikio/blob/branch-25.10/python/kvikio/kvikio/_lib/arr.pyx#L314 . Or may simply use `Any` to defer type checking to runtime.

TomAugspurger: Would `collections.abc.Buffer` be correct? https://docs.python.org/3/library/collections.abc.html#collections.abc.Buffer. That was added in Python 3.12, so I think using `Any` for now is fine.

python/kvikio/kvikio/mmap.py
```diff
@@ -0,0 +1,171 @@
...
+        buf : buffer-like or array-like
+            Address of the host or device memory (destination buffer).
+        size : int, optional
+            Size in bytes to read. If not specified, read starts from ``offset``
+            to the end of file.
+        offset : int, optional
+            File offset. Default is 0.
+        task_size : int, optional
+            Size of each task in bytes for parallel execution. If None, uses
+            the default task size from kvikio.defaults.task_size().
```
TomAugspurger:
```suggestion
            the default task size from `:func:`kvikio.defaults.task_size`.
```

python/kvikio/kvikio/_lib/mmap.pyx
```diff
@@ -0,0 +1,116 @@
...
+cdef class InternalMmapHandle:
+    cdef CppMmapHandle _handle
+
+    def __init__(self, file_path: os.PathLike,
+                 flags: str = "r",
+                 initial_map_size: Optional[int] = None,
+                 initial_map_offset: int = 0,
+                 mode: int = stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH,
+                 map_flags: Optional[int] = None):
+        if not os.path.exists(file_path):
```
TomAugspurger, line 48: Should this just be for `mode="r"`?

Oh, I see that `mode="w"` isn't implemented yet. Can we explicitly check the `mode` and raise a `NotImplementedError` if it's not `r`?

Edit: everywhere I said `mode` should probably be `flags` (with python's builtin `open` I think this would be called "mode". But I guess "flags" matches kvikio's other methods).

python/kvikio/kvikio/_lib/mmap.pyx
```diff
@@ -0,0 +1,116 @@
...
+            raise RuntimeError("Unable to open file")
+
+        cdef optional[size_t] cpp_initial_map_size
+        if initial_map_size is None:
+            cpp_initial_map_size = nullopt
+        else:
+            cpp_initial_map_size = <size_t>(initial_map_size)
+
+        path_bytes = str(pathlib.Path(file_path)).encode()
```
TomAugspurger: I think we're just using `pathlib.Path` for converting to the filesystem representaiton? If so, we can use `os.fspath` directly:

```suggestion
        path_bytes = str(os.fspath(file_path)).encode()
```

TomAugspurger: I think `str(bytestring).encode()` quite does what we want.

```python
In [34]: str(b'\xc3\xa9').encode()
Out[34]: b"b'\\xc3\\xa9'"

In [35]: (b'\xc3\xa9').decode().encode()
Out[35]: b'\xc3\xa9'
```

but I'm also not too familiar with how the different filesystems represent paths.

python/kvikio/tests/test_mmap.py
```diff
@@ -0,0 +1,188 @@
...
+
+import kvikio.defaults
+
+cupy = pytest.importorskip("cupy")
+numpy = pytest.importorskip("numpy")
+
+
+def test_no_file(tmp_path):
+    with pytest.raises(RuntimeError, match=r".*Unable to open file.*"):
+        nonexistent_file = tmp_path / "nonexistent_file"
```
TomAugspurger:
```suggestion
    nonexistent_file = tmp_path / "nonexistent_file"
        with pytest.raises(RuntimeError, match=r".*Unable to open file.*"):
```

Prefer to keep the code under `pytest.raises` as small as possible.

python/kvikio/kvikio/_lib/mmap.pyx
```diff
@@ -0,0 +1,116 @@
...
+
+
+cdef extern from "<kvikio/mmap.hpp>" namespace "kvikio" nogil:
+    cdef cppclass CppMmapHandle "kvikio::MmapHandle":
+        CppMmapHandle() noexcept
+        CppMmapHandle(string file_path, string flags, optional[size_t] initial_map_size,
+                      size_t initial_map_offset, fcntl.mode_t mode,
+                      optional[int] map_flags) except +
+        size_t initial_map_size()
+        size_t initial_map_offset()
```
madsbk:
```suggestion
        size_t initial_map_size() noexcept
        size_t initial_map_offset() noexcept
```

python/kvikio/kvikio/_lib/mmap.pyx
```diff
@@ -0,0 +1,116 @@
...
+        size_t file_size() except +
+        void close()
+        bool closed()
```
madsbk:
```suggestion
        void close() noexcept
        bool closed() noexcept
```

python/kvikio/kvikio/mmap.py
kingcrimsontianyu: Thanks. Done!

python/kvikio/tests/test_mmap.py
kingcrimsontianyu: Thanks. Done.

python/kvikio/kvikio/_lib/mmap.pyx
kingcrimsontianyu: Good catch. Done!

python/kvikio/kvikio/_lib/mmap.pyx
kingcrimsontianyu: Done.

python/kvikio/kvikio/_lib/mmap.pyx
kingcrimsontianyu: I'd like to let C++ throw the exception on the `"w"` flag. But thanks for raising this issue. There is actually a bug in the C++ implementation that prevents the exception being thrown properly on `"w"`. This has just been fixed. Now  with this:
```python
# Python snippet
import kvikio
h = kvikio.Mmap("LICENSE", "w")
```
The C++ library throws as expected
```
  File "/home/coder/kvikio/test.py", line 2, in <module>
    h = kvikio.Mmap("LICENSE", "w")
  File "/home/coder/kvikio/python/kvikio/kvikio/mmap.py", line 48, in __init__
    self._handle = InternalMmapHandle(
                   ~~~~~~~~~~~~~~~~~~^
        file_path, flags, initial_map_size, initial_map_offset, mode, map_flags
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "kvikio/_lib/mmap.pyx", line 66, in kvikio._lib.mmap.InternalMmapHandle.__init__
ValueError: KvikIO failure at: /home/coder/kvikio/cpp/src/mmap.cpp:287: File-backed mmap write is not supported yet
```
`NotImplementedError` is a more accurate category. For simplicity I'd just let Cython automatically translate C++'s `std::invalid_argument` to Python's `ValueError`.

cpp/src/mmap.cpp
```diff
@@ -274,13 +274,25 @@ MmapHandle::MmapHandle(std::string const& file_path,
...
+    }
+    case 'w': {
+      KVIKIO_FAIL("File-backed mmap write is not supported yet", std::invalid_argument);
+    }
+    default: {
+      KVIKIO_FAIL("Unknown file open flag", std::invalid_argument);
+    }
+  }
+
+  _file_wrapper = FileWrapper(file_path, flags, false /* o_direct */, mode);
```
kingcrimsontianyu, line 294: Place `flags` check in front of `_file_wrapper` initialization, so that when `flags == "w"`, the exception is thrown, and the file will _not_ be overwritten already.

vuule (approved):

python/kvikio/kvikio/_lib/mmap.pyx
kingcrimsontianyu: Yeah. For now, I've used `Any` for `buf`.
It seems the type of `buf` depends on the type of `obj` here: https://github.com/rapidsai/kvikio/blob/branch-25.10/python/kvikio/kvikio/_lib/arr.pyx#L79 , and `collections.abc.Buffer` as you mentioned is a good type hint solution for host buffer. But I'm not quite familiar with the CUDA buffer type / CUDA buffer protocol / their relations with `collections.abc.Buffer` atm.

python/kvikio/kvikio/_lib/mmap.pyx
kingcrimsontianyu: Thanks. Good to learn about `os.fspath`. I've replaced `str(pathlib.Path(xxx))` with `str(os.fspath(xxx))` for both `mmap` and `file_handle`.

kingcrimsontianyu:
I clicked too quickly and accidentally requested a re-review from @vuule :smile:

python/kvikio/kvikio/_lib/file_handle.pyx
```diff
@@ -97,7 +96,7 @@ cdef class CuFile:
     def __init__(self, file_path, flags="r"):
         self._handle = move(
             FileHandle(
-                str.encode(str(pathlib.Path(file_path))),
+                str.encode(str(os.fspath(file_path))),
```
TomAugspurger: TIL about `os.fsencode`: https://docs.python.org/3/library/os.html#os.fsencode

I think this does exactly what we want here.
```suggestion
                os.fsencode(file_path),
```

python/kvikio/kvikio/_lib/file_handle.pyx
```diff
@@ -194,7 +193,7 @@ def get_page_cache_info(file: Union[os.PathLike, str, int, io.IOBase]) \
         -> tuple[int, int]:
     if isinstance(file, os.PathLike) or isinstance(file, str):
         # file is a path or a string object
-        path_bytes = str(pathlib.Path(file)).encode()
+        path_bytes = str(os.fspath(file)).encode()
```
TomAugspurger:
```suggestion
        path_bytes = os.fsencode(file)
```

TomAugspurger (approved):
Small suggestion around how we get the filepath as bytes, and then I think things look good.

python/kvikio/kvikio/_lib/file_handle.pyx
kingcrimsontianyu: Done. Even better.

python/kvikio/kvikio/_lib/file_handle.pyx
kingcrimsontianyu: Done.

python/kvikio/kvikio/_lib/file_handle.pyx
kingcrimsontianyu: Also `str.encode(str(flags))` is changed to `str(flags).encode()` for clarity and style consistency.

python/kvikio/kvikio/_lib/mmap.pyi
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
madsbk: In the current structure where each public cython module has a thin Python wrapper (e.g. `mmap.py`), I would not include a `pyi` as well.

madsbk (approved):

python/kvikio/kvikio/_lib/mmap.pyi
kingcrimsontianyu: Thanks. Removed.

PR #744, opened 2025-06-02
Add the conda dependency libnuma necessary for GDS I/O

kingcrimsontianyu:
This PR adds the conda dependency `libnuma` necessary for GDS I/O. For this PR, only CUDA 12 is considered.


conda/recipes/libkvikio/recipe.yaml
```diff
@@ -85,7 +86,9 @@ outputs:
       run:
         - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
         - if: linux
-          then: libcufile-dev
+          then:
+            - libcufile-dev
+            - libnuma
```
bdice: `libnuma` has run-exports: https://github.com/conda-forge/numactl-feedstock/blob/f376b1baaaba15880f5a093f8af5dcbd4c2ed14d/recipe/meta.yaml#L36-L37

With run-exports, a package in `host:` will automatically populate itself in `run:` with an ABI-compatible pinning. In the case of `libnuma`, that is `max_pin='x'` (same major version used to build the package).

However, we will want to _exclude_ the `libnuma` run-export from any packages that do not directly use libnuma. I will make some changes for that.

bdice (changes requested):
@gforsyth @jameslamb Maybe one of you will see what's wrong here. I have `libnuma` in the cache build `host` environment, and I am expecting its run exports to add `libnuma` to the run dependencies of `libkvikio`. However, I am not seeing that happen. The finalized run dependencies don't have `libnuma` listed. https://github.com/rapidsai/kvikio/actions/runs/15396373757/job/43318308018?pr=744#step:10:612

Am I forgetting some rule of rattler run-exports?

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -82,10 +83,12 @@ outputs:
       host:
         - cuda-version =${{ cuda_version }}
         - libcurl ${{ libcurl_version }}
+        - libnuma
       run:
         - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
```
gforsyth, line 88: I don't know why the `libnuma` run export isn't showing up -- I'll try to come up with a smaller reproducer.  Could be something with outputs that pin themselves as their own run exports?

In any case, since it isn't showing up in the build logs, we could explicitly add this under `run` here to make sure we get the dependencies needed at runtime:

```yaml
${{ pin_compatible("libnuma", upper_bound="x") }}
```

gforsyth: Ok, I put together a minimal reproducer for both `rattler-build` and `conda-build` and in both cases, the `libnuma` run export doesn't get added to the runtime dependencies.  In contrast, if you add `numactl` to the `host` dependencies, then `libnuma` _does_ get added.

Anyway, not a problem with `rattler` but a problem with the upstream recipe (still need to figure out what exactly is going wrong).

gforsyth: Ok, yeah, the `libnuma` feedstock recipe is just wrong.  I'll try to upstream the fixes, but for now, I would add an explicit runtime dep

bdice: Good call. Done in b239a211f53bdf45d008e4125ef7bb2480b18ea6. I'll leave this open for now. Here's the upstream PR: https://github.com/conda-forge/numactl-feedstock/pull/18

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -82,10 +83,11 @@ outputs:
...
-        - if: linux
-          then: libcufile-dev
+        - ${{ pin_compatible("libnuma", upper_bound="x", lower_bound="x") }}
```
bdice, line 91:
```suggestion
        # Needed until libnuma run-exports are fixed:
        # https://github.com/conda-forge/numactl-feedstock/pull/18
        - ${{ pin_compatible("libnuma", upper_bound="x", lower_bound="x") }}
```

bdice (approved):
The final run dependencies now look correct to me. Approving. We will want to do a bit of cleanup after https://github.com/conda-forge/numactl-feedstock/pull/18 goes in.

PR #749, opened 2025-06-06
Use C++20 standard

kingcrimsontianyu:
This PR changes KvikIO C++ standard from 17 to 20.

### DIFF
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -42,7 +42,16 @@ function(kvikio_add_benchmark)
   endif()
 
   add_executable(${_KVIKIO_NAME} ${_KVIKIO_SOURCES})
-  set_target_properties(${_KVIKIO_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
+  set_target_properties(
+    ${_KVIKIO_NAME}
+    PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib"
+               CXX_STANDARD 20
+               CXX_STANDARD_REQUIRED ON
+               # For std:: support of __int128_t. Can be removed once using cuda::std
+               CXX_EXTENSIONS ON
+               CUDA_STANDARD 20
+               CUDA_STANDARD_REQUIRED ON
+  )
 
   target_link_libraries(${_KVIKIO_NAME} PUBLIC benchmark::benchmark kvikio::kvikio)
 
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -50,11 +50,11 @@ function(kvikio_add_test)
   set_target_properties(
     ${_KVIKIO_NAME}
     PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${KvikIO_BINARY_DIR}/gtests>"
-               CXX_STANDARD 17
+               CXX_STANDARD 20
                CXX_STANDARD_REQUIRED ON
                # For std:: support of __int128_t. Can be removed once using cuda::std
                CXX_EXTENSIONS ON
-               CUDA_STANDARD 17
+               CUDA_STANDARD 20
                CUDA_STANDARD_REQUIRED ON
   )
   target_link_libraries(
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -198,7 +198,7 @@ set_target_properties(
   kvikio
   PROPERTIES BUILD_RPATH "\$ORIGIN"
              INSTALL_RPATH "\$ORIGIN"
-             CXX_STANDARD 17
+             CXX_STANDARD 20
              CXX_STANDARD_REQUIRED ON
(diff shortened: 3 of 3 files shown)


cpp/benchmarks/CMakeLists.txt
```diff
@@ -42,7 +42,16 @@ function(kvikio_add_benchmark)
   endif()
 
   add_executable(${_KVIKIO_NAME} ${_KVIKIO_SOURCES})
-  set_target_properties(${_KVIKIO_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
+  set_target_properties(
+    ${_KVIKIO_NAME}
+    PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib"
+               CXX_STANDARD 20
+               CXX_STANDARD_REQUIRED ON
+               # For std:: support of __int128_t. Can be removed once using cuda::std
```
bdice: What actions do we need to take here? Where does kvikio use `__int128_t`? I think we want to move to `cuda::std` or remove the `CXX_EXTENSIONS ON` if not needed.

bdice (changes requested):

cpp/benchmarks/CMakeLists.txt
bdice: Please also remove this from `cpp/tests/CMakeLists.txt`, it seems like a copy-paste from cudf that is unnecessary in kvikio.

kingcrimsontianyu: Thanks. Done.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -42,7 +42,15 @@ function(kvikio_add_benchmark)
   endif()
 
   add_executable(${_KVIKIO_NAME} ${_KVIKIO_SOURCES})
-  set_target_properties(${_KVIKIO_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
+  set_target_properties(
+    ${_KVIKIO_NAME}
+    PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib"
+               CXX_STANDARD 20
+               CXX_STANDARD_REQUIRED ON
+               CXX_EXTENSIONS ON
```
bdice: Please remove this. This was the line indicated by the comment about int128 support.
```suggestion
```

cpp/tests/CMakeLists.txt
```diff
@@ -50,11 +50,10 @@ function(kvikio_add_test)
   set_target_properties(
     ${_KVIKIO_NAME}
     PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${KvikIO_BINARY_DIR}/gtests>"
-               CXX_STANDARD 17
+               CXX_STANDARD 20
                CXX_STANDARD_REQUIRED ON
-               # For std:: support of __int128_t. Can be removed once using cuda::std
                CXX_EXTENSIONS ON
```
bdice:
```suggestion
```

cpp/benchmarks/CMakeLists.txt
kingcrimsontianyu: I see. Done.

bdice (approved):

conda/recipes/libkvikio/conda_build_config.yaml
```diff
@@ -1,3 +1,9 @@
+c_compiler_version:
+  - 13
+
+cxx_compiler_version:
+  - 13
```
bdice: Unfortunately we can't build with GCC 13 and `cuda-version<12.2`, which is what we need for the ARM builds without libcufile support. We'll have to rethink some of the build infrastructure...

cc: @vyasr

vyasr: I don't think this is a build infrastructure problem really. As long as we need kvikio to support building with CUDA < 12.2 we can't leverage features that require a newer CUDA version and compiler. That's a hard constraint on the code in the library, not the build infra.

Note that https://github.com/rapidsai/kvikio/issues/773 is also relevant here. The current build failures are happening because that bug means that we are failing to constrain the compiler version independently for the two variant builds. We'll need to get that fixed, but that doesn't enable us using gcc 13 here, it just fixes the existing compilation.

vyasr: Clarifying from offline: this means that we can compile with -std=c++20, but we can only use features that were available as of gcc 11 (specifically the exact latest version that the CUDA 12.1 nvcc compiler supports).

bdice: GCC 12, I thought? But otherwise yes.

conda/recipes/libkvikio/conda_build_config.yaml
```diff
@@ -1,3 +1,9 @@
...
+
```
vyasr:
```suggestion
```

conda/recipes/libkvikio/conda_build_config.yaml
vyasr: Yeah I think it is 12. Did that in #784.

PR #753, opened 2025-06-12
Only use cufile on x86

vyasr:
This is a temporary patch for the problem that cufile is not available on arm before CUDA 12.2, which results in these libkvikio packages not currently being installable on (for example) CUDA 12.0 arm systems. We should follow this PR up with a fix to actually build separate packages for older and newer CUDA versions, but this quick fix should unblock other work today.


gforsyth (approved):
:shipit:

PR #754, opened 2025-06-12
Use variants to produce separate builds with and without cufile support

vyasr:


ci/build_cpp.sh
```diff
@@ -25,10 +25,15 @@ export RAPIDS_ARTIFACTS_DIR
 # populates `RATTLER_CHANNELS` array and `RATTLER_ARGS` array
 source rapids-rattler-channel-string
 
+# Use sed to replace the RAPIDS_CUDA_VERSION string in the variants file with
+# the value of that variable in this shell.
+sed -i "s/RAPIDS_CUDA_VERSION/${RAPIDS_CUDA_VERSION}/g" "conda/recipes/libkvikio/variants_$(arch).yaml"
```
jakirkham: One thing that has worked well before, is to [output this to a variant file]( https://github.com/rapidsai/pynvjitlink/blob/68e63ae785c4b2b32497ec2e091988761d64bd6b/ci/build_conda.sh#L20-L26 )

It does a similar thing admittedly. Though it can be a bit clearer about what is happening. That could just be personal preference though

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -4,11 +4,18 @@ schema_version: 1
...
   version: ${{ env.get("RAPIDS_PACKAGE_VERSION") }}
   minor_version: ${{ (version | split("."))[:2] | join(".") }}
-  cuda_version: ${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[:2] | join(".") }}
-  cuda_major: '${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[0] }}'
+  # We need to support three cases:
+  # 1. Linux x86_64, which always uses libcufile
+  # 2. Linux aarch64 with CUDA >= 12.2, which uses libcufile
+  # 3. Linux aarch64 with CUDA < 12.2, which does not use libcufile
+  # Each case has different cuda-version constraints as expressed below
+  should_use_cufile: ${{ x86_64 or (aarch64 and cuda_version >= "12.2") }}
```
jakirkham, line 12: Agree this is a good way to do this and document it

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -4,11 +4,18 @@ schema_version: 1
...
+  # When reverting, instances of cuda_key_string can be replaced with cuda_major
+  cuda_key_string: ${{ cuda_version | replace(".", "_") }}
+  #cuda_version: ${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[:2] | join(".") }}
+  #cuda_major: '${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[0] }}'
```
jakirkham, line 16: Should we drop these?

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -86,15 +93,25 @@ outputs:
...
       run:
-        - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
-        - if: linux and x86_64
+        - if: x86_64
+          then:
+          - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+          else:
+          - if: aarch64 and cuda_version >= "12.2"
+            then:
+            - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="12.2") }}
```
jakirkham:
```suggestion
            - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="12.2.0a0") }}
```

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -121,20 +138,29 @@ outputs:
...
           - libcufile-dev
       run:
-        - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+        - if: x86_64
+          then:
+          - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+          else:
+          - if: aarch64 and cuda_version >= "12.2"
+            then:
+            - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="12.2") }}
```
jakirkham:
```suggestion
            - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="12.2.0a0") }}
```

conda/recipes/libkvikio/variants_aarch64.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - 12
+  - 13
+
+cuda_version:
+  - 12.1 # The last version to not support cufile
+  - RAPIDS_CUDA_VERSION
+
+should_use_cufile:
+  - false
+  - true
```
jakirkham: So we have this in `context` and this variant file. We should pick one

Personally would prefer `context` as there is a nice clear explanation and it is derived in a DRY way

jakirkham (commented):
Thanks Vyas! 🙏

This looks generally reasonable

Had a few comments below

ci/build_cpp.sh
bdice: I like that alternative! I agree it is clearer.

conda/recipes/libkvikio/variants_aarch64.yaml
vyasr: Thanks for catching this, I forgot some clean up here. We can use the context variable.

cpp/include/kvikio/shim/cufile.hpp
```diff
@@ -128,11 +128,7 @@ bool is_cufile_available() noexcept;
  *
  * @return The version (1000*major + 10*minor) or zero if older than 1080.
  */
-#ifdef KVIKIO_CUFILE_FOUND
 int cufile_version() noexcept;
-#else
-constexpr int cufile_version() noexcept { return 0; }
-#endif
```
vyasr, line 135: This change is necessary because the dead code optimizer will generally elide constexpr functions from the final binary if no runtime usage is detected, but we need the symbol in the DSO for usage from other libraries (including Python extensions). For the same reason we also avoid inlining the definition here.

conda/recipes/libkvikio/recipe.yaml
vyasr: I chose to keep them to simplify reverting the changes, but if others feel like it would be cleaner to drop these I am fine with that.

ci/build_cpp.sh
vyasr: I considered this possibility as well. However, in this instance it's not quite so simple as the pynvjitlink example because we need other things to be encoded in the variants as well. The behavior on x86 and arm is different because the former is a single variant while the latter requires two, so we would need to write out variants conditionally based on that. Moreover, for arm we need to use a different compiler version for each CUDA version, so it's a longer file. If you both prefer that I can switch to it, we just wind up with two fairly long here-docs instead of this command. The flip side is that we can drop the committed variants files if we choose to go that route.

vyasr: I've made the changes in https://github.com/rapidsai/kvikio/pull/754/commits/00b642a1eec0e825982a36063dc8507910fe8aeb. If you like it we can stick with it, otherwise feel free to revert that commit.

ci/build_cpp.sh
```diff
@@ -25,10 +25,42 @@ export RAPIDS_ARTIFACTS_DIR
...
+    cxx_compiler_version:
+    - 12
+    - 13
+
+    cuda_version:
+    - 12.1 # The last version to not support cufile
+    - ${RAPIDS_CUDA_VERSION}
+EOF
+fi
+
```
gforsyth, line 58: Non-blocking, but I think this would be better to encode in individual `variant_$ARCH_$CUDA_VERSION.yaml` files and then the branching logic here can be to select the appropriate variant file for the build at hand.

That way the various build variants are all in the conda recipe file and not hidden away in this bash script.

I'm fine with this being done in a follow-up to get CI unblocked.

gforsyth (approved):

ci/build_cpp.sh
gforsyth: lol, I see that was your original approach.  well, not worth going back and forth over it.  

My general take is "repeat yourself" if it keeps things clean, and heredocs in a random bash script isn't worth the tradeoff to me.

bdice (approved):
Thank you!

ci/build_cpp.sh
vyasr: Yeah I liked the old way better too. This version feels wrong in how much of the logic is moved out of the recipe itself. I think we should reconsider what approach we like best, but I'll merge as-is to get things unblocked for now.

jakirkham: Provided a simplification on the current approach in PR: https://github.com/rapidsai/kvikio/pull/755

PR #755, opened 2025-06-13
Set compiler versions in context

jakirkham:
Simplify the logic necessary to handle compiler versions by folding it into the context.

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -26,33 +26,13 @@ export RAPIDS_ARTIFACTS_DIR
 source rapids-rattler-channel-string
 
 # Construct the extra variants according to the architecture
-if [[ "$(arch)" == "x86_64" ]]; then
-    cat > variants.yaml << EOF
-    c_compiler_version:
-      - 13
-
-    cxx_compiler_version:
-      - 13
-
-    cuda_version:
-      - ${RAPIDS_CUDA_VERSION}
+cat > variants.yaml << EOF
+cuda_version:
+  - ${RAPIDS_CUDA_VERSION}
 EOF
-else
-    cat > variants.yaml << EOF
-    zip_keys:
-    - [c_compiler_version, cxx_compiler_version, cuda_version]
-
-    c_compiler_version:
-    - 12
-    - 13
-
-    cxx_compiler_version:
-    - 12
-    - 13
-
-    cuda_version:
-    - 12.1 # The last version to not support cufile
-    - ${RAPIDS_CUDA_VERSION}
+if [[ "$(arch)" == "aarch64" ]]; then
+  cat >> variants.yaml << EOF
+  - 12.1 # The last version to not support cufile
 EOF
 fi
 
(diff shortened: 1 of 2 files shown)


ci/build_cpp.sh
```diff
@@ -26,33 +26,13 @@ export RAPIDS_ARTIFACTS_DIR
...
-      - 13
-
-    cxx_compiler_version:
-      - 13
-
-    cuda_version:
-      - ${RAPIDS_CUDA_VERSION}
+cat > variants.yaml << EOF
+cuda_version:
+  - ${RAPIDS_CUDA_VERSION}
```
jakirkham, line 31: This writes out the common variant file content for all cases

ci/build_cpp.sh
```diff
@@ -26,33 +26,13 @@ export RAPIDS_ARTIFACTS_DIR
...
-    cxx_compiler_version:
-    - 12
-    - 13
-
-    cuda_version:
-    - 12.1 # The last version to not support cufile
-    - ${RAPIDS_CUDA_VERSION}
+if [[ "$(arch)" == "aarch64" ]]; then
+  cat >> variants.yaml << EOF
+  - 12.1 # The last version to not support cufile
```
jakirkham, line 35: Then the legacy CUDA ARM builds are appended to the file

This also makes it easier to drop this section of code when it no longer applies

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -10,6 +10,8 @@ context:
   # 3. Linux aarch64 with CUDA < 12.2, which does not use libcufile
   # Each case has different cuda-version constraints as expressed below
   should_use_cufile: ${{ x86_64 or (aarch64 and cuda_version >= "12.2") }}
+  c_compiler_version: ${{ 13 if should_use_cufile else 12 }}
+  cxx_compiler_version: ${{ c_compiler_version }}
```
jakirkham, line 14: Instead of setting different compiler versions in the CUDA variant file, we just set them here in the context. Also this ties them to the relevant thing. Namely whether we can build with cuFile or not

We can see this works as intended with [the CUDA 12.1 ARM build]( https://github.com/rapidsai/kvikio/actions/runs/15642022587/job/44071531096?pr=755#step:10:1838 )

```
 │ │ │ cuda-version                     ┆ 12.1         ┆ h1d6eff3_3       ┆ conda-forge ┆  20.56 KiB │
 │ │ │ gcc_impl_linux-aarch64           ┆ 12.4.0       ┆ h628656a_2       ┆ conda-forge ┆  56.19 MiB │
 │ │ │ gcc_linux-aarch64                ┆ 12.4.0       ┆ heb3b579_10      ┆ conda-forge ┆  31.88 KiB │
 │ │ │ gxx_impl_linux-aarch64           ┆ 12.4.0       ┆ h0bf7a72_2       ┆ conda-forge ┆  11.36 MiB │
 │ │ │ gxx_linux-aarch64                ┆ 12.4.0       ┆ h3f57e68_10      ┆ conda-forge ┆  30.23 KiB │
```

jakirkham (commented):
Gave a rough overview of the relevant changes below

Generally the idea is push more things into `context` where it can be written compactly. Keeping only the relevant differentiator (CUDA version) in the variant file

vyasr (approved):
This approach seems sensible to me John, thanks for the cleanup!

jakirkham:
Happy to help :)

PR #759, opened 2025-06-25
Remove nvidia and dask channels

vyasr:
Now that we have dropped support for CUDA 11 we no longer require the nvidia channel.
With the changes in https://github.com/rapidsai/rapids-dask-dependency/pull/85, RAPIDS now only uses released versions of dask, so we no longer need the dask channel either.
Contributes to https://github.com/rapidsai/build-planning/issues/184


gforsyth:
Looks like we need the updated environment yaml files checked in, too

PR #760, opened 2025-06-26
refactor(shellcheck): enable for all files and fix remaining warnings

gforsyth:
xref rapidsai/build-planning#135

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -158,6 +158,7 @@ fi
 if (( NUMARGS == 0 )) || hasArg kvikio; then
     echo "building kvikio..."
     cd "${REPODIR}"/python/kvikio
-    SKBUILD_CMAKE_ARGS="-DCMAKE_PREFIX_PATH=${INSTALL_PREFIX};-DCMAKE_LIBRARY_PATH=${LIBKVIKIO_BUILD_DIR};$(IFS=';'; echo "${EXTRA_CMAKE_ARGS[*]}")" \
+    _EXTRA_CMAKE_ARGS=$(IFS=';'; echo "${EXTRA_CMAKE_ARGS[*]}")
+    SKBUILD_CMAKE_ARGS="-DCMAKE_PREFIX_PATH=${INSTALL_PREFIX};-DCMAKE_LIBRARY_PATH=${LIBKVIKIO_BUILD_DIR};$_EXTRA_CMAKE_ARGS" \
         python -m pip install "${PYTHON_ARGS_FOR_INSTALL[@]}" .
 fi
--- a/build.sh
+++ b/build.sh
@@ -40,7 +40,7 @@ VERBOSE_FLAG=""
 BUILD_TYPE=Release
 INSTALL_TARGET=install
 RAN_CMAKE=0
-PYTHON_ARGS_FOR_INSTALL="-v --no-build-isolation --no-deps --config-settings rapidsai.disable-cuda=true"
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -146,7 +147,7 @@ fi
 if (( NUMARGS == 0 )) || hasArg libkvikio; then
     ensureCMakeRan
     echo "building libkvikio..."
-    cmake --build "${LIBKVIKIO_BUILD_DIR}" -j${PARALLEL_LEVEL} ${VERBOSE_FLAG}
+    cmake --build "${LIBKVIKIO_BUILD_DIR}" -j"${PARALLEL_LEVEL}" ${VERBOSE_FLAG}
```
jakirkham, line 150: Should we quote `VERBOSE_FLAG` too?

```suggestion
    cmake --build "${LIBKVIKIO_BUILD_DIR}" -j"${PARALLEL_LEVEL}" "${VERBOSE_FLAG}"
```

build.sh
```diff
@@ -156,7 +157,7 @@ fi
 # Build and install the kvikio Python package
 if (( NUMARGS == 0 )) || hasArg kvikio; then
     echo "building kvikio..."
-    cd ${REPODIR}/python/kvikio
-    SKBUILD_CMAKE_ARGS="-DCMAKE_PREFIX_PATH=${INSTALL_PREFIX};-DCMAKE_LIBRARY_PATH=${LIBKVIKIO_BUILD_DIR};${EXTRA_CMAKE_ARGS}" \
-        python -m pip install ${PYTHON_ARGS_FOR_INSTALL} .
+    cd "${REPODIR}"/python/kvikio
+    SKBUILD_CMAKE_ARGS="-DCMAKE_PREFIX_PATH=${INSTALL_PREFIX};-DCMAKE_LIBRARY_PATH=${LIBKVIKIO_BUILD_DIR};$(IFS=';'; echo "${EXTRA_CMAKE_ARGS[*]}")" \
```
jakirkham: Does it make sense to separate out the `EXTRA_CMAKE_ARGS` expansion onto its own line?

Also does `SKBUILD_CMAKE_ARGS` need to be `export`ed

jakirkham (commented):
Thanks Gil! 🙏

Had a couple questions below

build.sh
gforsyth: What I've been doing is going through and trying to responsibly clean up and refactor things so that they pass the `shellcheck` checks -- it wouldn't hurt to quote `VERBOSE_FLAG`, but also we don't need to, because the only two values that it can possibly have are `""` and `"-v"`, neither of which will get split by bash's string handling. (This is also why `shellcheck` doesn't flag it). I'm not opposed to quoting it, but my main goal here is so that when other contributors add or change bash scripts, we avoid some footguns -- if we have a blanket "quote all variables" policy, then I can imagine us quickly ending up in a place where humans are double-checking this stuff instead of the linter.

So all-in-all, I guess I'm -0.5 on quoting variables that don't need it.

build.sh
gforsyth: We can separate it out, it would likely make it more readable.  That's the gnarliest bash in here, I think.

`SKBUILD_CMAKE_ARGS` doesn't need to be exported so long as the call to `pip install` (and therein to `scikit-build-core`) happens within the body of this build script, since those environment variables will be in scope for the duration of the script execution.

jameslamb (approved):

PR #765, opened 2025-07-08
Fix a case sensitivity bug in S3 remote I/O

kingcrimsontianyu:
AWS S3 object key name is case sensitive. Current implementation of `open_s3_url` converts all the letters of URL to lowercase before passing it to the wrapped C++ library. As a result, if the object key name contains any capital letter, the following error will occur:
```
RuntimeError: KvikIO failure at: /home/coder/kvikio/cpp/src/shim/libcurl.cpp:176: curl_easy_perform() error (The requested URL returned error: 404)
```
This PR fixes this issue by forwarding the user-provided URL as-is.

### DIFF
--- a/python/kvikio/kvikio/remote_file.py
+++ b/python/kvikio/kvikio/remote_file.py
@@ -130,12 +130,12 @@ class RemoteFile:
             The size of the file. If None, KvikIO will ask the server
             for the file size.
         """
-        url = url.lower()
-        if url.startswith("http://") or url.startswith("https://"):
+        url_lowercase = url.lower()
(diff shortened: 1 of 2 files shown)


TomAugspurger (approved):
Looks good, thanks.

For a future refactor, we could use `urllib.parse` https://docs.python.org/3/library/urllib.parse.html to get the scheme instead of `startswith`.

```python
parsed = urllib.parse.urlparse(url.lower())

if parsed.scheme in ("http", "https"):
    ...
if parsed.scheme == "s3":
    ....
```

but given that the current implementation is already using `startswith` let's just keep it as it.

TomAugspurger (approved):

PR #766, opened 2025-07-09
Fix remote I/O file size overflow in Python binding

kingcrimsontianyu:
KvikIO remote I/O interface requires users to provide a buffer to read the remote data into. The following pattern is often used:
```python
import cupy as cp

# Create a remote handle from the URL
remote_handle = kvikio.RemoteFile.open_s3_url(url)

# Query the remote file size and preallocate the user-provided buffer
buf = cp.empty(remote_handle.nbytes(), dtype=cp.int8)

# Read into the buffer
fut = remote_handle.pread(buf)
fut.get()
```
Currently in Cython, the `extern` method `nbytes()` (remote file size) is given a return type of `int`, whereas its initial return type in the C++ library is `std::size_t`. The `int` here is interpreted as the `int` in C++ as opposed to the variable-length `int` in Python. Consequently, integer overflow occurs when reading from a large-size file, in which case `nbytes()` returns negative values.
This PR fixes this bug.

### DIFF
--- a/python/kvikio/kvikio/_lib/remote_handle.pyx
+++ b/python/kvikio/kvikio/_lib/remote_handle.pyx
@@ -36,7 +36,7 @@ cdef extern from "<kvikio/remote_handle.hpp>" nogil:
             unique_ptr[cpp_RemoteEndpoint] endpoint, size_t nbytes
         ) except +
         cpp_RemoteHandle(unique_ptr[cpp_RemoteEndpoint] endpoint) except +
-        int nbytes() except +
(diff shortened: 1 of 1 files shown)


TomAugspurger (approved):
Thanks!

I assume a test here is challenging, since it would require writing how a large amount of data. I'm not able to easily see a way to test this so I think it's OK to merge (we could maybe do something with rewriting the HTTP header giving the response size in the test suite, but IMO that's more effort than it's worth right now).

PR #769, opened 2025-07-16
Remove CUDA 11 mentions from docs

pentschev:
As part of #768, remove CUDA 11 from docs.

### DIFF
--- a/cpp/doxygen/main_page.md
+++ b/cpp/doxygen/main_page.md
@@ -35,8 +35,6 @@ Install the **stable release** from the ``rapidsai`` channel with the following:
 mamba install -c rapidsai -c conda-forge libkvikio
 # Create new environment (CUDA 12)
 mamba create -n libkvikio-env -c rapidsai -c conda-forge cuda-version=12.9 libkvikio
-# Create new environment (CUDA 11)
-mamba create -n libkvikio-env -c rapidsai -c conda-forge cuda-version=11.8 libkvikio
 ```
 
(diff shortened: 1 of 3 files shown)


bdice (approved):

TomAugspurger:
Ah I probably broke that [here](https://github.com/zarr-developers/zarr-python/pull/3191/files#diff-a801b931f5dd8cd7df666e3f9936ff520669296303d3a3e469514443211885bdR43).

I'll get it fixed, probably upstream in zarr-python.

TomAugspurger:
zarr-python appears to be doing everything correctly in https://github.com/zarr-developers/zarr-python/blob/main/pyproject.toml#L423-L426, at least for itself. Maybe we need to duplicate that mark here, since we're running zarr-python's tests.

I'll try reproducing quick and will then will push a fix here if that's what's needed.

TomAugspurger:
https://github.com/rapidsai/kvikio/pull/769/commits/bdca513912dd6aa3053d6450c660ffab679250aa fixed the zarr tests for me locally.

pentschev:
Thanks @TomAugspurger for fixing that and @bdice for approval!

jakirkham:
Thanks all! 🙏

Think we can also simplify this code path given we no longer need to handle all of these older CUDA 11 versions

https://github.com/rapidsai/kvikio/blob/83a25ec05d75b11e0f04a37cf25639b425d61cb0/cpp/src/shim/cufile.cpp#L30-L40

jakirkham (approved):

PR #770, opened 2025-07-16
Remove CUDA 11 workarounds

pentschev:
As part of https://github.com/rapidsai/kvikio/issues/768, remove CUDA 11 workarounds that should not anymore be necessary given CUDA 11 support is being dropped.

### DIFF
--- a/cpp/src/shim/cufile.cpp
+++ b/cpp/src/shim/cufile.cpp
@@ -27,17 +27,7 @@ namespace kvikio {
 #ifdef KVIKIO_CUFILE_FOUND
 cuFileAPI::cuFileAPI()
 {
-  // CUDA versions before CUDA 11.7.1 did not ship libcufile.so.0, so this is
-  // a workaround that adds support for all prior versions of libcufile.
-  void* lib = load_library({"libcufile.so.0",
-                            "libcufile.so.1.3.0" /* 11.7.0 */,
-                            "libcufile.so.1.2.1" /* 11.6.2, 11.6.1 */,
-                            "libcufile.so.1.2.0" /* 11.6.0 */,
-                            "libcufile.so.1.1.1" /* 11.5.1 */,
-                            "libcufile.so.1.1.0" /* 11.5.0 */,
-                            "libcufile.so.1.0.2" /* 11.4.4, 11.4.3, 11.4.2 */,
-                            "libcufile.so.1.0.1" /* 11.4.1 */,
-                            "libcufile.so.1.0.0" /* 11.4.0 */});
+  void* lib = load_library({"libcufile.so.0"});
   get_symbol(HandleRegister, lib, KVIKIO_STRINGIFY(cuFileHandleRegister));
   get_symbol(HandleDeregister, lib, KVIKIO_STRINGIFY(cuFileHandleDeregister));
   get_symbol(Read, lib, KVIKIO_STRINGIFY(cuFileRead));
@@ -79,21 +69,6 @@ cuFileAPI::cuFileAPI()
     get_symbol(StreamRegister, lib, KVIKIO_STRINGIFY(cuFileStreamRegister));
     get_symbol(StreamDeregister, lib, KVIKIO_STRINGIFY(cuFileStreamDeregister));
   }
-
-  // cuFile is supposed to open and close the driver automatically but
-  // because of a bug in cuFile v1.4 (CUDA v11.8) it sometimes segfaults:
-  // <https://github.com/rapidsai/kvikio/issues/159>.
-  if (version < 1050) { driver_open(); }
-}
-
-// Notice, we have to close the driver at program exit (if we opened it) even though we are
-// not allowed to call CUDA after main[1]. This is because, cuFile will segfault if the
-// driver isn't closed on program exit i.e. we are doomed if we do, doomed if we don't, but
-// this seems to be the lesser of two evils.
-// [1] <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization>
-cuFileAPI::~cuFileAPI()
-{
-  if (version < 1050) { driver_close(); }
 }
 #else
 cuFileAPI::cuFileAPI() { KVIKIO_FAIL("KvikIO not compiled with cuFile.h", std::runtime_error); }


cpp/src/shim/cufile.cpp
```diff
@@ -79,21 +69,6 @@ cuFileAPI::cuFileAPI()
...
-  // <https://github.com/rapidsai/kvikio/issues/159>.
-  if (version < 1050) { driver_open(); }
-}
-
-// Notice, we have to close the driver at program exit (if we opened it) even though we are
-// not allowed to call CUDA after main[1]. This is because, cuFile will segfault if the
-// driver isn't closed on program exit i.e. we are doomed if we do, doomed if we don't, but
-// this seems to be the lesser of two evils.
-// [1] <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization>
-cuFileAPI::~cuFileAPI()
```
jakirkham, line 94: Seeing the following error on all CI jobs (like [this job]( https://github.com/rapidsai/kvikio/actions/runs/16328740410/job/46125867694?pr=770#step:10:308 )):

```cpp
./BASIC_IO_EXAMPLE: symbol lookup error: /opt/conda/envs/test/bin/tests/libkvikio/../../../lib/libkvikio.so: undefined symbol: _ZN6kvikio9cuFileAPID1Ev
```

Wondering if this is related to the destructor. Do we still want an empty constructor or do want to drop this code as well?

https://github.com/rapidsai/kvikio/blob/54c420a652bf29e3c7de9ac2e2e19af07de7c256/cpp/include/kvikio/shim/cufile.hpp#L66-L73

cpp/src/shim/cufile.cpp
```diff
@@ -27,17 +27,7 @@ namespace kvikio {
...
-  void* lib = load_library({"libcufile.so.0",
-                            "libcufile.so.1.3.0" /* 11.7.0 */,
-                            "libcufile.so.1.2.1" /* 11.6.2, 11.6.1 */,
-                            "libcufile.so.1.2.0" /* 11.6.0 */,
-                            "libcufile.so.1.1.1" /* 11.5.1 */,
-                            "libcufile.so.1.1.0" /* 11.5.0 */,
-                            "libcufile.so.1.0.2" /* 11.4.4, 11.4.3, 11.4.2 */,
-                            "libcufile.so.1.0.1" /* 11.4.1 */,
-                            "libcufile.so.1.0.0" /* 11.4.0 */});
+  void* lib = load_library("libcufile.so.0");
```
jakirkham, line 30: Since we are using this with only one library, think we could drop this as well

https://github.com/rapidsai/kvikio/blob/54c420a652bf29e3c7de9ac2e2e19af07de7c256/cpp/src/shim/utils.cpp#L36-L48

Though perhaps this functionality is useful in the future

jakirkham (commented):
Thanks Peter! 🙏

This looks good

Had a couple follow-up questions below

cpp/src/shim/cufile.cpp
jakirkham: Have pushed this change in commit: https://github.com/rapidsai/kvikio/pull/770/commits/53036833e74338fe9f888dc5ed6223fd2919e1bd

Let's see if that helps CI

cpp/src/shim/cufile.cpp
jakirkham: Handled in commit: https://github.com/rapidsai/kvikio/pull/770/commits/999042a487671dd6bc771f6b049703917026ac13

cpp/src/shim/cufile.cpp
```diff
@@ -79,21 +69,6 @@ cuFileAPI::cuFileAPI()
     get_symbol(StreamRegister, lib, KVIKIO_STRINGIFY(cuFileStreamRegister));
     get_symbol(StreamDeregister, lib, KVIKIO_STRINGIFY(cuFileStreamDeregister));
   }
-
-  // cuFile is supposed to open and close the driver automatically but
-  // because of a bug in cuFile v1.4 (CUDA v11.8) it sometimes segfaults:
-  // <https://github.com/rapidsai/kvikio/issues/159>.
-  if (version < 1050) { driver_open(); }
```
jakirkham, line 86: @vuule could you please double check whether this is ok for you?

Asking as you had originally opened issue ( https://github.com/rapidsai/kvikio/issues/159 ). However based on the comment above, this should not be an issue in CUDA 12.0+

cpp/src/shim/cufile.cpp
```diff
@@ -79,21 +69,6 @@ cuFileAPI::cuFileAPI()
...
-}
-
-// Notice, we have to close the driver at program exit (if we opened it) even though we are
-// not allowed to call CUDA after main[1]. This is because, cuFile will segfault if the
-// driver isn't closed on program exit i.e. we are doomed if we do, doomed if we don't, but
-// this seems to be the lesser of two evils.
-// [1] <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization>
-cuFileAPI::~cuFileAPI()
-{
-  if (version < 1050) { driver_close(); }
```
jakirkham, line 96: @vuule this is also part of the same question above

cpp/src/shim/cufile.cpp
jakirkham: Looks like that fixed CI 🎉

cpp/src/shim/cufile.cpp
jakirkham: Note: That KvikIO with newer cuFile versions handles driver opening/closing in this code

https://github.com/rapidsai/kvikio/blob/33a734cde6f4c5af5960fed826131e8797b5b46a/cpp/src/cufile/driver.cpp#L45-L56

https://github.com/rapidsai/kvikio/blob/33a734cde6f4c5af5960fed826131e8797b5b46a/cpp/src/shim/cufile.cpp#L108-L122

kingcrimsontianyu (approved):
I agree with all the changes.

PS: I haven't found a way to access archived cuFile API documentation, so I'm not sure whether the following version guards could be removed in addition. Let's keep them for this PR.
 https://github.com/rapidsai/kvikio/blob/999042a487671dd6bc771f6b049703917026ac13/cpp/src/shim/cufile.cpp#L59-L71

Issue #787, opened 2025-07-29
[FEA] Support for WebHDFS file read


kingcrimsontianyu:
There is a need for KvikIO to support HDFS (Hadoop Distributed File System), a distributed file system with the features of fault tolerance and high throughput. Files stored on HDFS cannot be accessed by standard system calls such as `read` and `write`. Instead, the following specialized methods are necessary:

- Command line utility: `hdfs dfs`
- HDFS Java API: This is the canonical way to access HDFS
- `libhdfs` in C, or `libhdfs3` in C++: Support a subset of features of the Java API
- WebHDFS REST API: HTTP-based access for non-Java applications
- HttpFS: Another HTTP-based alternative

We plan to start with WebHDFS due to its simplicity (no additional dependency required). Note that existing `RemoteHandle` and `HttpEndpoint` in KvikIO cannot be directly used for WebHDFS. Further extension is necessary:
- In WebHDFS, URL must contain query parameters. Operations such as "determining file size" and "reading from the file" use different query parameters. KvikIO shall be able to build the corresponding URL and parse the json response properly.
- In WebHDFS, reading from a range (offset, size) should also be specified via query parameters.
- In WebHDFS, file access request is always redirected from a _name node_ to a _data node_ (HTTP 307 redirect). KvikIO should take this into account by using either manual or automatic redirection in libcurl.

KvikIO should provide C++ implementation and Python binding for WebHDFS file read.

PR #788, opened 2025-07-30
Support WebHDFS (1/2): C++ implementation

kingcrimsontianyu:
## Summary
This PR adds WebHDFS support to KvikIO. The background information is available at #787.

## Limitations
This PR does not address:
- Idiomatic and secure URL parsing and validation
- Testing on URL encoding/decoding (which means percent-decoded URL may or may not work at the moment)
- Advanced authentication such as Kerberos

These features will be added in the future.

Partially addresses #787


cpp/src/remote_handle.cpp
```diff
@@ -174,6 +189,12 @@ std::size_t HttpEndpoint::get_file_size()
   return get_file_size_using_head_impl(*this, _url);
 }
 
+void HttpEndpoint::setup_range_request(CurlHandle& curl, std::size_t file_offset, std::size_t size)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
```
madsbk: Maybe a bit overkill? :)
```suggestion
```

madsbk: The same goes for the other `setopt`, I would expect their running times are similar to `KVIKIO_NVTX_FUNC_RANGE()` itself?

cpp/tests/CMakeLists.txt
```diff
@@ -76,6 +76,10 @@ kvikio_add_test(NAME ERROR_TEST SOURCES test_error.cpp)
 
 kvikio_add_test(NAME MMAP_TEST SOURCES test_mmap.cpp)
 
-kvikio_add_test(NAME REMOTE_HANDLE_TEST SOURCES test_remote_handle.cpp utils/env.cpp)
+if(KvikIO_REMOTE_SUPPORT)
+  kvikio_add_test(NAME REMOTE_HANDLE_TEST SOURCES test_remote_handle.cpp utils/env.cpp)
+
```
madsbk:
```suggestion
```

madsbk (approved):
Looks great, I haven't tested the hdfs protocol but the libcurl usage looks sound.

cpp/src/remote_handle.cpp
kingcrimsontianyu: Agree. Removed.

kingcrimsontianyu:
An update to the unit test in this PR: 
Since gtest does not guarantee that different `TEST_F`s from the same test suite are always run sequentially (e.g. by using `--gtest_parallel`), it is necessary to create/upload/destroy the remote test file `"/tmp/kvikio-test-webhdfs.bin"` only once per test suite, instead of per test, to avoid potential race condition. So `static void SetUpTestSuite()` and `static void TearDownTestSuite()` are used to replace `SetUp()` and `TearDown()`.

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -152,6 +153,7 @@ outputs:
             else:
             - ${{ pin_compatible("cuda-version", upper_bound="12.2.0a0", lower_bound="12.0") }}
         - cuda-cudart
+        - libcurl ${{ libcurl_version }}
```
kingcrimsontianyu: Fixing the "overlinking" error from conda according to Lawrence's (@wence- ) suggestion. One remaining question raised is why `libcurl` is not present in `libkvikio`'s `run` section. :thinking:

kingcrimsontianyu: PS: The generated `build.ninja` file indicates the use of static library `libcurl-d.a` (under debug build). This may be related to `libcurl` not being listed under `run` but `ignore_run_exports` section.
```
build gtests/HDFS_TEST: CXX_EXECUTABLE_LINKER__HDFS_TEST_Debug tests/CMakeFiles/HDFS_TEST.dir/test_hdfs.cpp.o tests/CMakeFiles/HDFS_TEST.dir/utils/hdfs_helper.cpp.o | libkvikio.so lib/libgmock.a lib/libgmock_main.a lib/libgtest.a lib/libgtest_main.a /usr/local/cuda-12.9/targets/x86_64-linux/lib/libcudart.so _deps/curl-build/lib/libcurl-d.a lib/libgmock.a lib/libgtest.a /usr/lib/x86_64-linux-gnu/librt.a /usr/lib/x86_64-linux-gnu/libssl.so /usr/lib/x86_64-linux-gnu/libcrypto.so /usr/lib/x86_64-linux-gnu/libz.so || _deps/curl-build/lib/libcurl-d.a lib/libgmock.a lib/libgmock_main.a lib/libgtest.a lib/libgtest_main.a libkvikio.so
```

gforsyth: Dependencies that are required at build/compile time should be listed under `host`.  Dependencies required at _runtime_ should be listed under `run`.  Plenty of dependencies require both.
Do we need `libcurl` at runtime, or just for the build?

gforsyth: The `ignore_run_exports` is a bit different.  A package (like `libcurl`) can define its own runtime dependencies that you may or may not want to accept.  In the case of `libkvikio`, we need `libcurl` at build time but we don't want or need the explicit runtime dependency on `libcurl`, so we add it to the `ignore_run_exports`.

You _probably_ want to do the same thing for the `libkvikio-tests` output

wence-: > Do we need libcurl at runtime, or just for the build?

We need it a build time, and naively at runtime because we depend on libcurl. _But_ we end up not needing it at runtime because we statically link libcurl into `libkvikio.so`.

If we were to change to dynamically linking, I presume we would also need it at runtime.

The above applies mutatis mutandis to the `libkvikio-tests` package that in this PR now also directly depends on `libcurl` at build time.

bdice: Yes, I agree -- ideally we shouldn't need `libcurl` in the runtime dependencies of `libkvikio` or `libkvikio-tests` because it should be statically linked.

KyleFromNVIDIA: Even if it becomes dynamically linked, wouldn't `libcurl` be a run export, and thus not need to be explicitly added?

gforsyth: If it is statically linked then we'd want to explicitly disable the run-export so we don't enforce an (admittedly light) unneeded dependency downstream.

Although for the tests package, I'm much less concerned about carrying around a runtime dep of `libcurl`

bdice: I investigated the C++ artifacts from 0507d78 (https://github.com/rapidsai/kvikio/actions/runs/16811025193/artifacts/3712834228).

The `HDFS_TEST` binary does link to `libcurl.so` as shown in [the CI output](https://github.com/rapidsai/kvikio/actions/runs/16811025193/job/47616245904?pr=788#step:10:1661). However, I think we're passing the overlinking check because `libcurl` is in the `host` environment. This is probably not a good solution and leads to finding `libcurl.so.4` from my system rather than conda.

In fact, the same thing is happening for `libkvikio.so` (it links to `libcurl.so.4` but doesn't have a run dependency on `libcurl`).

On further inspection, I think we're only statically linking `libcurl` in wheels but not conda packages where we still expect the shared library to be present. We haven't run into issues because `libcurl.so.4` is pretty widely available in conda environments and Linux OSes. We probably need to keep the `ignore_run_exports` here because [the run-export pinnings are really tight (exact)](https://github.com/conda-forge/curl-feedstock/blob/6dc1ed3d5de0175deaa1445098f8ab8fd0b77875/recipe/meta.yaml#L52) but we should add an explicit runtime pinning to avoid this problem in both `libkvikio` and `libkvikio-tests`. I'll push a commit here to do that.

Additional references:
- https://curl.se/libcurl/abi.html libcurl ABI compatibility promises are pretty good

bdice: Done in cb24034.

bdice: I checked myself again on this -- I couldn't believe that `libcurl`'s run-exports were exact. I was wrong and they are not. It says `pin_subpackage('libcurl')`, not `pin_subpackage('libcurl', exact=True)`. I just misread it earlier. We often use `exact=True` so I overlooked it at first.

The run-exports of the package we use at build time (libcurl 8.5.0) are `{"weak": ["libcurl >=8.5.0,<9.0a0"]}`. This is totally fine to use, we don't need to ignore `libcurl`'s run-exports and add our own pinnings. I pushed another commit fixing this (and simplifying what I did in the previous commit): 70be8f99c2cda060d8a9d807fc187cf54fab0909

kingcrimsontianyu: Thanks @bdice!

bdice (approved):
libcurl packaging now works as intended for libkvikio and libkvikio-tests. Approving!

PR #789, opened 2025-07-31
Support file size query for S3 presigned URL

### LINKED ISSUE #585: 403 with pre-signed S3 URL
S3 supports pre-signed URLs, a way to encode authorization into the URL so that they can be shared and used similarly to a public HTTP URL. Currently, it looks like they are not supported. A pre-signed URL can be generated through the console, CLI, or SDKs:

```python
In [1]: import boto3

In [2]: import boto3, httpx

In [3]: s3 = boto3.client("s3")

In [4]: url = s3.generate_presigned_url("get_object", Params={"Bucket": "kvikiobench-56481", "Key": "data/small/0000"}, ExpiresIn=600)

In [5]: httpx.get(url).status_code
Out[5]: 200
```

If we take that url and use it with kvikio, we get a 403 error:

```pytb
>>> import kvikio
>>> kvikio.RemoteFile.open_http(url="https://kvikiobench-56481.s3.us-east-2.amazonaws.com/data/small/0000?response-content-disposition=inline&X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Security-Token=...")
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[3], line 1
----> 1 kvikio.RemoteFile.open_http(url="https://kvikiobench-56481.s3.us-east-2.amazonaws.com/data/small/0000?response-content-disposition=inline&X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Security-Token=...")

File /raid/toaugspurger/envs/kvikio-env/lib/python3.12/site-packages/kvikio/remote_file.py:69, in RemoteFile.open_http(cls, url, nbytes)
     53 @classmethod
     54 def open_http(
     55     cls,
     56     url: str,
     57     nbytes: Optional[int] = None,
     58 ) -> RemoteFile:
     59     """Open a http file.
     60
     61     Parameters
   (...)
     67         for the file size.
     68     """
---> 69     return RemoteFile(_get_remote_module().RemoteFile.open_http(url, nbytes))

File remote_handle.pyx:92, in kvikio._lib.remote_handle.RemoteFile.open_http()

File remote_handle.pyx:81, in kvikio._lib.remote_handle.RemoteFile._from_endpoint()

RuntimeError: curl_easy_perform() error near /opt/conda/conda-bld/work/cpp/src/remote_handle.cpp:47(The requested URL returned error: 403)
```

kingcrimsontianyu:
## Background

Knowing the size of the remote file before reading is important in remote I/O, as it allows users to pre-allocate buffer to avoid expensive on-the-fly reallocation. Currently in KvikIO this is not possible for AWS S3 presigned URL, which is a special link generated by data owner to grant time-limited access without using AWS credentials.

As is described in #585, file size query in KvikIO results in the HTTP 403 (forbidden) status code. This is because the query method is based on the `HEAD` request, and AWS S3 does not allow `HEAD` for presigned URL.

## Proposed solution

This PR provides a solution. The idea is to send a `GET` request (instead of `HEAD`) with a 1-byte range, so that we can still obtain the header information at a negligible cost. Since the `content-length` header is now at a fixed value of 1, we instead extract the file size value from `content-range`.

This PR adds a new C++ endpoint `S3EndpointWithPresignedUrl` and Python API `kvikio.RemoteFile.open_s3_presigned_url(url)`.

## Result

The following code now works properly without 403 error:

```python
import kvikio
import cupy

presigned_url = "<long_url_generated_by_data_owner>"
remote_file = kvikio.RemoteFile.open_s3_presigned_url(presigned_url)
print("--> file size: {:}".format(remote_file.nbytes()))

buf = cupy.zeros(remote_file.nbytes() // 8)
fut = remote_file.pread(buf)
read_size = fut.get()

print("--> read_size: {:}", read_size)
print(buf)
```

## Limitation

This PR is tested manually using a presigned URL. In a future PR, we need to add unit tests using `boto`.

Closes #585

### DIFF
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -63,7 +63,7 @@ class RemoteEndpoint {
   virtual std::string str() const = 0;
 
   /**
-   * @brief Get the file size.
+   * @brief Get the size of the remote file.
    *
    * @return The file size
    */
@@ -84,15 +84,10 @@ class HttpEndpoint : public RemoteEndpoint {
    * @param url The full http url to the remote file.
    */
   HttpEndpoint(std::string url);
+
   ~HttpEndpoint() override = default;
   void setopt(CurlHandle& curl) override;
   std::string str() const override;
-
-  /**
-   * @brief Get the file size.
-   *
-   * @return The file size
-   */
   std::size_t get_file_size() override;
 };
 
@@ -200,18 +195,15 @@ class S3Endpoint : public RemoteEndpoint {
              std::optional<std::string> aws_session_token     = std::nullopt);
 
   ~S3Endpoint() override;
-
   void setopt(CurlHandle& curl) override;
   std::string str() const override;
-
-  /**
-   * @brief Get the file size.
-   *
-   * @return The file size
-   */
   std::size_t get_file_size() override;
 };
 
+/**
+ * @brief A remote endpoint using AWS's S3 protocol and expecting a presigned URL. File access via
+ * this type of URL is time-limited and does not require AWS credentials.
+ */
 class S3EndpointWithPresignedUrl : public RemoteEndpoint {
  private:
   std::string _url;
@@ -220,15 +212,8 @@ class S3EndpointWithPresignedUrl : public RemoteEndpoint {
   explicit S3EndpointWithPresignedUrl(std::string presigned_url);
 
   ~S3EndpointWithPresignedUrl() override = default;
-
   void setopt(CurlHandle& curl) override;
   std::string str() const override;
-
-  /**
-   * @brief Get the file size.
-   *
-   * @return The file size
-   */
   std::size_t get_file_size() override;
 };
 
@@ -267,7 +252,8 @@ class RemoteHandle {
   /**
    * @brief Get the file size.
    *
-   * Note, this is very fast, no communication needed.
+   * Note, this is usually very fast, and involves retrieving only the header data, or at most only
+   * a small amount of body data.
    *
    * @return The number of bytes.
    */
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -84,16 +84,16 @@ class HttpEndpoint : public RemoteEndpoint {
    * @param url The full http url to the remote file.
    */
   HttpEndpoint(std::string url);
-  virtual ~HttpEndpoint() override = default;
-  virtual void setopt(CurlHandle& curl) override;
-  virtual std::string str() const override;
+  ~HttpEndpoint() override = default;
+  void setopt(CurlHandle& curl) override;
+  std::string str() const override;
 
   /**
    * @brief Get the file size.
    *
    * @return The file size
    */
-  virtual std::size_t get_file_size() override;
+  std::size_t get_file_size() override;
 };
 
 /**
@@ -199,17 +199,17 @@ class S3Endpoint : public RemoteEndpoint {
              std::optional<std::string> aws_endpoint_url      = std::nullopt,
              std::optional<std::string> aws_session_token     = std::nullopt);
 
-  virtual ~S3Endpoint() override;
+  ~S3Endpoint() override;
 
-  virtual void setopt(CurlHandle& curl) override;
-  virtual std::string str() const override;
+  void setopt(CurlHandle& curl) override;
+  std::string str() const override;
 
   /**
    * @brief Get the file size.
    *
    * @return The file size
    */
-  virtual std::size_t get_file_size() override;
+  std::size_t get_file_size() override;
 };
 
 class S3EndpointWithPresignedUrl : public RemoteEndpoint {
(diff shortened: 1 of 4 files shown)


cpp/include/kvikio/remote_handle.hpp
```diff
@@ -229,7 +252,8 @@ class RemoteHandle {
   /**
    * @brief Get the file size.
    *
-   * Note, this is very fast, no communication needed.
+   * Note, this is usually very fast, and involves retrieving only the header data, or at most only
+   * a small amount of body data.
```
```suggestion
    * Note, the file size is retrieved at construction so this method is very fast, no communication needed.
```
I guess calling `.nbytes()` still doesn't require any communication. Maybe move the comment up to the ctor?

kingcrimsontianyu: You are right. I was thinking of different versions of `get_file_size()` instead of `nbytes()`. Updated the comment as suggested.

cpp/src/remote_handle.cpp
```diff
@@ -133,12 +134,46 @@ class BounceBufferH2D {
...
+ * permitted.
+ *
+ * This function works for the `HttpEndpoint` and `S3Endpoint`, but not for
+ * `S3EndpointWithPresignedUrl`, which does not allow `HEAD` request.
+ *
+ * @param endpoint The remote endpoint
+ * @param url The URL of the remote file
+ * @return The file size
+ */
+std::size_t get_file_size_using_head_impl(RemoteEndpoint& endpoint, std::string const& url)
```
wence-, line 148:
```suggestion
static std::size_t get_file_size_using_head_impl(RemoteEndpoint& endpoint, std::string const& url)
```
?

cpp/src/remote_handle.cpp
```diff
@@ -286,28 +321,98 @@ S3Endpoint::~S3Endpoint() { curl_slist_free_all(_curl_header_list); }
...
+{
+}
+
+void S3EndpointWithPresignedUrl::setopt(CurlHandle& curl)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  curl.setopt(CURLOPT_URL, _url.c_str());
+}
+
+std::string S3EndpointWithPresignedUrl::str() const { return _url; }
```
wence-, line 341: *nit*: Should these `str` methods be `const noexcept`?

cpp/src/remote_handle.cpp
```diff
@@ -286,28 +321,98 @@ S3Endpoint::~S3Endpoint() { curl_slist_free_all(_curl_header_list); }
...
+ * @brief Callback for the `CURLOPT_HEADERFUNCTION` parameter in libcurl
+ *
+ * The header callback is called once for each header and only complete header lines are passed on
+ * to the callback. The provided header line is not null-terminated.
+ *
+ * @param data Transfer buffer where new data is received
+ * @param size Curl internal implementation always sets this parameter to 1
+ * @param num_bytes The size of new data received
+ * @param userdata User-defined data
+ * @return The number of bytes consumed by the callback
```
wence-, line 354: Mention `@throws` ?

kingcrimsontianyu: Thanks for the suggestion. Done.
This impelled me to give another look at the exception handling here. Apparently, the exception message from `std::stol` is too concise to be useful at all --- just a string `"stol"`! So I've modified this part and let KvikIO throw with custom message along with source code location. Now in case the server does not know the file size, the following will be shown:
```
  File "/home/coder/kvikio/tmp/test.py", line 5, in <module>
    remote_file = kvikio.RemoteFile.open_s3_presigned_url(presigned_url)
  File "/home/coder/kvikio/python/kvikio/kvikio/remote_file.py", line 162, in open_s3_presigned_url
    _get_remote_module().RemoteFile.open_s3_from_http_presigned_url(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        presigned_url, nbytes
        ^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "kvikio/_lib/remote_handle.pyx", line 153, in kvikio._lib.remote_handle.RemoteFile.open_s3_from_http_presigned_url
  File "kvikio/_lib/remote_handle.pyx", line 90, in kvikio._lib.remote_handle.RemoteFile._from_endpoint
ValueError: KvikIO failure at: /home/coder/kvikio/cpp/src/remote_handle.cpp:382: File size information missing on the server side.
```

cpp/src/remote_handle.cpp
kingcrimsontianyu: Actually, this function is put inside an unnamed namespace in `remote_handle.cpp`. So we are fine; no potential ODR issue would occur.

cpp/src/remote_handle.cpp
kingcrimsontianyu: Since the return value is `std::string` instead of `std::string&`, I think we don't need `noexcept` because strictly speaking constructing the `std::string` object at the point of `return` is potentially throwing. 

Taking a closer look at this method. According to its comment in the abstract class, it is intended to "get a description of this remote point instance", and we've been using it to refer to URL in different places. My 2 cents are in a future PR, we could have a separate `url()` method to serve this purpose (which should be `const noexcept` naturally), and make `str()` contain only textual description. cc @madsbk

wence-: Thanks

Issue #634, opened 2025-02-13
Mock remote I/O


kingcrimsontianyu:
We should create C++ mock tests for the `CurlHandle` class to be able to check the code correctness on CI, without the need to perform actual remote file access.

Related to #556 .

PR #791, opened 2025-08-05
Support WebHDFS (2/2): Python binding

kingcrimsontianyu:
## Summary
This PR adds Python binding for the WebHDFS support

Depends on PR https://github.com/rapidsai/kvikio/pull/788

Closes https://github.com/rapidsai/kvikio/issues/787

Python's built-in package `http.server` is well suited to server mocking. It enables high-level testing for the client. Closes https://github.com/rapidsai/kvikio/issues/634 too.

### DIFF
--- a/python/kvikio/tests/test_hdfs_io.py
+++ b/python/kvikio/tests/test_hdfs_io.py
@@ -9,6 +9,7 @@ from http.server import BaseHTTPRequestHandler, HTTPServer
 from multiprocessing import Process, Queue
 from typing import Any, Generator
 
+import cupy as cp
 import numpy as np
 import numpy.typing as npt
 import pytest
@@ -124,18 +125,22 @@ class TestWebHdfsOperations:
         assert file_size == remote_file_data.file_size
 
     def test_parallel_read(
-        self, mock_webhdfs_server: str, remote_file_data: RemoteFileData
+        self, mock_webhdfs_server: str, remote_file_data: RemoteFileData, xp: Any
     ) -> None:
         url = f"{mock_webhdfs_server}{remote_file_data.file_path}"
         handle = remote_file.RemoteFile.open_webhdfs(url)
-        result_buf = np.arange(
+        result_buf = xp.arange(
             0, remote_file_data.num_elements, dtype=remote_file_data.dtype
         )
         fut = handle.pread(result_buf)
         read_size = fut.get()
 
         assert read_size == remote_file_data.file_size
-        assert np.array_equal(result_buf, remote_file_data.buf)
+
+        result_buf_np = result_buf
+        if isinstance(result_buf, cp.ndarray):
+            result_buf_np = cp.asnumpy(result_buf)
+        assert np.array_equal(result_buf_np, remote_file_data.buf)
 
     @pytest.mark.parametrize("size", [80, 8 * 9999])
     @pytest.mark.parametrize("offset", [0, 800, 8000, 8 * 9999])
@@ -149,6 +154,7 @@ class TestWebHdfsOperations:
         offset: int,
         num_threads: int,
         task_size: int,
+        xp: Any,
     ) -> None:
         url = f"{mock_webhdfs_server}{remote_file_data.file_path}"
         element_size = remote_file_data.buf.itemsize
@@ -159,12 +165,16 @@ class TestWebHdfsOperations:
         actual_num_elements = size // np.dtype(remote_file_data.dtype).itemsize
         with kvikio.defaults.set({"num_threads": num_threads, "task_size": task_size}):
             handle = remote_file.RemoteFile.open_webhdfs(url)
-            result_buf = np.zeros(actual_num_elements, dtype=remote_file_data.dtype)
+            result_buf = xp.zeros(actual_num_elements, dtype=remote_file_data.dtype)
             fut = handle.pread(result_buf, size, offset)
             read_size = fut.get()
 
             assert read_size == size
-            assert np.array_equal(result_buf, expected_buf)
+
+            result_buf_np = result_buf
+            if isinstance(result_buf, cp.ndarray):
+                result_buf_np = cp.asnumpy(result_buf)
+            assert np.array_equal(result_buf_np, expected_buf)
 
 
 class TestWebHdfsErrors:


kingcrimsontianyu:
## Verification

```python
import cupy
import numpy as np
import kvikio
import kvikio.defaults

kvikio.defaults.set("num_threads", 8)

url = "<path-to-pds-scale-100.0-partsupp.parquet>"
remote_file = kvikio.RemoteFile.open_webhdfs(url)
print("--> file size: {:}".format(remote_file.nbytes()))

buf = cupy.zeros(remote_file.nbytes(), dtype=np.uint8)
fut = remote_file.pread(buf)
read_size = fut.get()

print("--> read_size: {:}".format(read_size))
assert read_size == remote_file.nbytes()
```

image

madsbk (commented):
Looks good, but I think we need a basic pytest

madsbk:
What about leveraging Python to spin up a tiny WebHDFS-compatible server with pytest-httpserver? I’m not sure how much work it is, but it might be worth having a test that doesn’t rely on a remote HDFS server.

Perplexity suggestion something like:
```python
import pytest
from pytest_httpserver import HTTPServer
from kvikio import remote_file


@pytest.fixture
def webhdfs_server(httpserver: HTTPServer) -> HTTPServer:
    """
    Minimal WebHDFS OPEN mock:
      GET /webhdfs/v1/test.txt?op=OPEN&user.name=testuser  -> 307 with Location
      GET <Location> (same server, with &datanode=true)    -> 200 with body
    Optionally allow a HEAD on the first URL.
    """
    # Optional HEAD some clients issue
    httpserver.expect_request(
        "/webhdfs/v1/test.txt",
        method="HEAD",
        query_string={"op": "OPEN", "user.name": "testuser"},
    ).respond_with_data(b"", status=200)

    # Coordinator GET: send 307 redirect with Location
    redirect_target_path = "/webhdfs/v1/test.txt"
    redirect_qs = "op=OPEN&user.name=testuser&datanode=true"
    redirect_url = httpserver.url_for(redirect_target_path) + "?" + redirect_qs

    httpserver.expect_request(
        redirect_target_path,
        method="GET",
        query_string={"op": "OPEN", "user.name": "testuser"},
    ).respond_with_data(
        b"", status=307, headers={"Location": redirect_url}
    )

    # Final GET (datanode) returns the payload
    httpserver.expect_request(
        redirect_target_path,
        method="GET",
        query_string={"op": "OPEN", "user.name": "testuser", "datanode": "true"},
    ).respond_with_data(b"Hello World!", content_type="text/plain")

    return httpserver

def test_webhdfs_open_and_read(webhdfs_server: HTTPServer):
    base = webhdfs_server.url_for("/webhdfs/v1/test.txt")
    url = f"{base}?op=OPEN&user.name=testuser"

    # Full read
    with remote_file.RemoteFile.open_webhdfs(url) as f:
        buf = bytearray(11)
        n = f.read(buf)
    assert n == 11
    assert bytes(buf) == b"Hello World!"

    # Partial read via offset ("World")
    with remote_file.RemoteFile.open_webhdfs(url) as f:
        part = bytearray(5)
        n2 = f.read(part, file_offset=6)
    assert n2 == 5
    assert bytes(part) == b"World"

    # Ensure expectations hit; fail with details if not
    webhdfs_server.check_assertions()
```

I haven't studied the WebHDFS protocol so I don't know if it is correct but maybe it is worth exploring?

kingcrimsontianyu:
Thanks for making me aware of the `pytest-httpserver` plugin. It's a great tool for server mocking. With this, https://github.com/rapidsai/kvikio/issues/634 can be closed too.

It takes some debugging exercises to get it to work, documented as follows.

## Debugging the hang
Initially, testing KvikIO's WebHDFS Python binding using `pytest-httpserver` ended up with a hang. When `pytest-httpserver` was replaced with Python built-in `http.server` and the server was launched in a daemon thread, the hang still persisted. Tracing of HTTP messages showed that the hang arose from the call to `curl_easy_perform()` on the client side. It turns out that this was attributed to Python's GIL: `curl_easy_perform()` is a blocking call, and GIL prevented the daemon server thread from reacting. In the newly added unit test, `concurrent.futures.ProcessPoolExecutor` has been used to avoid this GIL issue.

kingcrimsontianyu:
Where is still a hang on CI, which I cannot reproduce locally:
```
python/kvikio/tests/test_hdfs_io.py::TestWebHdfsOperations::test_get_file_size[] PASSED [ 61%]
python/kvikio/tests/test_hdfs_io.py::TestWebHdfsOperations::test_get_file_size[?op=OPEN] PASSED [ 61%]
python/kvikio/tests/test_hdfs_io.py::TestWebHdfsOperations::test_parallel_read --> hang!
```

kingcrimsontianyu:
CI hang has been fixed. Previously, `pytest-httpserver` was run on the main process and KvikIO on the second process. Block waiting on KvikIO's result prevents the server from responding. 
The fix is to swap the server and client processes: KvikIO is now run on the main process, and the server on the second process. No block waiting is performed.
There seems no need to use `pytest-httpserver` any more, so I've removed the dependency. Python's built-in `http.server` alone is sufficient to create a simple mock server.

python/kvikio/tests/test_hdfs_io.py
```diff
@@ -0,0 +1,234 @@
...
+
+LOCALHOST = "127.0.0.1"
+
+
+def find_free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind((LOCALHOST, 0))
+        s.listen(1)
+        port = s.getsockname()[1]
+    return port
```
madsbk: Let's unify with the fixtures in `test_s3_io.py` and move them to `conftest.py`?

https://github.com/rapidsai/kvikio/blob/a35bf582306553c40dc6eaa9c29a1e49ddb969fc/python/kvikio/tests/test_s3_io.py#L28-L40

madsbk (approved):
Nice!

python/kvikio/tests/test_hdfs_io.py
kingcrimsontianyu: Thanks. Good idea.

I made them free functions and put them in a new file `utils.py`. The reason I didn't put them in `conftest.py` is I'm not aware of a flexible way to change the fixture's scope on a per-file basis. In `test_s3_io.py`, they have a scope of session, while in `test_hdfs_io.py`, I prefer letting them have the function scope in case we may run tests in parallel in the future where several servers co-exist, each having to use a different port from the localhost.

kingcrimsontianyu:
Final improvement: used the neat fixture `xp` defined in `conftest.py` to test with both the host and device buffers.

PR #793, opened 2025-08-07
Add a unified remote I/O interface that infers the endpoint type from URL (1/2): C++ implementation

kingcrimsontianyu:
This PR adds a new remote I/O utility function `RemoteHandle::open(url)` that infers the remote endpoint type from the URL to facilitate `RemoteHandle` creation.

- Supported endpoint types include S3, S3 with presigned URL, WebHDFS, and generic HTTP/HTTPS.
- Optionally, instead of letting `open` figure it out, users can explicitly specify the endpoint type by passing an enum argument `RemoteEndpointType`.
- Optionally, users can provide an allowlist that restricts the endpoint candidates
- Optionally, users can specify the expected file size. This design is to fully support the existing constructor overload `RemoteHandle(endpoint, nbytes)`.

A byproduct of this PR is an internal utility class `UrlParser` that uses the idiomatic libcurl URL API to validate the URL against "[RFC 3986 plus](https://curl.se/docs/url-syntax.html)".

## This PR depends on
- [x] #791
- [x] #788


cpp/include/kvikio/remote_handle.hpp
```diff
@@ -258,6 +387,13 @@ class RemoteHandle {
   RemoteHandle(RemoteHandle const&)            = delete;
   RemoteHandle& operator=(RemoteHandle const&) = delete;
 
+  /**
+   * @brief Get the type of the remote file.
+   *
+   * @return The type of the remote file.
+   */
+  [[nodiscard]] RemoteEndpointType type() const noexcept;
```
madsbk: Might be good with another name that doesn't conflict with Python's `type` function?

cpp/src/detail/url.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+#include <curl/curl.h>
+#include <kvikio/detail/url.hpp>
+#include <kvikio/error.hpp>
+
+#define CHECK_CURL_URL_ERR(err_code) check_curl_url_err(err_code, __LINE__, __FILE__)
+
+namespace kvikio::detail {
+namespace {
+inline void check_curl_url_err(CURLUcode err_code, int line_number, char const* filename)
```
madsbk:
```suggestion
void check_curl_url_err(CURLUcode err_code, int line_number, char const* filename)
```

cpp/src/detail/url.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+CURLU* CurlUrlHandle::get() const { return _handle; }
+
+std::optional<std::string> UrlParser::extract_component(
+  CurlUrlHandle const& handle,
+  CURLUPart part,
+  std::optional<unsigned int> bitmask_component_flags,
+  std::optional<CURLUcode> allowed_err_code)
+{
+  if (!bitmask_component_flags) { bitmask_component_flags = 0U; }
```
madsbk:
```suggestion
  if (!bitmask_component_flags.has_value()) { bitmask_component_flags = 0U; }
```

cpp/src/remote_handle.cpp
```diff
@@ -348,8 +421,31 @@ void S3Endpoint::setup_range_request(CurlHandle& curl, std::size_t file_offset,
...
+      std::smatch match_result;
+      return std::regex_search(parsed_url.path.value(), match_result, pattern);
+    } else if ((parsed_url.scheme == "http") || (parsed_url.scheme == "https")) {
+      return url_has_aws_s3_http_format(url) && !S3EndpointWithPresignedUrl::is_url_valid(url);
+    }
+
+    return false;
+  } catch (...) {
+    return false;
+  }
```
madsbk:
```suggestion
  } catch (...) {}
  return false;
```

cpp/src/remote_handle.cpp
```diff
@@ -439,6 +535,95 @@ void S3EndpointWithPresignedUrl::setup_range_request(CurlHandle& curl,
...
+      case RemoteEndpointType::S3:
+        if (!S3Endpoint::is_url_valid(url)) return nullptr;
+        if (scheme.value() == "s3") {
+          auto const [bucket, object] = S3Endpoint::parse_s3_url(url);
+          return std::make_unique<S3Endpoint>(std::pair{bucket, object});
+        }
+        return std::make_unique<S3Endpoint>(url);
+
+      case RemoteEndpointType::S3_PRESIGNED_URL:
+        if (!S3EndpointWithPresignedUrl::is_url_valid(url)) return nullptr;
```
madsbk: Avoid 
```suggestion
        if (!S3EndpointWithPresignedUrl::is_url_valid(url)) {return nullptr;}
```

cpp/src/remote_handle.cpp
```diff
@@ -439,6 +535,95 @@ void S3EndpointWithPresignedUrl::setup_range_request(CurlHandle& curl,
...
+    }
+  };
+
+  std::unique_ptr<RemoteEndpoint> endpoint;
+
+  if (remote_endpoint_type == RemoteEndpointType::AUTO) {
+    // Try each allowed type in the order of allowlist
+    for (auto const& type : allow_list.value()) {
+      endpoint = create_endpoint(type);
+      if (endpoint) break;
```
madsbk:
```suggestion
      if (endpoint) {
          break;
      }
```

cpp/include/kvikio/remote_handle.hpp
kingcrimsontianyu: Thanks for the foresight. Changed to `remote_endpoint_type()` in both the endpoints and the remote handle.

cpp/src/detail/url.cpp
kingcrimsontianyu: Done.

cpp/src/detail/url.cpp
kingcrimsontianyu: Done. Also added `has_value()` in place of the conversion function `bool()` in a few other places.

cpp/src/remote_handle.cpp
kingcrimsontianyu: Done.

cpp/src/remote_handle.cpp
kingcrimsontianyu: Agreed. Done!

cpp/src/remote_handle.cpp
kingcrimsontianyu: Done.

cpp/include/kvikio/detail/url.hpp
```diff
@@ -0,0 +1,199 @@
...
+  };
+
+  /**
+   * @brief Parses the given URL according to RFC 3986 plus and extracts its components.
+   *
+   * @param url The URL string to parse
+   * @param bitmask_url_flags Optional flags for URL parsing. Common flags include:
+   *                          - CURLU_DEFAULT_SCHEME: Allows URLs without schemes
+   *                          - CURLU_NON_SUPPORT_SCHEME: Accept non-supported schemes
+   *                          - CURLU_URLENCODE: URL encode the path
```
madsbk: Is this from clang-format?
```suggestion
   *  - CURLU_DEFAULT_SCHEME: Allows URLs without schemes
   *  - CURLU_NON_SUPPORT_SCHEME: Accept non-supported schemes
   *  - CURLU_URLENCODE: URL encode the path
```

cpp/include/kvikio/detail/url.hpp
```diff
@@ -0,0 +1,199 @@
...
+   * @param bitmask_component_flags Optional flags for component extraction. Common flags include:
+   *                                - CURLU_URLDECODE: URL decode the component
+   *                                - CURLU_PUNYCODE: Return host as punycode
```
madsbk:
```suggestion
   *  - CURLU_URLDECODE: URL decode the component
   *  - CURLU_PUNYCODE: Return host as punycode
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -235,6 +282,88 @@ class RemoteHandle {
...
+   *   - HTTP/HTTPS
+   * @param remote_endpoint_type The type of remote endpoint. Default is RemoteEndpointType::AUTO
+   * which automatically detects the endpoint type from the URL. Can be explicitly set to
+   * RemoteEndpointType::S3, RemoteEndpointType::S3_PRESIGNED_URL, RemoteEndpointType::WEBHDFS, or
+   * RemoteEndpointType::HTTP to force a specific endpoint type.
+   * @param allow_list Optional list of allowed endpoint types. If provided:
+   *   - If remote_endpoint_type is RemoteEndpointType::AUTO, Types are tried in the exact order
+   * specified until a match is found.
+   *   - In explicit mode, the specified type must be in this list, otherwise an exception is
+   * thrown.
```
```suggestion
   *   - If remote_endpoint_type is RemoteEndpointType::AUTO, Types are tried in the exact order
   *     specified until a match is found.
   *   - In explicit mode, the specified type must be in this list, otherwise an exception is
   *     thrown.
```

cpp/src/remote_handle.cpp
```diff
@@ -348,8 +424,31 @@ void S3Endpoint::setup_range_request(CurlHandle& curl, std::size_t file_offset,
...
+
+      // Check whether the S3 object key exists
+      std::regex const pattern(R"(^/[^/]+$)", std::regex::icase);
+      std::smatch match_result;
+      return std::regex_search(parsed_url.path.value(), match_result, pattern);
+    } else if ((parsed_url.scheme == "http") || (parsed_url.scheme == "https")) {
+      return url_has_aws_s3_http_format(url) && !S3EndpointWithPresignedUrl::is_url_valid(url);
+    }
+
+    return false;
```
madsbk:
```suggestion
```

madsbk (approved):
Looks good, I only have minor suggestions

cpp/include/kvikio/detail/url.hpp
kingcrimsontianyu: I was trying to align with the first non-Doxygen comment character for each parameter, but that doesn't look good. Clang-format doesn't seem to care about the list style. Changed as suggested.

cpp/include/kvikio/detail/url.hpp
kingcrimsontianyu: Done.

cpp/include/kvikio/remote_handle.hpp
kingcrimsontianyu: Done.

cpp/src/remote_handle.cpp
kingcrimsontianyu: Removed!

PR #796, opened 2025-08-08
Improve KvikIO Python binding performance by releasing GIL wherever deemed necessary

### LINKED ISSUE #795: Release the GIL
We should release the GIL in our Cython bindings by using `with nogil:` around all API calls.

kingcrimsontianyu:
This PR improves the Python binding performance by releasing the Global Interpreter Lock (GIL) wherever necessary. The tasks include:
- For function declarations, add `nogil` if missing. Only one such case has been identified, which defines an embedded template function.
- At the call site of a C++ function, add `with nogil` context if missing. All the other changes fall into this category.

Closes #795

### DIFF
--- a/python/kvikio/kvikio/_lib/remote_handle.pyx
+++ b/python/kvikio/kvikio/_lib/remote_handle.pyx
@@ -124,7 +124,8 @@ cdef class RemoteFile:
         nbytes: Optional[int],
     ):
         cdef pair[string, string] bucket_and_object_names = _to_string_pair(
-            bucket_name, object_name)
+            bucket_name, object_name
+        )
         cdef unique_ptr[cpp_RemoteEndpoint] cpp_endpoint
 
         with nogil:


python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -111,12 +123,17 @@ cdef class RemoteFile:
         object_name: str,
         nbytes: Optional[int],
     ):
+        cdef pair[string, string] bucket_and_object_names = _to_string_pair(
+            bucket_name, object_name)
```
madsbk:
```suggestion
            bucket_name, object_name
        )
```

madsbk (approved):
Thanks @kingcrimsontianyu, looks good

python/kvikio/kvikio/_lib/remote_handle.pyx
kingcrimsontianyu: Thanks. Updated.

PR #800, opened 2025-08-18
Upgrade to nvCOMP 5.0.0.6

vuule:
Upgrade the nvCOMP dependency to 5.0.0.6. 
This library is not used directly, but it's till vendored and used in libcudf wheels. Future changes will completely remove the dependency in kvikIO.

Depends on https://github.com/rapidsai/rapids-cmake/pull/896

### DIFF
--- a/cmake/rapids_config.cmake
+++ b/cmake/rapids_config.cmake
@@ -11,7 +11,6 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 # =============================================================================
-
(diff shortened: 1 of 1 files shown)


vuule:
Added do not merge label so we don't accidentally merge before reverting changes here https://github.com/rapidsai/kvikio/pull/800/files#diff-64b517b8bafdd2024e48760683d9d249c60ce85f52807058570f1123036d0ac9, after the rapids-cmake PR is merged.

bdice (approved):
Approved, but we'll need to revert the rapids-cmake change before merging.

PR #798, opened 2025-08-14
Remove Python nvCOMP bindings and Zarr 2 support

vuule:
Removes the features that utilize nvCOMP - Python bindings and Zarr 2 support.

PR #801, opened 2025-08-19
Remove remaining nvcomp references

vyasr:
https://github.com/rapidsai/kvikio/pull/798 removed usage of nvcomp but left the linkage in place, kvikio extension modules still relied on nvcomp existing even though they didn't actually use any of its functionality. That is now causing problems in #800. Removing the linkage entirely here (while still revendoring manually until we can move the vendoring to cudf) should resolve that.


python/kvikio/kvikio/_lib/CMakeLists.txt
```diff
@@ -30,15 +30,5 @@ endif()
 rapids_cython_create_modules(
   CXX
   SOURCE_FILES "${cython_modules}"
-  LINKED_LIBRARIES kvikio::kvikio nvcomp::nvcomp
```
vyasr, line 33: This line is the main change needed to fix the current build failures in #800 (like https://github.com/rapidsai/kvikio/actions/runs/17052265746/job/48342529698?pr=800). The rest of the PR is just good cleanup.

PR #797, opened 2025-08-11
Use build cluster in devcontainers

trxcllnt:

PR #803, opened 2025-08-19
Build and test with CUDA 13.0.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test

Contributes to https://github.com/rapidsai/build-planning/issues/68

* updates to CUDA 13 dependencies in fallback entries in `dependencies.yaml` matrices (i.e., the ones that get written to `pyproject.toml` in source control)

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.


cpp/doxygen/main_page.md
```diff
@@ -30,20 +30,29 @@ For convenience we release Conda packages that makes it easy to include KvikIO i
...
 Install the **stable release** from the ``rapidsai`` channel with the following:
+
 ```sh
 # Install in existing environment
 mamba install -c rapidsai -c conda-forge libkvikio
+
 # Create new environment (CUDA 12)
 mamba create -n libkvikio-env -c rapidsai -c conda-forge cuda-version=12.9 libkvikio
+
+# Create new environment (CUDA 13)
```
bdice, line 38: I would like to advertise newer versions first.

cpp/src/shim/cuda.cpp
```diff
@@ -56,11 +56,21 @@ cudaAPI::cudaAPI()
...
+  // Because dlsym only considers function name, not the parameters, the wrapper MemcpyBatchAsync
+  // would fail at the call site if KvikIO was built with CUDA driver 13.0 and run on CUDA
+  // driver 12.8, or vice versa. We therefore require that the CUDA driver API versions at
+  // compile-time and at runtime be both within [12.8, 13.0), or both above 13.0, in order to use
+  // batch copy. If this is not satisfied, KvikIO falls back to the non-batch copy.
   try {
-    decltype(cuMemcpyBatchAsync)* fp;
-    get_symbol(fp, lib, KVIKIO_STRINGIFY(cuMemcpyBatchAsync));
-    MemcpyBatchAsync.set(fp);
+    if ((CUDA_VERSION >= 12080 && CUDA_VERSION < 13000 && driver_version >= 12080 &&
```
bdice: I feel like there must be a way for us to support CUDA 12.8 runtime with CUDA 13.0 driver. Can we investigate this a little further?

dependencies.yaml
```diff
@@ -190,31 +188,35 @@ dependencies:
...
-              - libnuma
+        packages:
+          - libcufile-dev
+          - libnuma
   depends_on_cupy:
     common:
       - output_types: conda
         packages:
-          - cupy>=12.0.0
+          - &cupy_unsuffixed cupy>=12.0.0
```
bdice: We need to bump to `cupy>=13.6` for CUDA 13. Maybe we should require that same pinning on CUDA 12, too? Historically I believe we've bumped `cupy` bounds for all CUDA versions whenever some new CUDA version needs a higher pinning.

I don't think we are testing `cupy==12.0.0` in our "oldest" dependencies jobs right now so I don't have a good idea of whether 12.0 still works. Bumping to 13.6 seems safest.

dependencies.yaml
```diff
@@ -326,9 +333,17 @@ dependencies:
     specific:
       - output_types: [conda, requirements, pyproject]
         matrices:
-          - matrix: # All CUDA 12 versions
+          - matrix:
+              cuda: "12.*"
             packages:
               - cuda-python>=12.6.2,<13.0a0
```
bdice: This should be bumped to 12.9.2 to address various segfaults / issues we've been seeing recently.
```suggestion
              - cuda-python>=12.9.2,<13.0a0
```

dependencies.yaml
```diff
@@ -326,9 +333,17 @@ dependencies:
...
+          - matrix:
+              cuda: "13.*"
+            packages:
+              - &cuda_python_cu13 cuda-python>=13.0.0,<14.0a0
```
bdice: Similarly:
```suggestion
              - &cuda_python_cu13 cuda-python>=13.0.1,<14.0a0
```

docs/source/install.rst
```diff
@@ -51,7 +61,7 @@ In order to setup a development environment, we recommend Conda:
 .. code-block::
 
   # CUDA 12
```
bdice:
```suggestion
  # CUDA 13
```

bdice (commented):
Requesting a few related version bumps that should ensure a smoother CUDA 13 experience.

cpp/src/shim/cuda.cpp
bdice: Many of our CUDA 12 tests are running with driver 580, which should cover this case.

docs/source/install.rst
jameslamb: Thanks for catching that, done in https://github.com/rapidsai/kvikio/pull/803/commits/6e7aff368526eb368cb2bf45913dc95f896b21f1

dependencies.yaml
jameslamb: updated in https://github.com/rapidsai/kvikio/pull/803/commits/6e7aff368526eb368cb2bf45913dc95f896b21f1

dependencies.yaml
jameslamb: updated in https://github.com/rapidsai/kvikio/pull/803/commits/6e7aff368526eb368cb2bf45913dc95f896b21f1

dependencies.yaml
jameslamb: Ok thanks for that historical context.

On https://anaconda.org/conda-forge/cupy/files?sort=basename&sort_order=desc&version=13.6.0 I do see `cupy==13.6.0` packages published for CUDA 11, 12, and 13, so hopefully this should be ok.

Updated in https://github.com/rapidsai/kvikio/pull/803/commits/6e7aff368526eb368cb2bf45913dc95f896b21f1, and I'll update the other CUDA 13 PRs (including `rapids-reviser` example).

cpp/doxygen/main_page.md
jameslamb: done in https://github.com/rapidsai/kvikio/pull/803/commits/6e7aff368526eb368cb2bf45913dc95f896b21f1, and I'll adjust the other CUDA 13 PRs accordingly.

cpp/src/shim/cuda.cpp
kingcrimsontianyu: Thanks.
It looks like because of backward compatibility the code compiled against CUDA driver 12.9 is able to dynamically load its own version of batched copy function from `libcuda.so.580.65.06` when run on the system with CUDA driver 13.0.
Handling simplified.

.devcontainer/cuda13.0-conda/devcontainer.json
bdice, line 1: @jameslamb Just as a heads-up, make sure the newly created devcontainers include Paul's build cluster changes. I think this looks right but just verify that for all the repos as you add CUDA 13 support to them.

docs/source/install.rst
```diff
@@ -15,19 +15,26 @@ Install the **stable release** from the ``rapidsai`` channel like:
 
   # Install in existing environment
   mamba install -c rapidsai -c conda-forge kvikio
+
   # Create new environment (CUDA 12)
   mamba create -n kvikio-env -c rapidsai -c conda-forge python=3.13 cuda-version=12.9 kvikio
 
+  # Create new environment (CUDA 13)
+  mamba create -n kvikio-env -c rapidsai -c conda-forge python=3.13 cuda-version=13.0 kvikio
```
bdice:
```suggestion
  # Create new environment (CUDA 13)
  mamba create -n kvikio-env -c rapidsai -c conda-forge python=3.13 cuda-version=13.0 kvikio

  # Create new environment (CUDA 12)
  mamba create -n kvikio-env -c rapidsai -c conda-forge python=3.13 cuda-version=12.9 kvikio
```

bdice (approved):
One small docs suggestion, otherwise LGTM

docs/source/install.rst
jameslamb: blegh missed this one, thank you

.devcontainer/cuda13.0-conda/devcontainer.json
jameslamb: Good note, thank you. I think we have what's needed here, based on what I see in the diff from #797

The only changes in the `devcontainer.json` files themselves were things like this, which this branch has:

```diff
  "runArgs": [
    "--rm",
    "--name",
    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-25.10-cuda13.0-conda",
+    "--ulimit",
+    "nofile=500000"
  ],
```

https://github.com/jameslamb/kvikio/blob/3162121ad72d303b150e7e80a0e61c0bac06fd3b/.devcontainer/cuda13.0-conda/devcontainer.json#L15-L16

PR #810, opened 2025-08-26
Fix an S3 parsing bug in the open function. Improve regex usage

kingcrimsontianyu:
AWS S3 provides a non-standard S3 scheme for internal use (such as for AWS CLI). The URL takes the form `s3://<bucket-name>/<object-name>`, where `<object-name>` may contain `/` characters indicating subdirectories.

The newly added `open` function for remote I/O currently uses an incorrect regular expression, causing object names containing subdirectories to be rejected. This PR fixes this bug.

This PR also improves the usage of regular expression by making the pattern constant `static`.

### DIFF
--- a/cpp/src/remote_handle.cpp
+++ b/cpp/src/remote_handle.cpp
@@ -320,7 +320,7 @@ std::pair<std::string, std::string> S3Endpoint::parse_s3_url(std::string const& 
 {
   KVIKIO_NVTX_FUNC_RANGE();
   // Regular expression to match s3://<bucket>/<object>
-  std::regex const pattern{R"(^s3://([^/]+)/(.+))", std::regex_constants::icase};
+  std::regex static const pattern{R"(^s3://([^/]+)/(.+))", std::regex_constants::icase};
   std::smatch matches;
   if (std::regex_match(s3_url, matches, pattern)) { return {matches[1].str(), matches[2].str()}; }
   KVIKIO_FAIL("Input string does not match the expected S3 URL format.", std::invalid_argument);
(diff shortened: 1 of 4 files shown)


cpp/tests/test_remote_handle.cpp
```diff
@@ -37,6 +37,7 @@ class RemoteHandleTest : public testing::Test {
     _sample_urls = {
       // Endpoint type: S3
       {"s3://bucket-name/object-key-name", kvikio::RemoteEndpointType::S3},
+      {"s3://bucket-name/object-key-name-dir/object-key-name-file", kvikio::RemoteEndpointType::S3},
```
kingcrimsontianyu, line 40: `branch-25.10` will fail on this test data.

cpp/src/remote_handle.cpp
```diff
@@ -434,9 +434,8 @@ bool S3Endpoint::is_url_valid(std::string const& url) noexcept
       if (!parsed_url.path.has_value()) { return false; }
 
       // Check whether the S3 object key exists
-      std::regex const pattern(R"(^/[^/]+$)", std::regex::icase);
-      std::smatch match_result;
-      return std::regex_search(parsed_url.path.value(), match_result, pattern);
+      std::regex static const pattern(R"(^/.+$)");
+      return std::regex_search(parsed_url.path.value(), pattern);
```
kingcrimsontianyu, line 438: `R"(^/[^/]+$)" ` is a wrong choice (matching a string starting with `/` and then with 1 or more non-`/` characters until the end). This won't work for `/foo/bar/tee`.
One solution is to simply drop the `$`. The other simple solution is to use `R"(^/.+$)"` .

madsbk (approved):

PR #815, opened 2025-09-09
Added KVIKIO_REMOTE_VERBOSE option

TomAugspurger:
Our HTTP library, libcurl, includes a [`CURLOPT_VERBOSE`](https://curl.se/libcurl/c/CURLOPT_VERBOSE.html) setting that can be useful for debugging. To help our users debug things, I've added a new `KVIKIO_REMOTE_VERBOSE` option that configures this. By default, it's off (no change). If the user sets `KVIKIO_REMOTE_VERBOSE=1` then information from the HTTP requests and responses will be printed to stderr.


cpp/src/shim/libcurl.cpp
```diff
@@ -112,6 +112,9 @@ CurlHandle::CurlHandle(LibCurl::UniqueHandlePtr handle,
 
   // Make requests time out after `value` seconds.
   setopt(CURLOPT_TIMEOUT, kvikio::defaults::http_timeout());
+
+  // Enable verbose output if configured in defaults
+  if (kvikio::defaults::http_verbose()) { setopt(CURLOPT_VERBOSE, 1L); }
```
kingcrimsontianyu: So the purpose of having the `http_verbose` parameter in the `defaults` class would be to allow runtime parameter change. Because of how CURL handles are managed in KvikIO, the change will only be applied to newly generated handles, unless we also explicitly reconfigure existing handles. I'm wondering if we really need that. Perhaps we could simply use:
```cpp
CurlHandle::CurlHandle(...) {
...
    auto const verbose = getenv_or("KVIKIO_HTTP_VERBOSE", false);
    if(verbose) {
        setopt(CURLOPT_VERBOSE, 1L);
    }
...
}
```

PS: The CURL handles in KvikIO are reused to minimize handle creation overhead. When an object of `CurlHandle` goes out of scope, it is not cleaned up, but instead handed back to a global singleton `LibCurl` which tracks a vector of free CURL handles eligible for reuse. So if we want to apply any runtime parameter change to CURL handles, we need to iterate all the free handles in `LibCurl`.

docs/source/runtime_settings.rst
```diff
@@ -53,3 +53,14 @@ KvikIO will retry a request should any of the HTTP status code in ``KVIKIO_HTTP_
...
+HTTP Verbose ``KVIKIO_HTTP_VERBOSE``
+------------------------------------
+
+For debugging HTTP requests, you can enable verbose output that shows detailed information about HTTP communication including headers, request/response bodies, connection details, and SSL handshake information.
+
+Set the environment variable ``KVIKIO_HTTP_VERBOSE`` to ``true``, ``on``, or ``yes`` (case-insensitive) to enable verbose output. Otherwise, verbose output is disabled by default.
+
+This setting can also be accessed using :py:func:`kvikio.defaults.http_verbose` (getter) and :py:func:`kvikio.defaults.set` (setter).
+
+**Warning** this may show sensitive contents from headers and data.
```
kingcrimsontianyu:
```suggestion
.. warning::

   This may show sensitive contents from headers and data.
```

kingcrimsontianyu: Sample output:

Image

cpp/doxygen/main_page.md
```diff
@@ -137,6 +137,13 @@ Note that if you're reading a large file that has been split into multiple reads
 
 These settings can also be controlled by `defaults::http_max_attempts()`, `defaults::http_max_attempts_reset()`, `defaults::http_status_codes()`, and `defaults::http_status_codes_reset()`.
 
+#### HTTP Verbose (KVIKIO_HTTP_VERBOSE)
+For debugging HTTP requests, you can enable verbose output that shows detailed information about HTTP communication including headers, request/response bodies, connection details, and SSL handshake information.
+
+Set the environment variable `KVIKIO_HTTP_VERBOSE` to `true`, `on`, or `yes` (case-insensitive) to enable verbose output. Otherwise, verbose output is disabled by default.
```
kingcrimsontianyu: Update suggested: `KVIKIO_REMOTE_VERBOSE`

docs/source/runtime_settings.rst
```diff
@@ -53,3 +53,14 @@ KvikIO will retry a request should any of the HTTP status code in ``KVIKIO_HTTP_
 The maximum number of attempts to make before throwing an exception is controlled by ``KVIKIO_HTTP_MAX_ATTEMPTS``. The default value is 3. This setting can also be accessed using :py:func:`kvikio.defaults.http_max_attempts` (getter) and :py:func:`kvikio.defaults.set` (setter).
 
 The maximum duration of each HTTP request is controlled by ``KVIKIO_HTTP_TIMEOUT``. The default value is 60, which is the duration in seconds to allow. This setting can also be accessed using :py:func:`kvikio.defaults.http_timoeout` (getter) and :py:func:`kvikio.defaults.set` (setter).
+
+HTTP Verbose ``KVIKIO_REMOTE_VERBOSE``
+------------------------------------
```
Matt711:
```suggestion
HTTP Verbose ``KVIKIO_REMOTE_VERBOSE``
--------------------------------------
```

TomAugspurger: Thanks!

cpp/doxygen/main_page.md
```diff
@@ -137,6 +137,13 @@ Note that if you're reading a large file that has been split into multiple reads
 
 These settings can also be controlled by `defaults::http_max_attempts()`, `defaults::http_max_attempts_reset()`, `defaults::http_status_codes()`, and `defaults::http_status_codes_reset()`.
 
+#### Remote Verbose (KVIKIO_REMOTE_VERBOSE)
+For debugging HTTP requests, you can enable verbose output that shows detailed information about HTTP communication including headers, request/response bodies, connection details, and SSL handshake information.
+
+Set the environment variable `KVIKIO_REMOTE_VERBOSE` to `true`, `on`, or `yes` (case-insensitive) to enable verbose output. Otherwise, verbose output is disabled by default.
```
bdice: How about `1`? That is my favorite way to enable things.

bdice (approved):

cpp/doxygen/main_page.md
TomAugspurger: Yep, based on https://github.com/rapidsai/kvikio/blob/f4e022e9fcb768421c6e6ab5dddb0611b746fac8/cpp/src/defaults.cpp#L41-L42.

Updated the docs to note that.

Issue #806, opened 2025-08-20
Support public S3 URLs with anonymous requests


TomAugspurger:
Some S3 buckets are configured to allow anonymous access. For example,

```
❯ aws s3 ls  --region us-west-2 --no-sign-request s3://overturemaps-us-west-2/release/2024-09-18.0/theme=base/type=water/part-00000-efc2874f-4a6c-49f9-a908-df13eb0e1c63-c000.zstd.parquet
2024-09-17 16:39:28 1474217942 part-00000-efc2874f-4a6c-49f9-a908-df13eb0e1c63-c000.zstd.parquet
```

Currently, trying to read from that URL with kvikio will raise an error:

```
In [2]: kvikio.RemoteFile.open_s3_url("s3://overturemaps-us-west-2/release/2024-09-18.0/theme=base/type=water/part-00000-efc2874f-4a6c-49f9-a908-df13eb0e1c63-c000.zstd.parquet")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 kvikio.RemoteFile.open_s3_url("s3://overturemaps-us-west-2/release/2024-09-18.0/theme=base/type=water/part-00000-efc2874f-4a6c-49f9-a908-df13eb0e1c63-c000.zstd.parquet")

File /raid/toaugspurger/envs/gh/rapidsai/cudf/lib/python3.13/site-packages/kvikio/remote_file.py:141, in RemoteFile.open_s3_url(cls, url, nbytes)
    136     return RemoteFile(
    137         _get_remote_module().RemoteFile.open_s3_from_http_url(url, nbytes)
    138     )
    139 if parsed_result.scheme == "s3":
    140     return RemoteFile(
--> 141         _get_remote_module().RemoteFile.open_s3_from_s3_url(url, nbytes)
    142     )
    143 raise ValueError(f"Unsupported protocol: {url}")

File kvikio/_lib/remote_handle.pyx:174, in kvikio._lib.remote_handle.RemoteFile.open_s3_from_s3_url()

ValueError: KvikIO failure at: /__w/kvikio/kvikio/cpp/src/remote_handle.cpp:217: S3: must provide `aws_access_key` if AWS_ACCESS_KEY_ID isn't set.
```

In this case, we set `AWS_DEFAULT_REGION` for kvikio, but we don't have an access key ID or secret access key to set. We want to make an unsigned request to the S3 service.

kingcrimsontianyu:
[...]

PR #820, opened 2025-09-12
Support access to public S3

kingcrimsontianyu:
## Background
KvikIO supports access to private S3 objects that require AWS credentials:
```python
# Method 1
kvikio.RemoteFile.open_s3(bucket, key)

# Method 2
kvikio.RemoteFile.open_s3_url(url)

# Method 3
kvikio.RemoteFile.open(url)
```
For public S3 object, these functions will throw the following exceptions.
>S3: must provide `aws_region` if AWS_DEFAULT_REGION isn't set.

A workaround is to simply use the generic HTTP/HTTPS endpoint:
```python
# Method 1
kvikio.RemoteFile.open_http(http_url)

# Method 2
kvikio.RemoteFile.open(url, RemoteEndpointType.HTTP)
```

However, this workaround loses the feature of S3 URL syntax check.

## This PR
- Adds support for accessing public S3 objects in C++ and Python by having a new endpoint type `S3PublicEndpoint`. This endpoint does not require AWS credentials.
- Updates the unified interface `open(url)` that can automatically infer the endpoint type. Under `AUTO` mode, for a syntactically valid S3 URL using HTTP/HTTPS protocol, KvikIO now checks the connectivity using a private S3 endpoint, and if failed proceeds to use a public S3 endpoint.
- Updates the comments on each endpoint to further improve clarity.
- Adjusts Python APIs `kvikio.RemoteFile.open_*` from class method to static method (which is a breaking change).

Closes https://github.com/rapidsai/kvikio/issues/806

### DIFF
--- a/python/kvikio/kvikio/remote_file.py
+++ b/python/kvikio/kvikio/remote_file.py
@@ -101,8 +101,9 @@ class RemoteFile:
         assert isinstance(handle, _get_remote_module().RemoteFile)
         self._handle = handle
 
-    @staticmethod
+    @classmethod
     def open_http(
+        cls,
         url: str,
         nbytes: Optional[int] = None,
     ) -> RemoteFile:
@@ -116,10 +117,11 @@ class RemoteFile:
             The size of the file. If None, KvikIO will ask the server
             for the file size.
         """
-        return RemoteFile(_get_remote_module().RemoteFile.open_http(url, nbytes))
+        return cls(_get_remote_module().RemoteFile.open_http(url, nbytes))
 
-    @staticmethod
+    @classmethod
     def open_s3(
+        cls,
         bucket_name: str,
         object_name: str,
         nbytes: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


python/kvikio/kvikio/remote_file.py
```diff
@@ -96,13 +101,12 @@ def __init__(self, handle):
         assert isinstance(handle, _get_remote_module().RemoteFile)
         self._handle = handle
 
-    @classmethod
+    @staticmethod
```
kingcrimsontianyu: These `open_*` functions don't change class-scope variables. So I think for simplicity, we could just make them all `static`, removing the unused parameter `cls`.

madsbk: Since they are factory functions, I would prefer `@classmethod`. But let's use the `cls` like:
```python
return cls.__class__(_get_remote_module().RemoteFile.open_http(url, nbytes))
```
WDYT?

kingcrimsontianyu: Thanks! Modified as suggested.
For the return value I used `cls(...)` to create a new instance.

Issue #485, opened 2024-10-03
[FEA]: CI job to build kvikio in no-cuda mode


wence-:
As seen in (for example) #482, it is easy to accidentally introduce cuda-isms into the build that break no-cuda mode.

We should add a CI job that attempts to build with cuda enabled (on a system that has no CTK install) to catch these preemptively.

PR #822, opened 2025-09-15
Add a conda C++ CPU-only build

### LINKED ISSUE #736: CPU-only s3 feftching?
after seeing the NVIDIA blog post where kvikio is significantly faster to fetch S3 data due to the task architecture and multithreadedness, I wanted to try this on a CPU-only instance, without CUDA, however, I couldn't find clear guidance on how to do this.

the build system wants to determine the cuda version so it can appent cu11 or cu12 so that fails without CUDA, despite there being a build option enabling CPU-only builds.

is this supposed to work? if so, can you provide guidance how to achieve such a thing?

vyasr:
We should decide what suitable constraints on the CPU-only package are. I don't think there is ever a reason to support installing the CPU-only package into an environment with CUDA since the CUDA-enabled package's functionality is a strict superset, so I've added a `run_constraints` on `cuda-version<0` on the CPU package. `__cuda<0` is too strong of a restriction since it would prevent creating an environment to test CPU-only kvikio on a machine with CUDA drivers installed on the host.

Resolves #736 and #485

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -28,6 +28,9 @@ source rapids-rattler-channel-string
 # Construct the extra variants according to the architecture
 if [[ "$(arch)" == "x86_64" ]]; then
     cat > variants.yaml << EOF
+    zip_keys:
+    - [cuda_version, supports_cuda]
+
     c_compiler_version:
       - 14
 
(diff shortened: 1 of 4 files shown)


ci/build_cpp.sh
```diff
@@ -28,31 +28,47 @@ source rapids-rattler-channel-string
...
       - 14
 
     cxx_compiler_version:
       - 14
 
     cuda_version:
+      - 0
       - ${RAPIDS_CUDA_VERSION%.*}
+
+    supports_cuda:
```
bdice, line 44: Maybe rename this to `cuda_support_enabled`? The CMake option is `-DKvikio_CUDA_SUPPORT`.

vyasr:
Upon further discussion, we decided that we no longer need to support a no-cuda build at all and are simply planning to remove it. I'll close this PR and open that instead.

Issue #823, opened 2025-09-15
403 Error when reading S3 URLs with special characters


TomAugspurger:
This snippet gets a 403 error when reading an S3 URL with `=` in the object key:

```python
import boto3
import os
import kvikio
import rmm

client = boto3.client(
    's3',
    aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
    aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
    aws_session_token=os.environ['AWS_SESSION_TOKEN'],
    region_name=os.environ['AWS_DEFAULT_REGION']
)

bucket = ...

for key in ["test/date-2025-09-15", "test/date=2025-09-15"]:
    client.put_object(Bucket=bucket, Key=key, Body=b'Hello, world!')
    client.head_object(Bucket=bucket, Key=key)

    print("boto3 done", key)

    buf = rmm.DeviceBuffer(size=13)
    f = kvikio.RemoteFile.open_s3(bucket, key)
    f.read(buf)

    print("kvikio done", key)
```

Which outputs

```
boto3 done test/date-2025-09-15
kvikio done test/date-2025-09-15
boto3 done test/date=2025-09-15
Traceback (most recent call last):
  File "/home/coder/kvikio/debug.py", line 23, in <module>
    f = kvikio.RemoteFile.open_s3(bucket, key)
  File "/home/coder/kvikio/python/kvikio/kvikio/remote_file.py", line 146, in open_s3
    _get_remote_module().RemoteFile.open_s3(bucket_name, object_name, nbytes)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "kvikio/_lib/remote_handle.pyx", line 178, in kvikio._lib.remote_handle.RemoteFile.open_s3
  File "kvikio/_lib/remote_handle.pyx", line 135, in kvikio._lib.remote_handle.RemoteFile._from_endpoint
RuntimeError: KvikIO failure at: /home/coder/kvikio/cpp/src/shim/libcurl.cpp:183: curl_easy_perform() error (The requested URL returned error: 403)
```

This is similar to https://github.com/curl/curl/issues/13754, which is fixed in recent versions of curl. It's possible that this is fixed if we use a newer version of libcurl.

TomAugspurger:
Some more debugging: Add

```
import logging

[...]

PR #824, opened 2025-09-15
Upgrade CURL to 8.16.0

kingcrimsontianyu:


TomAugspurger:
I've rerun my manual test from https://github.com/rapidsai/kvikio/issues/823, and it still fails on this branch unfortunately.

I've also tried to add a moto-based unit test:

```python
def test_read_with_special_characters(s3_base):
    bucket_name = "bucket"
    object_name = "date=2025-09-15"
    data = b"file content"
    with s3_context(
        s3_base=s3_base, bucket=bucket_name, files={object_name: bytes(data)}
    ) as server_address:
        with kvikio.RemoteFile.open_s3_url(f"s3://{bucket_name}/{object_name}") as f:
            assert f.nbytes() == len(data)
            got = bytearray(len(data))
            assert f.read(got) == len(got)
```

Unfortunately, it *doesn't* fail on `branch-25.10` or this branch. So however moto is handling tokens, it isn't emulating this portion the same as S3.

PR #829, opened 2025-09-18
Remove ability to build kvikio without CUDA

### LINKED ISSUE #485: [FEA]: CI job to build kvikio in no-cuda mode

### LINKED ISSUE #736: CPU-only s3 feftching?

vyasr:
This change supersedes #822 and makes CUDA a requirement for building kvikio. We have no compelling use case for this support outside of [legate](https://docs.nvidia.com/legate/latest/api/cpp/generated/group/group__io-kvikio.html), and after discussion with the team they are happy to sunset this support as well.

Resolves #736 
Resolves #485


bdice (approved):

bdice:
Due to the proximity to burndown, maybe we should shift this to 25.12 tomorrow?

PR #846, opened 2025-09-25
Add named parameters for S3 credentials to Python API

TomAugspurger:
`S3Endpoint` takes optional parameters for the AWS region, access key ID, etc. If these aren't set, they're looked up from the environment.

Previously, the only way to specify these from Python was via environment variables. This adds named parameters to `f = kvikio.RemoteFile.open_s3` so that users can specify the credentials programatically. The default behavior is unchanged: environment variables are used when not specified otherwise.

Here's a test snippet against an S3 bucket:

```python
import sys
import boto3
import kvikio
import rmm


bucket, access_key_id, secret_access_key, session_token, default_region = sys.argv[1:]

client = boto3.client(
    's3',
    aws_access_key_id=access_key_id,
    aws_secret_access_key=secret_access_key,
    aws_session_token=session_token,
    region_name=default_region,
)

key = "test/date-2025-09-16"
client.put_object(Bucket=bucket, Key=key, Body=b'Hello, world!')
client.head_object(Bucket=bucket, Key=key)

buf = rmm.DeviceBuffer(size=13)
f = kvikio.RemoteFile.open_s3(bucket, key, access_key_id=access_key_id, secret_access_key=secret_access_key, session_token=session_token, region_name=default_region)
f.read(buf)
print(buf.tobytes())
```

I've set those variables to `_`-prefixed versions. When run, that prints

```
❯ python debug.py kvikiobench-33622 $_AWS_ACCESS_KEY_ID $_AWS_SECRET_ACCESS_KEY $_AWS_SESSION_TOKEN $_AWS_DEFAULT_REGION
b'Hello, world!'
```

### DIFF
--- a/python/kvikio/kvikio/_lib/remote_handle.pyx
+++ b/python/kvikio/kvikio/_lib/remote_handle.pyx
@@ -102,7 +102,7 @@ cdef pair[string, string] _to_string_pair(str s1, str s2):
     """Wrap two Python string objects in a C++ pair"""
     return pair[string, string](_to_string(s1), _to_string(s2))
 
-cdef optional[string] _to_optional_string(Optional[str] s):
+cdef optional[string] _to_optional_string(str s):
     """Convert Python object to a C++ optional string (if None, return nullopt)"""
     cdef optional[string] result
     if s is None:


Matt711:
> I opened this, since I thought it might be a blocker for [rapidsai/cudf#19757](https://github.com/rapidsai/cudf/issues/19757). But upon reflection I don't think it will be; cudf-polars uses kvikio via pylibcudf. My understanding is that that won't use this Python API.
> 
> Still, it might be useful to support this.

Yeah it shouldn't. FYI I was planning on tackling https://github.com/rapidsai/cudf/issues/19757 in 25.12. Is that okay with you? I wanted to ask beforehand.

TomAugspurger:
It's all yours :)

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -35,7 +35,22 @@ cdef extern from "<kvikio/remote_handle.hpp>" namespace "kvikio" nogil:
 
     cdef cppclass cpp_S3Endpoint "kvikio::S3Endpoint"(cpp_RemoteEndpoint):
         cpp_S3Endpoint(string url) except +
```
kingcrimsontianyu: This can be removed since it is unused now.

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -35,7 +35,22 @@ cdef extern from "<kvikio/remote_handle.hpp>" namespace "kvikio" nogil:
...
+        cpp_S3Endpoint(
+            string url,
+            optional[string] aws_region,
+            optional[string] aws_access_key,
+            optional[string] aws_secret_access_key,
+            optional[string] aws_session_token
+        ) except +
         cpp_S3Endpoint(pair[string, string] bucket_and_object_names) except +
```
kingcrimsontianyu: Same as above. The unary overload can be removed.

python/kvikio/kvikio/remote_file.py
```diff
@@ -125,16 +125,25 @@ def open_s3(
         bucket_name: str,
         object_name: str,
         nbytes: Optional[int] = None,
+        region_name: Optional[str] = None,
```
kingcrimsontianyu: I think we could make the parameter naming more consistent by adding `aws_` prefix here and in the Cython code, to match the environment variables which have the `AWS_` prefix.

python/kvikio/kvikio/_lib/remote_handle.pyx
TomAugspurger: Thanks, I was wondering about that. Does open_s3_public still use this version at https://github.com/rapidsai/kvikio/blob/19ea7581b1ffc595cb56b6f39bc73fd99dd86744/python/kvikio/kvikio/_lib/remote_handle.pyx#L235?

python/kvikio/kvikio/_lib/remote_handle.pyx
```diff
@@ -89,6 +104,15 @@ cdef pair[string, string] _to_string_pair(str s1, str s2):
     """Wrap two Python string objects in a C++ pair"""
     return pair[string, string](_to_string(s1), _to_string(s2))
 
+cdef optional[string] _to_optional_string(s):
```
kingcrimsontianyu:
```suggestion
cdef optional[string] _to_optional_string(str s):
```
Nit: specify the type as we did for `_to_string`.

python/kvikio/kvikio/_lib/remote_handle.pyx
kingcrimsontianyu: `cpp_S3PublicEndpoint` is a separate class from `cpp_S3Endpoint`, so we are good :)

python/kvikio/kvikio/_lib/remote_handle.pyx
TomAugspurger: Should the type be `Optional[str]`? That seems to compile at least.

python/kvikio/kvikio/_lib/remote_handle.pyx
TomAugspurger: Good point :)

python/kvikio/kvikio/_lib/remote_handle.pyx
kingcrimsontianyu: Interesting. I'm not sure to what extent Python's type hint system applies to the parameter type of `cdef` function. If here `Optional[str]` is the idiomatic way, then the utility function `_to_string(str s)` should also use  `Optional[str]`. Could you check if no error would occur at runtime if `None` is passed to these functions?

TomAugspurger: All seems to work well either way:

<details>

```python
import os
import boto3
import kvikio
import rmm


bucket = 'kvikiobench-33622'
aws_access_key_id = os.environ['AWS_ACCESS_KEY_ID']
aws_secret_access_key = os.environ['AWS_SECRET_ACCESS_KEY']
aws_session_token = os.environ['AWS_SESSION_TOKEN']
aws_region_name = os.environ['AWS_DEFAULT_REGION']

client = boto3.client(
    's3',
)

key = "test/date-2025-09-16"
client.put_object(Bucket=bucket, Key=key, Body=b'Hello, world!')
client.head_object(Bucket=bucket, Key=key)

buf = rmm.DeviceBuffer(size=13)
f = kvikio.RemoteFile.open_s3(bucket, key, aws_access_key_id=None, aws_secret_access_key=None, aws_session_token=None, aws_region_name=None)
f.read(buf)
print("Read env")
print(buf.tobytes())

buf = rmm.DeviceBuffer(size=13)
print("Read, explicit")
f = kvikio.RemoteFile.open_s3(bucket, key, aws_access_key_id=aws_access_key_id, aws_secret_access_key=aws_secret_access_key, aws_session_token=aws_session_token, aws_region_name=aws_region_name)
f.read(buf)
print(buf.tobytes())
```

</details>

https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#extension-types-and-none documents this I think:

> In [cdef](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#cdef) declarations and C-like function argument declarations (func(list x)), when you declare an argument or C variable as having an extension or Python builtin type, Cython will allow it to take on the value None as well as values of its declared type

So I think we're OK using either. I've matched the way it's already done.

kingcrimsontianyu: Thanks for locating the doc.

kingcrimsontianyu: Btw, I tested `_to_string(Optional[str] s)` and `_to_optional_string(Optional[str] s)`. Both work well. So this may be a new feature in Cython. We may still use `str s` in this PR to be consistent with the current doc.
```python
# remote_handle.pyx
cdef extern from * nogil:
    """
    std::string duplicate(std::string s)
    {
        if(s.empty()) {
            return "empty input";
        }
        return s + s;
    }
    """
    cdef string duplicate(string s) except +

def check(s: Optional[str]) -> str:
    print(duplicate(_to_string(s)))

cdef extern from * nogil:
    """
    std::string duplicate2(std::optional<std::string> s)
    {
        if(s == std::nullopt) {
            return "nullopt input";
        }
        return s.value() + s.value();
    }
    """
    cdef string duplicate2(optional[string] s) except +

def check2(s: Optional[str]) -> str:
    print(duplicate2(_to_optional_string(s)))
```
```python
# remote_file.py
class RemoteFile:
    @staticmethod
    def check(s: Optional[str]):
        _get_remote_module().check(s)

    @staticmethod
    def check2(s: Optional[str]):
        _get_remote_module().check2(s)
```
```python
# test
    kvikio.RemoteFile.check("s3")  # b's3s3'
    kvikio.RemoteFile.check(None)  # b'empty input'

    kvikio.RemoteFile.check2("s3")  # b's3s3'
    kvikio.RemoteFile.check2(None)  # b'nullopt input'
```

kingcrimsontianyu (approved):
Thanks for the work!

PR #847, opened 2025-09-30
feat: Add range-based locking for parallel file I/O

a-hirota:
## Description
  This PR implements range-based locking mechanism for kvikio to enable truly parallel non-overlapping writes to files. This significantly
   improves performance for multi-GPU workloads.

  ## Key Changes
  - Added `RangeLockManager` class for managing non-overlapping range locks
  - Added `FileHandleWithRangeLock` extending `FileHandle` with range lock support
  - Comprehensive test coverage in both C++ and Python

  ## Performance Benefits
  - Non-overlapping writes can execute in parallel
  - Reduces contention for multi-GPU file I/O operations
  - Maintains data integrity by serializing only overlapping ranges

  ## Usage Example

  ### C++ Usage
  ```cpp
  #include <kvikio/file_handle_rangelock.hpp>

  // Create a file handle with range lock support
  kvikio::FileHandleWithRangeLock file("output.bin", "w+");

  // Multiple threads/GPUs can write to non-overlapping regions in parallel
  std::vector<std::thread> threads;

  // GPU 0 writes to first half of file
  threads.emplace_back([&]() {
      void* gpu0_data = ...;  // GPU 0 data
      auto future = file.pwrite_rangelock(gpu0_data, size, 0);
      future.get();
  });

  // GPU 1 writes to second half - executes in parallel!
  threads.emplace_back([&]() {
      void* gpu1_data = ...;  // GPU 1 data
      auto future = file.pwrite_rangelock(gpu1_data, size, size);
      future.get();
  });

  for (auto& t : threads) {
      t.join();
  }
```

  ### Python Usage (Future API)
```python
  import kvikio
  import cupy as cp
  from concurrent.futures import ThreadPoolExecutor

  # When Python bindings are added:
  def write_partition(gpu_id, file_handle, offset, data):
      with cp.cuda.Device(gpu_id):
          # This would execute in parallel for non-overlapping regions
          file_handle.pwrite_rangelock(data, file_offset=offset)

  # Multiple GPUs writing to different file regions
  with kvikio.FileHandleWithRangeLock("output.bin", "w+") as f:
      with ThreadPoolExecutor() as executor:
          futures = []
          for gpu_id in range(num_gpus):
              offset = gpu_id * partition_size
              futures.append(
                  executor.submit(write_partition, gpu_id, f, offset, data[gpu_id])
              )
          # All non-overlapping writes execute in parallel
          for future in futures:
              future.result()
```

  ##  Testing

  - Added C++ tests in cpp/tests/test_range_lock.cpp
  - Added Python tests in python/kvikio/tests/test_range_lock.py
  - Tests cover:
    - Non-overlapping parallel writes
    - Overlapping range serialization
    - Move semantics
    - Performance benchmarks

  ##  Performance Impact

  In our tests with 2 GPUs writing to non-overlapping regions:
  - Without range lock: Serial execution (one GPU waits for the other)
  - With range lock: Parallel execution (both GPUs write simultaneously)
  - Expected speedup: Near-linear with number of GPUs for non-overlapping writes

### DIFF
--- a/cpp/include/kvikio/file_handle_rangelock.hpp
+++ b/cpp/include/kvikio/file_handle_rangelock.hpp
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Modified FileHandle with range-based locking support
+ */
+#pragma once
+
+#include <kvikio/file_handle.hpp>
+#include <kvikio/range_lock.hpp>
+#include <memory>
+
+namespace kvikio {
+
+class FileHandleWithRangeLock : public FileHandle {
+private:
+    mutable RangeLockManager range_lock_manager_;
+
+public:
+    using FileHandle::FileHandle;  // Inherit constructors
+
+    /**
+     * @brief Write with range-based locking
+     *
+     * This version acquires a lock only for the specific range being written,
+     * allowing non-overlapping writes to proceed in parallel.
+     */
+    std::future<std::size_t> pwrite_rangelock(void const* buf,
+                                              std::size_t size,
+                                              std::size_t file_offset = 0,
+                                              std::size_t task_size = defaults::task_size(),
+                                              std::size_t gds_threshold = defaults::gds_threshold(),
+                                              bool sync_default_stream = true) {
+
+        // Acquire range lock for this write
+        auto range_lock = range_lock_manager_.lock_range(file_offset, file_offset + size);
+
+        // Perform the write using the base class implementation
+        auto future = this->pwrite(buf, size, file_offset, task_size, gds_threshold, sync_default_stream);
+
+        // Create a wrapper future that releases the lock when done
+        return std::async(std::launch::deferred, [future = std::move(future),
+                                                  lock = std::move(range_lock)]() mutable {
+            auto result = future.get();
+            // Lock will be automatically released when this lambda exits
+            return result;
+        });
+    }
+
+    /**
+     * @brief Read with range-based locking (optional, for consistency)
+     */
+    std::future<std::size_t> pread_rangelock(void* buf,
+                                             std::size_t size,
+                                             std::size_t file_offset = 0,
+                                             std::size_t task_size = defaults::task_size(),
+                                             std::size_t gds_threshold = defaults::gds_threshold(),
+                                             bool sync_default_stream = true) {
+
+        // For reads, we could use shared locks if needed
+        // For now, using exclusive locks for simplicity
+        auto range_lock = range_lock_manager_.lock_range(file_offset, file_offset + size);
+
+        auto future = this->pread(buf, size, file_offset, task_size, gds_threshold, sync_default_stream);
+
+        return std::async(std::launch::deferred, [future = std::move(future),
+                                                  lock = std::move(range_lock)]() mutable {
+            auto result = future.get();
+            return result;
+        });
+    }
+
+    /**
+     * @brief Check if a range is currently locked
+     */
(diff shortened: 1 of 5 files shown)


cpp/include/kvikio/file_handle_rangelock.hpp
```diff
@@ -0,0 +1,87 @@
...
+     */
+    std::future<std::size_t> pwrite_rangelock(void const* buf,
+                                              std::size_t size,
+                                              std::size_t file_offset = 0,
+                                              std::size_t task_size = defaults::task_size(),
+                                              std::size_t gds_threshold = defaults::gds_threshold(),
+                                              bool sync_default_stream = true) {
+
+        // Acquire range lock for this write
+        auto range_lock = range_lock_manager_.lock_range(file_offset, file_offset + size);
```
kingcrimsontianyu, line 35: KvikIO has transitioned from a header-only library to a shared library. So all the implementations should go to corresponding `.cpp` files.

cpp/include/kvikio/file_handle_rangelock.hpp
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Modified FileHandle with range-based locking support
+ */
```
kingcrimsontianyu, line 5: Preamble should follow this complete format: https://github.com/rapidsai/kvikio/blob/branch-25.12/cpp/include/kvikio/file_handle.hpp

cpp/include/kvikio/file_handle_rangelock.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+public:
+    using FileHandle::FileHandle;  // Inherit constructors
+
+    /**
+     * @brief Write with range-based locking
+     *
+     * This version acquires a lock only for the specific range being written,
+     * allowing non-overlapping writes to proceed in parallel.
+     */
```
kingcrimsontianyu, line 26: Function parameters and return values should be commented in Doxygen format.

cpp/include/kvikio/file_handle_rangelock.hpp
```diff
@@ -0,0 +1,87 @@
...
+ */
+#pragma once
+
+#include <kvikio/file_handle.hpp>
+#include <kvikio/range_lock.hpp>
+#include <memory>
+
+namespace kvikio {
+
+class FileHandleWithRangeLock : public FileHandle {
```
kingcrimsontianyu, line 14: I think the class needs to be commented in detail too, to explain the purpose of this class and also key implementation details. For example, when multiple write requests contend on a common range, what would the behavior be?

cpp/include/kvikio/file_handle_rangelock.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+        // Perform the write using the base class implementation
+        auto future = this->pwrite(buf, size, file_offset, task_size, gds_threshold, sync_default_stream);
+
+        // Create a wrapper future that releases the lock when done
+        return std::async(std::launch::deferred, [future = std::move(future),
+                                                  lock = std::move(range_lock)]() mutable {
+            auto result = future.get();
+            // Lock will be automatically released when this lambda exits
+            return result;
```
kingcrimsontianyu, line 45: For parallel I/O, the step to wait for the chunked tasks is performed in the thread pool  (https://github.com/rapidsai/kvikio/blob/branch-25.12/cpp/include/kvikio/parallel_operation.hpp#L175-L184). Would it be possible that this `unlock` step here follows suit and gets pushed to the thread pool's task queue as well?

PR #848, opened 2025-10-07
Make parameter name more consistent for S3 utility functions

kingcrimsontianyu:
This small PR adds `aws_` prefix to the parameter `session_token` to make the parameter names more consistent for S3 utility functions.


madsbk (approved):

TomAugspurger (approved):
Sorry for missing these. No deprecation is needed since we haven't had a release with these parameters yet.

PR #851, opened 2025-10-15
Minimize transitive includes by moving NVTX-related code to the detail namespace

kingcrimsontianyu:
This PR is part of the effort to minimize transitive includes in KvikIO shared library. It moves the NVTX-related code from the public headers to the `detail` namespace.
As a result, the files `parallel_operation.hpp` and `posix_io.hpp` have also been moved to the `detail` namespace.

### DIFF
--- a/cpp/include/kvikio/threadpool_wrapper.hpp
+++ b/cpp/include/kvikio/threadpool_wrapper.hpp
@@ -16,8 +16,6 @@
 
 #pragma once
 
-#include <functional>
-
 #include <BS_thread_pool.hpp>
 
 namespace kvikio {
@@ -26,29 +24,18 @@ template <typename pool_type>
 class thread_pool_wrapper : public pool_type {
  public:
   /**
-   * @brief Construct a new thread pool wrapper, and invoke an initialization function in each
-   * worker thread.
+   * @brief Construct a new thread pool wrapper.
    *
    * @param nthreads The number of threads to use.
-   * @param worker_thread_init_func Initialization function invoked in each worker thread.
    */
-  thread_pool_wrapper(
-    unsigned int nthreads, std::function<void()> worker_thread_init_func = [] {})
-    : pool_type{nthreads, worker_thread_init_func}
-  {
-  }
+  thread_pool_wrapper(unsigned int nthreads) : pool_type{nthreads} {}
 
   /**
-   * @brief Reset the number of threads in the thread pool, and invoke an initialization function in
-   * each worker thread.
+   * @brief Reset the number of threads in the thread pool.
    *
    * @param nthreads The number of threads to use.
-   * @param worker_thread_init_func Initialization function invoked in each worker thread.
    */
-  void reset(unsigned int nthreads, std::function<void()> worker_thread_init_func = [] {})
-  {
-    pool_type::reset(nthreads, worker_thread_init_func);
-  }
+  void reset(unsigned int nthreads) { pool_type::reset(nthreads); }
 };
 
 using BS_thread_pool = thread_pool_wrapper<BS::thread_pool>;


cpp/include/kvikio/threadpool_wrapper.hpp
```diff
@@ -20,43 +20,35 @@
...
-   * each worker thread.
+   * @brief Construct a new thread pool wrapper, and invoke an initialization function in each
+   * worker thread.
    *
    * @param nthreads The number of threads to use.
+   * @param worker_thread_init_func Initialization function invoked in each worker thread.
    */
-  thread_pool_wrapper(unsigned int nthreads) : pool_type{nthreads, worker_thread_init_func}
+  thread_pool_wrapper(
+    unsigned int nthreads, std::function<void()> worker_thread_init_func = [] {})
```
kingcrimsontianyu: The BS thread pool does not check if the function object is empty, ~~which is considered a bug~~. As a workaround, here we have to let the default argument be a lambda `[] {}` instead of an empty function object `{}`.

cpp/include/kvikio/threadpool_wrapper.hpp
```diff
@@ -20,43 +20,35 @@
...
     pool_type::reset(nthreads, worker_thread_init_func);
   }
-
- private:
-  inline static std::function<void()> worker_thread_init_func{[] {
-    KVIKIO_NVTX_FUNC_RANGE();
-    // Rename the worker thread in the thread pool to improve clarity from nsys-ui.
-    // Note: This NVTX feature is currently not supported by nsys-ui.
-    NvtxManager::rename_current_thread("thread pool");
-  }};
```
kingcrimsontianyu, line 59: This is removed since thread renaming for a custom domain has never worked in `nsys-ui`, and on top of that, this static member makes it challenging to completely relieve the public headers of NVTX-related code.

cpp/include/kvikio/threadpool_wrapper.hpp
wence-: Can we report an issue upstream to fix this?

cpp/include/kvikio/threadpool_wrapper.hpp
wence-: Since we never previously exposed this initialisation function is there a reason to do so now?

cpp/include/kvikio/threadpool_wrapper.hpp
kingcrimsontianyu: Hmm. Pondered this a bit. The upstream library asks for a callable (`F&& f`). So it should be user's (my) responsibility to ensure that `std::function` object contains a valid target function.

cpp/include/kvikio/threadpool_wrapper.hpp
kingcrimsontianyu: My thought was that as a thread pool wrapper that in the future may support our own implementation, it would be a convenience to accept user-defined initialization function. May I know if there is any downside with this change in your mind?

cpp/include/kvikio/threadpool_wrapper.hpp
wence-: Ah, ok, makes sense.

cpp/include/kvikio/threadpool_wrapper.hpp
wence-: Maybe, but perhaps it's easier to defer creating that argument (which previously didn't exist) until such time as we find the need for it.

kingcrimsontianyu: Sure! Cleaned up.

PR #856, opened 2025-10-20
Use ruff and ruff-format instead of flake8, black, isort

KyleFromNVIDIA:
Issue: https://github.com/rapidsai/build-planning/issues/130

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,9 +16,6 @@ ignore = [
     "E501",
 ]
 fixable = ["ALL"]
-exclude = [
-    # "notebooks/*",
-]
 
 [tool.ruff.lint.per-file-ignores]
 "*.pyx" = [


.pre-commit-config.yaml
```diff
@@ -6,28 +6,12 @@ repos:
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.12.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.9.3
```
bdice: It'd be nice to adopt a newer release. The latest is v0.14.1. Is there anything blocking us from upgrading?

.pre-commit-config.yaml
```diff
@@ -6,28 +6,12 @@ repos:
...
-  - repo: https://github.com/PyCQA/flake8
-    rev: 7.1.1
-    hooks:
-      - id: flake8
-        args: ["--config=.flake8"]
-        files: python/.*$
-        types: [file]
-        types_or: [python, cython]
-        additional_dependencies: ["flake8-force"]
+      - id: ruff
```
bdice: If we can upgrade to a newer release, this should be updated to `ruff-check`.
```suggestion
      - id: ruff-check
```

.pre-commit-config.yaml
KyleFromNVIDIA: This config was copied from [cudf](https://github.com/rapidsai/cudf/blob/febc7ef3f1a6abcfdb9ddf12d52487bd21b284b2/.pre-commit-config.yaml#L143). I have no objection to upgrading, but I need to try it locally first.

KyleFromNVIDIA: Local update was successful. Just pushed it.

PR #859, opened 2025-10-26
Implement simple username-based authentication for WebHDFS

kingcrimsontianyu:
With this PR, KvikIO will support username-based authentication for WebHDFS via environment variable `KVIKIO_WEBHDFS_USERNAME`.

Note: `libcudf` uses KvikIO's utility function `open(url)` to infer endpoint type, where currently the access credentials can only be specified via environment variables instead of programmatically as function parameters. We will address this limitation in the future.

This PR is breaking in that:
- It moves S3 endpoint's utility function `unwrap_or_default` to the detailed namespace, considering that this utility function is supposed to be an implementation detail.
- It adds `username` parameter to one of the two WebHDFS endpoint constructors for completeness (the other constructor has already had `username` as its parameter).

### DIFF
--- a/cpp/include/kvikio/hdfs.hpp
+++ b/cpp/include/kvikio/hdfs.hpp
@@ -43,7 +43,7 @@ class WebHdfsEndpoint : public RemoteEndpoint {
    * @param host Host
    * @param port Port
    * @param remote_file_path Remote file path
-   * @param username Optional user name
+   * @param username Optional user name.
    *
    * @note The optional username for authentication is determined in the following descending
    * priority order:


cpp/include/kvikio/detail/env.hpp
```diff
@@ -0,0 +1,27 @@
...
+ * - If environment variable `env_var` is set, return its value (even if empty)
+ * - Return std::nullopt if err_msg is std::nullopt; throw an exception otherwise
+ *
+ * @param value The value to unwrap.
+ * @param env_var The name of the environment variable to check if `value` isn't set.
+ * @param err_msg Optional error message that controls whether to throw an exception if neither
+ * source provides a value.
+ * @return The resolved value, or std::nullopt if neither source provides a value.
+ */
+std::optional<std::string> unwrap_or_env(std::optional<std::string> value,
```
kingcrimsontianyu, line 24: Several small changes have been made compared to the initial version from the `kvikio` namespace,:
- `unwrap_or_default` changed to `unwrap_or_env`, because the function does not actually provide a default value but instead attempts to extract one from the environment variable.
- Return value changed from empty string to `std::nullopt`, because the former cannot differentiate the case of "unset environment variable" from "set-but-empty environment variable" (e.g. `export MY_ENV=`). In the latter case, `std::getenv()` does not return a `nullptr` but instead an empty string (with only a `NUL` character).

cpp/src/remote_handle.cpp
```diff
@@ -589,6 +575,7 @@ RemoteHandle RemoteHandle::open(std::string url,
                                 std::optional<std::vector<RemoteEndpointType>> allow_list,
                                 std::optional<std::size_t> nbytes)
 {
+  KVIKIO_NVTX_FUNC_RANGE();
```
kingcrimsontianyu, line 578: Curious about the endpoint type resolution overhead. Hence the NVTX annotation.

cpp/include/kvikio/hdfs.hpp
```diff
@@ -27,16 +27,28 @@ class WebHdfsEndpoint : public RemoteEndpoint {
    * @brief Create an WebHDFS endpoint from a url.
    *
    * @param url The WebHDFS HTTP/HTTPS url to the remote file.
+   * @param username Optional user name
```
madsbk:
```suggestion
   * @param username Optional user name.
```

cpp/include/kvikio/hdfs.hpp
```diff
@@ -27,16 +27,28 @@ class WebHdfsEndpoint : public RemoteEndpoint {
...
+  explicit WebHdfsEndpoint(std::string url, std::optional<std::string> username = std::nullopt);
 
   /**
    * @brief Create an WebHDFS endpoint from the host, port, file path and optionally username.
    *
    * @param host Host
    * @param port Port
    * @param remote_file_path Remote file path
-   * @param username User name
+   * @param username Optional user name
```
madsbk:
```suggestion
   * @param username Optional user name.
```

PR #860, opened 2025-10-28
Require CUDA 12.2+

jakirkham:
Make CUDA 12.2 the new minimum. Drops logic related to CUDA 12.0 & 12.1. Thus simplifying the CUDA build and dependency logic.

<hr>

Part of issue:
* https://github.com/rapidsai/build-planning/issues/223


java/pom.xml
```diff
@@ -24,7 +24,7 @@ SPDX-License-Identifier: Apache-2.0
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
     <junit.version>5.4.2</junit.version>
-    <jcuda.version>12.0.0</jcuda.version>
+    <jcuda.version>12.2.2</jcuda.version>
```
jakirkham: Cannot tell if this or other Java files will need changes. Would be great to hear feedback from Java developers

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -5,12 +5,6 @@ schema_version: 1
...
   version: ${{ env.get("RAPIDS_PACKAGE_VERSION") }}
   minor_version: ${{ (version | split("."))[:2] | join(".") }}
-  # We need to support three cases:
-  # 1. Linux x86_64, which always uses libcufile
-  # 2. Linux aarch64 with CUDA >= 12.2, which uses libcufile
-  # 3. Linux aarch64 with CUDA < 12.2, which does not use libcufile
-  # Each case has different cuda-version constraints as expressed below
-  should_use_cufile: ${{ x86_64 or (aarch64 and cuda_version >= "12.2") }}
   # When reverting, instances of cuda_key_string can be replaced with cuda_major
   cuda_key_string: ${{ cuda_version | replace(".", "_") }}
```
bdice: I think we want this to go back to `cuda_major` as the comment suggests. Delete `cuda_key_string`, uncomment `cuda_version` / `cuda_major`, and replace `cuda_key_string` with `cuda_major`.

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -5,12 +5,6 @@ schema_version: 1
 context:
   version: ${{ env.get("RAPIDS_PACKAGE_VERSION") }}
   minor_version: ${{ (version | split("."))[:2] | join(".") }}
```
bdice: I think this is unused.

cpp/src/shim/cufile.cpp
```diff
@@ -38,13 +38,14 @@ cuFileAPI::cuFileAPI()
...
 
   // Some symbols were introduced in later versions, so version guards are required.
-  // Note: `version` is 0 for cuFile versions prior to v1.8 because `cuFileGetVersion`
-  // did not exist. As a result, the batch and stream APIs are not loaded in versions
-  // 1.6 and 1.7, respectively, even though they are available. This trade-off is made
-  // for improved robustness.
+  // Note: `version` is v1.7 (in CUDA 12.2) for cuFile versions prior to v1.8 (CUDA 12.3)
+  // because `cuFileGetVersion` did not exist until CUDA 12.3 and CUDA 12.2 is the oldest
+  // version we support. This trade-off is made for improved robustness.
   if (version >= 1060) {
```
jakirkham: CUDA 12.2.0 shipped with cuFile 1.7.0.49

In other words it had the batch and async features all supported. Given this all of the symbols checked for below are defined and the checks could be dropped. For now have just bumped the `version` value to represent 1.7.0

Refs:
* https://docs.nvidia.com/cuda/archive/12.2.0/cuda-toolkit-release-notes/index.html#cuda-toolkit-major-component-versions
* https://docs.nvidia.com/gpudirect-storage/release-notes/index.html#new-features-and-changes

conda/recipes/libkvikio/recipe.yaml
jakirkham: That appears to be true. Though that also appears to be true before this PR

@gforsyth it appears you added `minor_version` in PR ( https://github.com/rapidsai/kvikio/pull/678 ) when converting this to the v1 recipe format. However even that PR does not appear to use `minor_version`. Is there some context we are missing about what `minor_version` is being used for? Or can it be dropped?

conda/recipes/libkvikio/recipe.yaml
jakirkham: Am ambivalent here. On the one hand can see that point. However on the other it could be useful to call out CUDA 12.2 is a minimum here

java/pom.xml
jakirkham: Looks like [`jcuda` on Maven]( https://repo.maven.apache.org/maven2/org/jcuda/jcuda/ ) can only be 12.0.0 or 12.6.0. Please let us know if you have thoughts on that and any other Java changes that would be needed

madsbk: Maybe we can just remove the Java bindings? https://github.com/rapidsai/kvikio/issues/861

conda/recipes/libkvikio/recipe.yaml
bdice: We've been dropping unused variables in this section, this is probably just an oversight. Let's fix it.
```suggestion
```

conda/recipes/libkvikio/recipe.yaml
bdice: I don't understand how this helps. We already set a lower bound on `cuda-version`. Let's stop using special-case behavior here and revert as the comment recommends.

java/pom.xml
jakirkham: Good question

Will also reach out offline

If we don't hear much response, will roll back the Java change in this PR

Would like to keep the scope of this PR to bumping CUDA 12.0 to 12.2

conda/recipes/libkvikio/recipe.yaml
jakirkham: The same argument could be made for the explicit inclusion of the CUDA version in our build string currently

It is also already handled via `cuda-version` and the build string hash

Think it is worth discussing this point

conda/recipes/libkvikio/recipe.yaml
jakirkham: Your point is noted

Would like to hear back from Gil before proceeding

gforsyth: It should be fine to drop it.  I don't recall adding it for any specific reason, so it was likely a copy-paste from another port.

jakirkham: Thanks Gil! 🙏

Looks like `minor_version` shows up in both recipes and is unused in both. So have gone ahead and dropped it from both: https://github.com/rapidsai/kvikio/pull/860/commits/119338ab34d5d101ed731c207a47f83ccedc5929

java/pom.xml
jakirkham: @aslobodaNV any thoughts on what we should be doing with the Java code here or elsewhere when dropping CUDA 12.0 & 12.1?

conda/recipes/libkvikio/recipe.yaml
vyasr: > The same argument could be made for the explicit inclusion of the CUDA version in our build string currently

How? conda doesn't allow you to upload two builds that have identical versions and build strings but different dependencies. You have to include some bit of information that makes the build unique. I guess we could use PKG_HASH to distinguish the builds in the build string, but if you have to include something it might as well be more meaningful than that.

That being said, I don't look at the build string as a useful way of conveying the minimum supported CUDA version given that the cuda-version dependency is part of the package metadata as well, so my vote would be to go back to the previous status. 12 vs 13 tells me enough about the purpose of the two packages. I'm not looking for anything more than that.

java/pom.xml
jakirkham: Going to revert this change

```suggestion
    <jcuda.version>12.0.0</jcuda.version>
```

Will follow along on the discussion in issue: https://github.com/rapidsai/kvikio/issues/861

conda/recipes/libkvikio/recipe.yaml
jakirkham: So that argument goes like this

If `cuda-version` is in the variants (like `conda_build_config.yaml` or `variants.yaml`), then it is included in the string hash. So there is no need to manually add it to the string

The challenge with including it manually (as we have to date) is it means writing the build string from scratch each time. As a result we end up missing details that would have otherwise been captured for us if it was added as a variant and included in the string hash

<hr>

Right so there are essentially two viewpoints we encounter. One view is the as you have articulated that the build string is not a reliable or consistent way to find this information so it shouldn't be relied on and package dependencies (like `cuda-version`) should be used to do that. The other view is that it can be useful to have the relevant info in one place on the relevant package to simplify the end-user workflow

Given today we already recommend users specify `cuda-version`, there is no real simplification offered by adding this info to the build string. So would tend to agree with the first view for our case (similar to you). And would go further, that it would be better to minimize string customization and rely on variants and the string hash

vyasr: The reason we have never included the CUDA version in the variants directly is that we need the flexibility of building different variants in different CI jobs on different runners, whereas conda-build and rattler-build will build all the variants at once (the slowdown in end-to-end CI time that would result from serializing the builds has too big a negative impact on developers). With the switch to rattler-build we have a bit of additional flexibility because rattler-build supports providing the variant config on the CLI, but ultimately that means we'd be constructing the variants on the fly in CI jobs which leads to more divergence between local dev and CI where we have been aiming to close the gap over time. If you have a solution that allows us to specify all the information in the variants file and avoid environment and recipe context variables, I'm all for it. Otherwise I think we're stuck with some degree of manual string customization.

jakirkham: Both [conda-build]( https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#conda-variables-and-command-line-arguments-to-conda-build ) and [rattler-build]( https://rattler.build/latest/variants/#overriding-the-variant-configuration-from-the-command-line ) can specify variants by CLI

Think we are conflating two things:

1. How the matrix is specified
2. How CUDA version is specified

We need not change 1 to change 2. Am focused here on 2

vyasr: If you have an idea of how to implement a proposed improved, feel free to push a commit with the change and I'm happy to review.

jakirkham: I do. Though will approach this in a follow up PR so we can discuss. We can then decide if it merits broader rollout

For now have simply reverted the relevant bits here

cpp/src/shim/cufile.cpp
kingcrimsontianyu: Thanks for the clean-up! This will help to simplify the logic in C++ compatibility mode manager as well.

jakirkham: Thanks for the encouragement Tianyu! 🙏

Went ahead and dropped all of this logic

cpp/CMakeLists.txt
kingcrimsontianyu, line 85: Should this be removed too?

kingcrimsontianyu, line 102: ... and this one as well

jakirkham: Was thinking about these. I think we may still want to check, but instead error if it doesn't work

What do you think?

jakirkham: Let's discuss in the above thread: https://github.com/rapidsai/kvikio/pull/860#discussion_r2500697656

We can apply any changes needed to both after

cpp/src/compat_mode_manager.cpp
```diff
@@ -74,22 +74,18 @@ CompatModeManager::CompatModeManager(std::string const& file_path,
...
-    if (!is_stream_api_available()) { err_msg += "Missing the cuFile stream api."; }
-
-    // When checking for availability, we also check if cuFile's config file exists. This is
+    // When checking for availability, we check if cuFile's config file exists. This is
     // because even when the stream API is available, it doesn't work if no config file exists.
-    if (config_path().empty()) { err_msg += " Missing cuFile configuration file."; }
+    if (config_path().empty()) { err_msg += "Missing cuFile configuration file."; }
 
     KVIKIO_FAIL(err_msg, std::runtime_error);
   }
```
kingcrimsontianyu, line 96:
```suggestion
void CompatModeManager::validate_compat_mode_for_async() const
{
  KVIKIO_NVTX_FUNC_RANGE();
  // cuFile stream API will not work if the config file does not exist.
  if (_is_compat_mode_preferred_for_async && _compat_mode_requested == CompatMode::OFF &&
      config_path().empty()) {
    KVIKIO_FAIL("Missing cuFile configuration file.", std::runtime_error);
  }
}
```

cpp/src/shim/cufile.cpp
```diff
@@ -37,27 +37,25 @@ cuFileAPI::cuFileAPI()
     CUfileError_t const error = GetVersion(&ver);
     if (error.err == CU_FILE_SUCCESS) { version = ver; }
   } catch (std::runtime_error const&) {
+    version = 1070;
   }
+#else
+  version = 1070;
 #endif
```
kingcrimsontianyu, line 44: To be consistent with previous behavior, we don't need to set the version number here in the else branch. It will take the default value of 0 (specified in `cuFileAPI`'s constructor). I think we could add back some of the comments related to the version checking API. So I'd suggest this change:
```suggestion
  // Note: `version` is 0 for cuFile versions prior to v1.8 because `cuFileGetVersion`
  // did not exist.
#ifdef KVIKIO_CUFILE_VERSION_API_FOUND
  try {
    get_symbol(GetVersion, lib, KVIKIO_STRINGIFY(cuFileGetVersion));
    int ver;
    CUfileError_t const error = GetVersion(&ver);
    if (error.err == CU_FILE_SUCCESS) { version = ver; }
  } catch (std::runtime_error const&) {
  }
#endif
```

cpp/CMakeLists.txt
kingcrimsontianyu: Since the batch/stream APIs are guaranteed to exist in cuFile under CTK 12.2+, I think we can remove the check here for simplicity.

cpp/src/compat_mode_manager.cpp
kingcrimsontianyu: On a side note, I didn't know the presence of config file is required for the stream API to function properly. This is quite interesting. I'll run some tests and see if this behavior has changed in recent cuFile API.

kingcrimsontianyu (changes requested):
Suggested a couple of changes. Thanks!

cpp/src/compat_mode_manager.cpp
jakirkham: Included a variant of this cleanup

cpp/src/shim/cufile.cpp
jakirkham: CUDA 12.2.0 included cuFile 1.7.0.49

Setting version to 0 doesn't make sense any more as that would imply disabling batch and stream APIs, which no longer can happen after these changes

Refs:
* https://docs.nvidia.com/cuda/archive/12.2.0/cuda-toolkit-release-notes/index.html#cuda-toolkit-major-component-versions
* https://docs.nvidia.com/gpudirect-storage/release-notes/index.html#new-features-and-changes

kingcrimsontianyu: Previously for cuFile v1.7 where cuFile batch/stream API is available but `cuFileGetVersion` is not, `version` has a value of 0, indicating the missing of the version query API.
Since v1.7 is now the only possible version that does not have the `cuFileGetVersion` symbol, it is correct to have a default value of `1070`.

cpp/CMakeLists.txt
jakirkham: Converted these to build errors if they are missing

cpp/src/shim/cufile.cpp
jakirkham: Ok great. Glad we agree. Will mark this as resolved then

jameslamb (approved):
No suggestions, this looks good to me. Thanks!

Very glad we could get this simplification in 😁

kingcrimsontianyu (approved):
Approved the C++ part. Since the merge conflict occurs to the yaml file, I hope @bdice or @vyasr could give some additional scrutiny.

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -62,7 +53,7 @@ cache:
         SCCACHE_NO_CACHE: ${{ env.get("SCCACHE_NO_CACHE", default="") }}
         SCCACHE_RECACHE: ${{ env.get("SCCACHE_RECACHE", default="") }}
         SCCACHE_REGION: ${{ env.get("SCCACHE_REGION", default="") }}
-        SCCACHE_S3_KEY_PREFIX: libkvikio/${{ env.get("RAPIDS_CONDA_ARCH") }}/cuda${{ cuda_key_string }}
+        SCCACHE_S3_KEY_PREFIX: libkvikio/${{ env.get("RAPIDS_CONDA_ARCH") }}/cuda${{ cuda_major }}
```
jakirkham, line 56: The merge conflict only occurred because we change `cuda_key_string` to `cuda_major` on this line

Upstream had other changes to the environment variables, which we just pulled in

Issue #761, opened 2025-06-26
[Feature] Add option to pass `mode` option to python Kvikio.Cufile for when bypassing CuFile for POSIX directly.


fstrug:
It would be useful to add the ability for users to define when in compatibility mode whether the python `Kvikio.Cufile` read API uses the O_DIRECT flag or not with POSIX. For users wanting to evaluate the performance between kvikio's CuFile and POSIX functionality caching might effect performance - especially when trying to evaluate performance on an average of multiple runs ran on the same filepath. This will give users greater control and confidence in their benchmark results.

![Image](https://github.com/user-attachments/assets/e3d14205-f4fa-4813-a4ec-569dc522ffb7)
From https://developer.nvidia.com/blog/gpudirect-storage/?ncid=no-ncid

Currently "when in compatibility mode, FileHandle only opens the file once (without the O_DIRECT flag)". https://github.com/rapidsai/kvikio/pull/410#issue-2424383037

kingcrimsontianyu:
Thanks for raising the issue. Support for `O_DIRECT` for the compatibility mode is being considered. Implementation is likely to take non-trivial code changes: `O_DIRECT` usually imposes stringent alignment requirements where the user-provided buffer, I/O segment length, and file offset have to be aligned to a platform-dependent value. Granted, for simplicity we may choose to overalign everything to the page size, we still need to cover all the cases where one or more of the 3 components mentioned above are unaligned.

If your purpose is to benchmark, I'd recommend using the new [kvikio.cufile.clear_page_cache](https://docs.rapids.ai/api/kvikio/nightly/api/#kvikio.cufile.clear_page_cache) feature available in 25.08 nightly. Calling this function prior to each run of your benchmark program is expected to eliminate the impact of kernel page cache. If 25.08 nightly is not a viable option, then use these Linux commands that have the same cache clearing effect with the new API:
```
sync
sudo /sbin/sysctl vm.drop_caches=3
./my_benchmark_program
```

TomAugspurger:
kvikio-cu12 is at https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/kvikio-cu12. I think the index page at https://pypi.anaconda.org/rapidsai-wheels-nightly/simple comes from anaconda.org's implementation of https://packaging.python.org/en/latest/specifications/simple-repository-api/. I'm not sure if it missing some packages from the HTML page is expected, but they are there.

kingcrimsontianyu:
[...]

Issue #520, opened 2024-10-28
Double buffed bounce buffer


madsbk:
We should investigate the performance benefits of implementing double buffering when using bounce buffers in POSIX and Remote IO.

PR #863, opened 2025-10-31
Add opportunistic Direct I/O support to POSIX backend

kingcrimsontianyu:
This PR significantly improves POSIX I/O write performance as well as cold-page-cache read by opportunistically using Direct I/O. The speedup for sequential write is approximately 3~4x.

The opportunistic POSIX Direct I/O feature can be controlled in two ways:
- Environment variables:
  - `"KVIKIO_AUTO_DIRECT_IO_READ"`: defaults to `false`.
  - `"KVIKIO_AUTO_DIRECT_IO_WRITE"`: defaults to `true`.

- C++/Python API
  - `defaults::set_auto_direct_io_read(bool flag)`/`kvikio.defaults.set("posix_auto_direct_io_read", flag)`
  - `defaults::set_auto_direct_io_write(bool flag)`/`kvikio.defaults.set("posix_auto_direct_io_write", flag)`

In addition, this PR refactors the bounce buffer class. To improve clarity, relevant classes and variables have been renamed and a lot of comments added. The bounce buffer class is now templated by allocator to accommodate different use cases:
- `PageAlignedBounceBufferPool`: used for Direct I/O to/from unaligned host buffer. Does not require CUDA context.
- `CudaPinnedBounceBufferPool`: used for buffered I/O to/from device buffer. Requires CUDA context. This is the original implementation on the main branch.
- `CudaPageAlignedPinnedBounceBufferPool`: used for Direct I/O to/from device buffer. Requires CUDA context.

## Performance results
See https://github.com/rapidsai/kvikio/pull/863#issuecomment-3479016816

## Goal
- Addresses most part of https://github.com/rapidsai/kvikio/issues/761
- Addresses the reported write performance issue in cudf

## Non-goal
- This PR does not add opportunistic Direct I/O as file handle's function parameters. This will be revisited in a future PR.
- This PR does not address one of the objectives in #520, which is to unify the implementation of the bounce buffer in POSIX IO and in Remote IO. This will be revisited in a future PR.

### DIFF
--- a/cpp/include/kvikio/detail/posix_io.hpp
+++ b/cpp/include/kvikio/detail/posix_io.hpp
@@ -224,8 +224,9 @@ std::size_t posix_device_io(int fd_direct_off,
 {
   // Direct I/O requires page-aligned bounce buffers. CudaPinnedBounceBufferPool uses
   // cudaMemHostAlloc which does not guarantee page alignment.
-  if (fd_direct_on != -1 && std::is_same_v<BounceBufferPoolType, CudaPinnedBounceBufferPool>) {
-    KVIKIO_FAIL(
+  if (std::is_same_v<BounceBufferPoolType, CudaPinnedBounceBufferPool>) {
+    KVIKIO_EXPECT(
+      fd_direct_on == -1,
       "Direct I/O requires page-aligned bounce buffers. CudaPinnedBounceBufferPool does not "
       "guarantee page alignment. Use CudaPageAlignedPinnedBounceBufferPool instead.");
   }


kingcrimsontianyu:
# Performance results

## System setup
- 1x NVIDIA L40 GPU
- 2x Micron_9300_MTFDHAL6T4TDR as RAID0
  - Sequential Read : 3500 Mbps
  - Sequential Write : 3500 Mbps
  - Random Read : 850k Iops
  - Random Write : 310k Iops

## KvikIO microbenchmark
- Sequential write
  - Setting
    - KvikIO compatibility mode (no GDS)
    - 8 GiB binary data
    - File open/close per iteration
 
  - Results
 
  |  | Thread count | Main [MiB/s] | This PR [MiB/s] |
  | ------------- | ------------- | ------------- | ------------- |
  | host to file  | 1  | 1493.0335 | 4161.8697|
  | device to file | 1 | 1387.2123  | 4566.8671 |
  | host to file  | 4  | 1094.1003 | 5027.2060 |
  | device to file  | 4 | 1321.4875  | 5188.5864 |

    - This PR significantly improves sequential write performance
    - Whether to drop page cache does not affect sequential write performance


## cuDF parquet writer benchmark
  - Setting
    - KvikIO compatibility mode (no GDS)
    - 4 threads by default
    - Non-profile uses `--min-samples 40 --timeout 60`
    - Profile uses `--min-samples 10 --timeout 10`
    - Command line:
```
PARQUET_WRITER_NVBENCH \
-d 0 \
-b parquet_write_io_compression \
-a compression_type=SNAPPY \
-a io_type=FILEPATH \
-a cardinality=0 \
-a run_length=1 \
--min-samples 40 --timeout 60 --csv stdout
```

  - Results

  |  | Thread count | Main [s] | This PR [s] |
  | ------------- | ------------- | ------------- | ------------- |
  | total benchmark wall time | 4  | 1.172820521125 | 0.7172160462023811 |
  | total pwrite time across all threads | 4  | 13.646 | 4.661 |

kingcrimsontianyu:
# Documentation update

image

docs/source/runtime_settings.rst
```diff
@@ -17,42 +17,42 @@ Under ``AUTO``, KvikIO falls back to the compatibility mode:
...
 ------------------------------
 KvikIO splits parallel IO operations into multiple tasks. Set the environment variable ``KVIKIO_TASK_SIZE`` to the maximum task size (in bytes). If not set, the default value is 4194304 (4 MiB).
 
-This setting can also be accessed using :py:func:`kvikio.defaults.task_size` (getter) and :py:func:`kvikio.defaults.set` (setter).
+This setting can also be accessed using :py:func:`kvikio.defaults.get` (getter) and :py:func:`kvikio.defaults.set` (setter).
 
 GDS Threshold ``KVIKIO_GDS_THRESHOLD``
 --------------------------------------
-In order to improve performance of small IO, ``.pread()`` and ``.pwrite()`` implement a shortcut that circumvent the threadpool and use the POSIX backend directly. Set the environment variable ``KVIKIO_GDS_THRESHOLD`` to the minimum size (in bytes) to use GDS. If not set, the default value is 1048576 (1 MiB).
+In order to improve performance of small IO, ``.pread()`` and ``.pwrite()`` implement a shortcut that circumvent the threadpool and use the POSIX backend directly. Set the environment variable ``KVIKIO_GDS_THRESHOLD`` to the minimum size (in bytes) to use GDS. If not set, the default value is 16384 (16 KiB).
```
kingcrimsontianyu, line 36: Fixed a doc error

python/kvikio/kvikio/_lib/buffer.pyx
```diff
@@ -30,11 +30,12 @@ def memory_deregister(buf) -> None:
 
 
 cdef extern from "<kvikio/bounce_buffer.hpp>" nogil:
-    size_t cpp_alloc_retain_clear "kvikio::AllocRetain::instance().clear"() except +
+    size_t cpp_cuda_pinned_bounce_buffer_pool_clear \
+        "kvikio::CudaPinnedBounceBufferPool::instance().clear"() except +
```
kingcrimsontianyu, line 37: ~~Here `CudaPinnedBounceBufferPool` is chosen since it is the same with the previous version of bounce buffer. But how about the other two new types? Should we add Python API for them too? Or should we hide the bounce buffer from the public API and move it to the detail namespace? I'm not quite sure atm.~~

To fix the unit test failure, I have made the Python API `bounce_buffer_free()` clear all 3 bounce buffer pools.

cpp/include/kvikio/bounce_buffer.hpp
```diff
@@ -11,48 +11,153 @@
...
+ * @brief Allocator for page-aligned AND CUDA-registered pinned host memory
+ *
+ * Combines the benefits of both page alignment (for Direct I/O) and CUDA registration
+ * (for efficient host-device transfers). Uses std::aligned_alloc followed by
+ * cudaMemHostRegister to achieve both properties.
+ *
+ * @note This is the required allocator for Direct I/O with device memory. Requires a valid CUDA
+ * context when allocating.
+ */
+class CudaPageAlignedPinnedAllocator {
```
vuule, line 77: This combinatorial set of types reminds of the design from "Modern C++ Design", where each of these options would be a separate template parameter (or something like that). Probably not required here since there are only three types, but wanted to leave a note for (future) consideration.

cpp/include/kvikio/detail/utils.hpp
```diff
@@ -44,4 +44,22 @@ namespace kvikio::detail {
...
 [[nodiscard]] void* align_down(void* addr, std::size_t alignment) noexcept;
 
+/**
+ * @brief Whether `value` is a multiple of `alignment`
+ *
+ * @param value Value to be checked
+ * @param alignment Must be a power of 2
+ * @return Boolean answer
+ */
+bool is_aligned(std::size_t value, std::size_t alignment) noexcept;
```
vuule: Given the context in which these are used, they don't benefit from being `noexcept`. I think it would be better to have a `KVIKIO_EXPECT `check and remove `noexcept`.

cpp/tests/test_basic_io.cpp
```diff
@@ -86,3 +97,130 @@ TEST_F(BasicIOTest, write_read_async)
...
+ public:
+  using value_type = std::int64_t;
+
+ protected:
+  void SetUp() override
+  {
+    TempDir tmp_dir{false};
+    _filepath = tmp_dir.path() / "test";
+
+    // Skip the fixture is Direct I/O is not supported
```
vuule:
```suggestion
    // Skip the fixture if Direct I/O is not supported
```

cpp/tests/test_basic_io.cpp
```diff
@@ -86,3 +97,130 @@ TEST_F(BasicIOTest, write_read_async)
...
+    try {
+      [[maybe_unused]] auto fd =
+        kvikio::open_fd(_filepath.c_str(), "w", true /* o_direct */, kvikio::FileHandle::m644);
+    } catch (...) {
+      GTEST_SKIP() << "Direct I/O is not supported for the test file: " << _filepath;
+    }
+
+    // Create a sequence of numbers as a ground truth
+    _num_elements = 8ULL * 1024ULL * 1024ULL + 1234ULL;
```
vuule: this is almost 40GB of data, right?
How long does it take to run these tests (including the fixture init)?

cpp/tests/utils/utils.hpp
```diff
@@ -178,4 +178,65 @@ inline void expect_equal(DevBuffer<T> const& a, DevBuffer<T> const& b)
...
+
+  template <class U>
+  struct rebind {
+    using other = CustomHostAllocator<U, ali, element_offset>;
+  };
+
+  [[nodiscard]] T* allocate(std::size_t num_elements)
+  {
+    if (num_elements > std::numeric_limits<std::size_t>::max() / sizeof(T))
+      throw std::bad_array_new_length();
```
vuule, line 216:
```suggestion
      {throw std::bad_array_new_length();}
```

cpp/src/detail/posix_io.cpp
```diff
@@ -46,26 +46,40 @@ CUstream StreamsByThread::get()
...
-                              std::size_t devPtr_offset)
+                              std::size_t devPtr_offset,
+                              int fd_direct_on)
 {
   KVIKIO_NVTX_FUNC_RANGE(size);
-  return detail::posix_device_io<IOOperationType::READ>(
-    fd, devPtr_base, size, file_offset, devPtr_offset);
+  // If Direct I/O is supported and requested
+  if (fd_direct_on != -1 && defaults::auto_direct_io_read()) {
+    return detail::posix_device_io<IOOperationType::READ, CudaPageAlignedPinnedBounceBufferPool>(
```
vuule, line 59: could we always use page aligned bounce buffers?

cpp/include/kvikio/detail/posix_io.hpp
```diff
@@ -183,20 +296,29 @@ std::size_t posix_host_read(int fd, void* buf, std::size_t size, std::size_t fil
...
+ * @param fd_direct_off File descriptor without Direct I/O.
  * @param buf Base address of buffer in host memory.
  * @param size Size in bytes to write.
  * @param file_offset Offset in the file to write to.
+ * @param fd_direct_on Optional file descriptor with Direct I/O.
  * @return Size of bytes that were successfully read.
  */
 template <PartialIO PartialIOStatus>
-std::size_t posix_host_write(int fd, void const* buf, std::size_t size, std::size_t file_offset)
+std::size_t posix_host_write(int fd_direct_off,
```
vuule, line 310: can a user pass the same descriptor (with o_direct) for both parameters?

cpp/src/file_handle.cpp
```diff
@@ -323,4 +324,6 @@ const CompatModeManager& FileHandle::get_compat_mode_manager() const noexcept
   return _compat_mode_manager;
 }
 
+bool FileHandle::direct_io_suppored() const noexcept { return _file_direct_on.fd() != -1; }
```
vuule:
```suggestion
bool FileHandle::is_direct_io_supported() const noexcept { return _file_direct_on.fd() != -1; }
```

kingcrimsontianyu: Thanks. Done. Also added this method to Python API.

cpp/tests/test_basic_io.cpp
kingcrimsontianyu: Per file, it's a little above 64 MiB :smiley:  On my PC it tooks <5 sec to complete `BASIC_IO_TEST`. I wanted this number to be above the default 4 MiB task size and also not an exact multiple. I've tuned this down further to a little above 8 MiB. It takes < 2 sec now.

cpp/tests/utils/utils.hpp
kingcrimsontianyu: Done.

cpp/src/detail/posix_io.cpp
kingcrimsontianyu: I thought about this. What I was unsure of is whether the page-aligned host buffer + CUDA register would have the same performance (the API call itself as well as the subsequent H<->D copies) with the CUDA host alloc-ed buffer. I remember there was some issues with CUDA register on GH200 in the past, so the old non-page aligned buffer is used here to play safe.
That being said, this Direct I/O feature hasn't been tested on GH200 yet and I'll leave it as a future work.

cpp/include/kvikio/detail/posix_io.hpp
kingcrimsontianyu: Does "both parameters" refer to with/without O_DIRECT? A file descriptor has to be either using the O_DIRECT flag or not, not both. Let me know if you meant something else.

cpp/tests/test_basic_io.cpp
kingcrimsontianyu: Done. Thanks!

cpp/tests/test_basic_io.cpp
vuule: 5s for the whole suite seems fine.

cpp/include/kvikio/detail/posix_io.hpp
vuule: Is it valid to call `posix_host_write` and pass the same descriptor for both `fd_direct_off` and `fd_direct_on`, if this descriptor is using the `O_DIRECT` flag? As in, there's no requirement for `fd_direct_off` to not use O_DIRECT?

kingcrimsontianyu: That's correct. And if this happens, then at runtime, the I/O arguments (buffer address, segment length, file offset) passed to the syscall `pread` that uses `fd_direct_off` (with `O_DIRECT`) have to be properly aligned. Otherwise, the syscall will fail with invalid argument errors.

`posix_host_write/read` is KvikIO's internal function, and `fd_direct_on` and `fd_direct_off` are initialized in `kvikio::FileHandle`'s constructor. So regular users do not have to be aware of the fact above. Advanced users may take advantage of [int fd(bool o_direct)](https://github.com/rapidsai/kvikio/blob/main/cpp/include/kvikio/file_handle.hpp#L110) being part of the public API and enforce Direct I/O all the time. So if they manipulate `fd_direct_off` the following way, they have to ensure satisfying the alignment requirement.
```cpp
kvikio::FileHandle file("test.bin", "r");

// No alignment restriction
auto fut = kvikio::pread(buf, segment_length, file_offset);
fut.get();

// Add O_DIRECT flag to the fd
auto flags = fcntl(fd, F_GETFL);
fcntl(file.fd(), F_SETFL, flags | O_DIRECT);

// Now the args must meet alignment requirement
auto fut = kvikio::pread(buf, segment_length, file_offset);
fut.get();
```

vuule: Got it, thanks for the detailed explanation. I was thinking that `fd_direct_off` could just be `fd`, but the fact that pread calls can fail does make the current name more accurate.

cpp/include/kvikio/detail/posix_io.hpp
```diff
@@ -66,91 +68,196 @@ class StreamsByThread {
...
+                            int fd_direct_on = -1)
 {
-  auto alloc              = AllocRetain::instance().get();
+  // Direct I/O requires page-aligned bounce buffers. CudaPinnedBounceBufferPool uses
+  // cudaMemHostAlloc which does not guarantee page alignment.
+  if (fd_direct_on != -1 && std::is_same_v<BounceBufferPoolType, CudaPinnedBounceBufferPool>) {
+    KVIKIO_FAIL(
+      "Direct I/O requires page-aligned bounce buffers. CudaPinnedBounceBufferPool does not "
+      "guarantee page alignment. Use CudaPageAlignedPinnedBounceBufferPool instead.");
+  }
```
vuule: I think this reads better and is, in theory, more efficient
```suggestion
  if constexpr(std::is_same_v<BounceBufferPoolType, CudaPinnedBounceBufferPool>) {
    KVIKIO_EXPECTS(fd_direct_on == -1, "Direct I/O requires page-aligned bounce buffers. CudaPinnedBounceBufferPool does not "
      "guarantee page alignment. Use CudaPageAlignedPinnedBounceBufferPool instead.");
  }
```

kingcrimsontianyu: Agreed. Thanks!

PR #874, opened 2025-11-19
Add C++ benchmarks: FileHandle sequential read (part 2/n)

kingcrimsontianyu:
Partially addresses https://github.com/rapidsai/kvikio/issues/606

Issue #850, opened 2025-10-08
[Discussion] What’s the appropriate KVIKIO_NTHREADS setting for GDS?


JigaoLuo:
Hi KVIKIO team,

I’d like to share a performance debugging story from this afternoon that I think is somewhat unique, and hopefully useful to others.

Let me start by describing the setup: 
- hardware: a DGX A100 system with 4 PCIe4.0 SSDs, each capable of ~6.4 GB/s bandwidth, measured by GDSIO. 
  - Previously, these SSDs were configured as a single RAID-0 volume. 
  - This week, we split the RAID and mounted each SSD as an independent filesystem.
- my workload involves multistream, multithreaded Parquet reading, which is **a strong assumption**: I have implemented a multistream, multithreaded Parquet reader based on libcudf. In my design, each Parquet read call retrieves only a small chunk (such as a row group like 32MB), and multiple worker threads concurrently issue these Parquet read calls. This approach differs from traditional readers, which may process larger chunks.

My Parquet reader setup is a bit different—I operate at the row group level, which means I typically launch around 60 separate cudf `read_parquet` calls per file.

## With RAID-0

Life was simple with RAID-0. I could easily saturate the SSD bandwidth:
- GDSIO: Using 4 threads and large read chunks 
- In my parquet reader: with KVIKIO_NTHREADS=16 and 8 worker threads issuing cudf read_parquet calls

Performance was excellent—reading Parquet at ~23 GB/s into one A100. Decoding and decompression were pipelined, and the overhead from 16 KVIKIO threads plus 8 cudf workers was kind of minimal. Everything worked beautifully thanks to KVIKIO and GDS.

## Without RAID

For various reasons, we decided to un-RAID the SSDs and mount them individually. I reran my previous setup:
- With GDSIO, I could still saturate a single SSD (~6.4 GB/s) using 4 threads. 
  - By issuing 4 concurrent GDSIO commands—each targeting a different SSD and using 4 threads—I again reached ~23 GB/s. That was expected.
- Using my Parquet reader with KVIKIO_NTHREADS=16 and 8 cudf worker threads, I could read from a single SSD at ~6.5 GB/s. Also expected.

So far, everything looked good.

### Performance hunting

[...]

PR #876, opened 2025-11-20
Add thread pool as function parameters to C++ API

kingcrimsontianyu:
This PR introduces two changes in KvikIO C++ API:
- For the `pread` (and `pwrite` if applicable) method in `FileHandle`, `RemoteHandle` and `MmapHandle`'s , this PR adds the thread pool as a function parameter. By default, the global thread pool is used.
- After previous PR's cleanup (https://github.com/rapidsai/kvikio/pull/851/files/30543a3ccb953b0eb2afb7bb91d36ceda482dd69#r2432486309), the `thread_pool_wrapper` class merely forwards calls without adding useful functionality. This PR removes it and adds a simple type alias `ThreadPool` for the underlying `BS:thread_pool`.

This PR is a dependency of https://github.com/rapidsai/kvikio/pull/874 which facilitates investigation into a multi-drive scaling problem reported by https://github.com/rapidsai/kvikio/issues/850.

### DIFF
--- a/cpp/include/kvikio/mmap.hpp
+++ b/cpp/include/kvikio/mmap.hpp
@@ -164,15 +164,17 @@ class MmapHandle {
    * @param offset File offset
    * @param task_size Size of each task in bytes
    * @param thread_pool Thread pool to use for parallel execution. Defaults to the global default
-   * thread pool.
+   * thread pool. The caller is responsible for ensuring that the thread pool remains valid until
+   * the returned future is consumed (i.e., until `get()` or `wait()` is called on it).
    * @return Future that on completion returns the size of bytes that were successfully read.
    *
    * @exception std::out_of_range if the read region specified by `offset` and `size` is
    * outside the initial region specified when the mapping handle was constructed
    * @exception std::runtime_error if the mapping handle is closed
    *
-   * @note The `std::future` object's `wait()` or `get()` should not be called after the lifetime of
-   * the MmapHandle object ends. Otherwise, the behavior is undefined.
+   * @note The returned `std::future` object must not outlive either the MmapHandle or the thread
+   * pool. Calling `wait()` or `get()` on the future after the MmapHandle or thread pool has been
+   * destroyed results in undefined behavior.
    */
   std::future<std::size_t> pread(void* buf,
                                  std::optional<std::size_t> size = std::nullopt,
--- a/cpp/include/kvikio/file_handle.hpp
+++ b/cpp/include/kvikio/file_handle.hpp
@@ -230,11 +230,13 @@ class FileHandle {
    * operation is always default stream ordered like the rest of the non-async CUDA API. In this
    * case, the value of `sync_default_stream` is ignored.
    * @param thread_pool Thread pool to use for parallel execution. Defaults to the global default
-   * thread pool.
+   * thread pool. The caller is responsible for ensuring that the thread pool remains valid until
+   * the returned future is consumed (i.e., until `get()` or `wait()` is called on it).
    * @return Future that on completion returns the size of bytes that were successfully read.
    *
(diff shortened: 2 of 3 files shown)


cpp/include/kvikio/mmap.hpp
```diff
@@ -174,7 +177,8 @@ class MmapHandle {
   std::future<std::size_t> pread(void* buf,
                                  std::optional<std::size_t> size = std::nullopt,
                                  std::size_t offset              = 0,
-                                 std::size_t task_size           = defaults::task_size());
+                                 std::size_t task_size           = defaults::task_size(),
+                                 ThreadPool* thread_pool         = &defaults::thread_pool());
```
madsbk, line 183: Consider using `std::shared_ptr<ThreadPool>` throughout to encourage easier and safer lifetime management for users.

madsbk (commented):
Looks good, but I think it would be useful to add some tests that mix different thread pools.

cpp/include/kvikio/mmap.hpp
kingcrimsontianyu: Thanks. Thinking over passing thread pool as a shared pointer, when the asynchronous function `pread/pwrite` returns, the shared pointer is destroyed. So in order to properly extend the pool's lifetime for the async operation and prevent use-after-free, we need to further share its ownership with the I/O task, either each task or the last aggregate task. The pro is no concern over thread pool lifetime at the point the `std::future` 's result is being waited for. The con is the slight increase in implementation complexity and runtime overhead.

If we pass a raw pointer instead, we claim no ownership responsibility and require users to maintain the pool's lifetime throughout the I/O operations. The pro is simplicity, and the con is loss of bonus of smart pointers.

kingcrimsontianyu: There appears to be a very tricky problem.

To extend the lifetime of the thread pool properly during the asynchronous operations, we need to capture `std::shared_ptr<ThreadPool>` in the last task:
https://github.com/rapidsai/kvikio/blob/main/cpp/include/kvikio/detail/parallel_operation.hpp#L166
```cpp
auto last_task = [=, thread_pool = thread_pool, tasks = std::move(tasks)]() mutable -> std::size_t {
```
Suppose reference count is exactly 1 when the last task is being executed. When it is done, the task goes out of scope precisely at https://github.com/bshoshany/thread-pool/blob/v4.1.0/include/BS_thread_pool.hpp#L938, and the reference count will reach 0 and the pool start being destroyed. In the destructor, we wait (sleep) (https://github.com/bshoshany/thread-pool/blob/v4.1.0/include/BS_thread_pool.hpp#L336) for the condition that `tasks_running == 0`, which will not happen because `--tasks_running` takes place at the beginning of the worker's loop (https://github.com/bshoshany/thread-pool/blob/v4.1.0/include/BS_thread_pool.hpp#L915). So `tasks_running` will always be 1 and we are waiting forever in the destructor. Strangely, I haven't seen this in my unit test, but I fear that the deadlock may appear in production.

kingcrimsontianyu: So I think if we do want to extend the lifetime of the thread pool, we need to add it directly to the returned future's results ("shared state" in C++ terminology), i.e. instead of `std::future<std::size_t>` we probably need `std::future<std::pair<std::size_t, std::shared_ptr<ThreadPool>>>`.

kingcrimsontianyu: At this point, I'm inclined to go back to the raw pointer approach, and ask users to shoulder the responsibility of lifetime management for the thread pool. :thinking:

madsbk: sounds good

madsbk (approved):

PR #878, opened 2025-11-21
[experimental] Implement round-robin thread pool

kingcrimsontianyu:
This PR implements a thread pool with per-thread task queues, to which the tasks are submitted in a round-robin fashion. This thread pool will be used to investigate cuDF pipelining issue in the parquet reader.


kingcrimsontianyu:
Little to no impact to performance.

PR #880, opened 2025-11-27
Add per-block-device thread pools for multi-drive I/O parallelism

kingcrimsontianyu:
## Summary

We have identified the root cause of https://github.com/rapidsai/kvikio/issues/850: With a single global thread pool, I/O tasks for different drives could cluster on submission in a sequential fashion, preventing true hardware parallelism. This PR tackles this problem by creating dedicated thread pools for each physical block device, to ensure that each drive receives concurrent I/O requests independently, achieving expected bandwidth scaling across multiple drives.

This PR resolves the file paths to their underlying block device and creates a new thread pool for each separate block device used by the I/O tasks. For efficient operations, KvikIO first searches for the file path in a `file path --> thread pool` cache, and if not found, performs the device resolving and searches for the block device in a `block device --> thread pool` mapping.

## Usage

The new feature of per-block-device thread pool can be queried and set programmatically in C++:

```cpp
kvikio::defaults::thread_pool_per_block_device();
kvikio::defaults::set_thread_pool_per_block_device(flag);
```
It can also be controlled using the environment variable `KVIKIO_THREAD_POOL_PER_BLOCK_DEVICE=1/true/yes/on` (case insensitive; disabled by default)

## Limitation
- `KVIKIO_NTHREADS` now means the number of threads per pool. There is currently no way to limit the total number of threads across all pools.
- For device-mapper devices (LVM, dm-crypt), this returns the dm device ID, not the underlying physical device(s). This may be suboptimal when multiple LVs share the same underlying physical drive (over-subscription) or when a single LV is striped across multiple drives (under-utilization).

Closes https://github.com/rapidsai/kvikio/issues/850


kingcrimsontianyu:
# Performance result

## System setup
- 1 x H100 GPU
- 4 x Micron_7450_MTFDKCB1T9TFR NVMes

## KvikIO sequential read
- Setting
  - Read four 4-GiB files to the device
  - Each file is on a separate NVMe
  - All `pread` calls are made on the main thread in sequence
  - KvikIO compatibility mode (no GDS)
  - File open/close per iteration
  - Cold page cache, direct I/O

| case  | number of files | total number of threads | bandwidth [MiB/s] |
| -------------   | ------------- | ------------- |------------- |
| main| 1  | 1 | 3141.3071 |
| main| 1  | 4 | 6120.5389 |
| main  | 4  | 4 | 6236.5598 |
| main  | 4  | 16 | 6236.5131|
| this PR  | 4  | 4 | 11257.1692 |
| this PR  | 4  | 16 | 24384.0260 |

cpp/src/file_handle.cpp
```diff
@@ -23,13 +30,155 @@
...
+ *
+ * Limitation: For device-mapper devices (LVM, dm-crypt), this returns the dm device ID, not the
+ * underlying physical device(s). This may be suboptimal when multiple LVs share the same underlying
+ * physical drive (over-subscription) or when a single LV is striped across multiple drives
+ * (under-utilization).
+ *
+ * @param file_path Path to the file whose block device ID is to be determined.
+ * @return The block device ID for the physical block device.
+ */
+dev_t get_block_device_id(std::string const& file_path)
```
madsbk: I think we should make this a public function so users can see how their setup is discovered. It would also be useful if our benchmark scripts printed this information.

In a follow-up PR, it might be useful to introduce an environment variable that allows the setup to be specified manually.

kingcrimsontianyu: Good idea. Beyond `dev_t` alone, other ID-related info can be returned, such as device major/minor number (which together uniquely determine the `dev_t`) and block device name (`nvme1`). I'll make the changes.

>In a follow-up PR, it might be useful to introduce an environment variable that allows the setup to be specified manually.

Could you expand on this a bit? This PR has provided the `KVIKIO_THREAD_POOL_PER_BLOCK_DEVICE` env var to toggle the feature. Were you referring to a feature of printing out the info such as `KVIKIO_PRINT_BLOCK_DEVICE_INFO`?

madsbk: I just mean a way to map file paths to thread pools when `get_block_device_id()` cannot determine it. This might require introducing a configuration file, although that may be more work than we want right now.

cpp/src/file_utils.cpp
```diff
@@ -253,4 +257,86 @@ bool clear_page_cache(bool reclaim_dentries_and_inodes, bool clear_dirty_pages)
...
+
+  struct stat st;
+  SYSCALL_CHECK(stat(file_path.c_str(), &st));
+
+  dev_major_id = major(st.st_dev);
+  dev_minor_id = minor(st.st_dev);
+
+  // Construct sysfs path (which is a symlink) for the block device
+  // e.g., /sys/dev/block/259:8
+  std::string sysfs_path =
```
madsbk:
```suggestion
  std::string const sysfs_path =
```

kingcrimsontianyu: Thanks. I'm working on a branch to expose `get_block_device_info()` and `(set_)thread_pool_per_block_device()` in Python API. I'll improve on the Python benchmark program in a separate PR.

madsbk (approved):

cpp/src/file_utils.cpp
```diff
@@ -253,4 +257,86 @@ bool clear_page_cache(bool reclaim_dentries_and_inodes, bool clear_dirty_pages)
   }
   return false;
 }
+
+BlockDeviceInfo get_block_device_info(std::string const& file_path)
```
vuule, line 261: Should this have an NVTX range?
Also, would be nice to know how long this takes. AFAICT there are opportunities to cache the results, if needed.

kingcrimsontianyu: Thanks. NVTX range and additional cache for `get_block_device_info` (`file name -> block device info` mapping) have been added.

The internal function `get_thread_pool_per_block_device` has already cached the results (`file name -> thread pool` mapping), so the cache for the public API `get_block_device_info` is to improve the speed for standalone use.

Timing results have been added to the last section of https://github.com/rapidsai/kvikio/pull/880#issuecomment-3592197721 .

cpp/src/file_handle.cpp
```diff
@@ -23,13 +26,77 @@
...
+          it != file_path_to_thread_pool_map.end()) {
+        return it->second.get();
+      }
+    }
+
+    // Resolve file path to its underlying block device
+    auto block_dev_info = get_block_device_info(file_path);
+
+    // Check if we already have a thread pool for this block device
+    std::lock_guard lock(mtx);
```
kingcrimsontianyu, line 71: This lock lasts until the function is returned from the try block.

cpp/src/file_handle.cpp
```diff
@@ -23,13 +26,77 @@
...
+  if (!defaults::thread_pool_per_block_device()) { return &defaults::thread_pool(); }
+
+  static std::mutex mtx;
+  static std::unordered_map<std::string, std::shared_ptr<ThreadPool>> file_path_to_thread_pool_map;
+  static std::unordered_map<dev_t, std::shared_ptr<ThreadPool>> dev_id_to_thread_pool_map;
+
+  try {
+    // Fast path: check if this exact file path has been seen before
+    {
+      std::lock_guard lock(mtx);
```
kingcrimsontianyu, line 60: This lock lasts until the block ends. It is to prevent race condition when one thread updates `file_path_to_thread_pool_map` and another thread reads from it.

cpp/src/file_handle.cpp
```diff
@@ -23,13 +26,77 @@
...
+      if (auto it = file_path_to_thread_pool_map.find(file_path);
+          it != file_path_to_thread_pool_map.end()) {
+        return it->second.get();
+      }
+    }
+
+    // Resolve file path to its underlying block device
+    auto block_dev_info = get_block_device_info(file_path);
```
kingcrimsontianyu, line 68: This part is not locked to permit concurrent execution.

cpp/src/file_utils.cpp
```diff
@@ -253,4 +257,106 @@ bool clear_page_cache(bool reclaim_dentries_and_inodes, bool clear_dirty_pages)
...
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+
+  static std::mutex mtx;
```
kingcrimsontianyu, line 265: Similar to the internal function `get_thread_pool_per_block_device`, the public API `get_block_device_info` locks the mutex twice to prevent race condition. The block device resolution is not locked so as to avoid serialization and permit concurrent execution. It is possible that two threads concurrently resolve two exactly same, uncached file path, in which case only one thread will update the `file_path_to_info_map` cache, and the second thread holding the same result will skip it.

cpp/src/file_handle.cpp
```diff
@@ -23,13 +26,77 @@
...
+
+    // Check if we already have a thread pool for this block device
+    std::lock_guard lock(mtx);
+    if (auto it = dev_id_to_thread_pool_map.find(block_dev_info.id);
+        it != dev_id_to_thread_pool_map.end()) {
+      // Cache the file path mapping for future fast-path lookups
```
vuule, line 74: would it be possible to cache using a parent directory, so the result applies to multiple files? I'm really not sure what's feasible here, just an idea.

cpp/src/file_utils.cpp
```diff
@@ -253,4 +257,106 @@ bool clear_page_cache(bool reclaim_dentries_and_inodes, bool clear_dirty_pages)
...
+    if (auto it = file_path_to_info_map.find(file_path); it != file_path_to_info_map.end()) {
+      return it->second;
+    }
+  }
+
+  BlockDeviceInfo block_dev_info{};
+  unsigned dev_major_id{};
+  unsigned dev_minor_id{};
+
+  struct stat st;
```
vuule:
```suggestion
  struct stat st{};
```

cpp/src/file_handle.cpp
```diff
@@ -215,6 +293,13 @@ std::future<std::size_t> FileHandle::pwrite(void const* buf,
                                             ThreadPool* thread_pool)
 {
   KVIKIO_EXPECT(thread_pool != nullptr, "The thread pool must not be nullptr");
+  auto* actual_thread_pool{thread_pool};
+  // Use the block-device-specific pool only if it exists and the user didn't explicitly provide a
+  // custom pool
+  if (_thread_pool != nullptr && thread_pool == &defaults::thread_pool()) {
+    actual_thread_pool = _thread_pool;
+  }
```
vuule: optional
```suggestion
  // Use the block-device-specific pool only if it exists and the user didn't explicitly provide a
  // custom pool
  auto const actual_thread_pool = (_thread_pool != nullptr && thread_pool == &defaults::thread_pool()) ? _thread_pool : thread_pool;
```

cpp/src/file_handle.cpp
kingcrimsontianyu: The challenge is that `/mnt/file1.bin` and `/mnt/file2.bin` are not guaranteed to be on the same drive. It is possible that `/mnt/file1.bin` is a symbolic link to a file on another drive.

cpp/src/file_handle.cpp
kingcrimsontianyu: Thanks. Done. I kept the `*` (i.e. `auto* const actual_thread_pool`) as a reminder that it is a pointer.

cpp/src/file_utils.cpp
kingcrimsontianyu: Thanks done.

PR #893, opened 2026-01-06
Expose cuFile stream register API to Python

kingcrimsontianyu:
This PR exposes cuFile stream register and deregister functions to KvikIO Python API.
```python
kvikio.stream_register(stream, flags)
kvikio.stream_deregister(stream)
```

The stream register and deregister is an optional, performance-tuning API. The users specify the precondition of the I/O parameters by setting the `flags` that indicate when the I/O parameters become valid and whether they are page aligned.

The added Python API can be unit tested using:
```bash
pytest ~/kvikio/python/kvikio/tests/test_async_io.py::test_stream_register_deregister -m cufile -vsx
```
 
A known issue is that, under CUDA 12.9, which is the only setting tested, if the file path is not on a drive with full GDS support, the above unit test will pass but end with a segmentation fault. This is beyond the scope of this PR and is likely not an issue in KvikIO. Further investigation is a good-to-have.

### DIFF
--- a/cpp/src/stream.cpp
+++ b/cpp/src/stream.cpp
@@ -94,9 +94,14 @@ StreamFuture::~StreamFuture() noexcept
 
 void stream_register(CUstream stream, unsigned flags)
 {
+  KVIKIO_NVTX_FUNC_RANGE();
   cuFileAPI::instance().StreamRegister(stream, flags);
 }
 
-void stream_deregister(CUstream stream) { cuFileAPI::instance().StreamDeregister(stream); }
+void stream_deregister(CUstream stream)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  cuFileAPI::instance().StreamDeregister(stream);
+}
 
 }  // namespace kvikio


python/kvikio/kvikio/stream.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from kvikio._lib import stream as stream_module  # type: ignore
+
+
+def stream_register(stream, flags: int) -> None:
```
madsbk:
```suggestion
def stream_register(stream: int, flags: int) -> None:
```

python/kvikio/kvikio/stream.py
```diff
@@ -0,0 +1,30 @@
...
+        CUDA stream which queues the async I/O operations
+    flags: int
+        Specifies when the I/O parameters become valid (submission time or execution
+        time) and what I/O parameters are page-aligned. For details, refer to
+        https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufilestreamregister
+    """
+    stream_module.stream_register(stream, flags)
+
+
+def stream_deregister(stream) -> None:
```
madsbk:
```suggestion
def stream_deregister(stream: int) -> None:
```

madsbk (commented):
We might consider using something like the following for the CUDA Stream bindings
```python
from rmm.pylibrmm.stream import Stream
```
https://docs.rapids.ai/api/rmm/stable/python/pylibrmm/#rmm.pylibrmm.stream.Stream

madsbk (approved):

cpp/src/stream.cpp
```diff
@@ -92,4 +92,11 @@ StreamFuture::~StreamFuture() noexcept
   }
 }
 
+void stream_register(CUstream stream, unsigned flags)
+{
+  cuFileAPI::instance().StreamRegister(stream, flags);
```
vuule, line 98: do we need NVTX ranges?

kingcrimsontianyu: Good catch. Yes we do. The cuFile API will add their own NVTX ranges (disabled by default) which are shown on the global NVTX domain in nsys-ui. For consistency, KvikIO ought to add its own annotation shown in its own domain.
Added!

Issue #786, opened 2025-07-28
Improve libcurl use to improve performance with lower thread count


vuule:
Explore using [libcurl's multi-handle interface](https://curl.se/libcurl/c/libcurl-multi.html) to reduce the CPU utilization when reading remote files.

Not yet clear if we can simply batch slices into a single request, or if multiple read calls from the queue should be grouped into a single multi-handle call.

PR #896, opened 2026-01-12
Add a new remote I/O backend based on libcurl poll-based multi API

kingcrimsontianyu:
This PR adds a new backend for KvikIO remote I/O client using libcurl's poll-based multi interface. Three new settings are introduced, configurable both via environment variables or programmatically:

- `KVIKIO_REMOTE_BACKEND`: `LIBCURL_EASY` (default, using the existing easy handle backend), `LIBCURL_MULTI_POLL` (using the new backend created in this PR).
- `KVIKIO_REMOTE_MAX_CONNECTIONS`: The number of easy handles attached to a multi handle (defaults to 8).
- `KVIKIO_NUM_BOUNCE_BUFFERS`: The k value in k-way bounce buffer.

The current implementation has the following details, subject to future changes:
- When the poll-based backend is used, a `pread` call splits the I/O into chunks of fixed size, just like the easy handle backend. But the chunked reads are not processed in parallel in the thread pool, they are instead processed on a single thread from the thread pool in a multiplexed fashion, enabled by libcurl's multi interface.
- For the same remote handle, different `pread` calls are executed in sequence, enforced by a mutex.
- For different remote handles, different `pread` can be executed in parallel in the thread pool.
 
This PR depends on #898, which makes the bounce buffer moveable.

Addresses most of #786


cpp/include/kvikio/detail/remote_handle_poll_based.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * @brief Check a libcurl easy interface return code and throw on error.
+ *
+ * @param err_code The CURLcode to check.
+ * @param filename Source filename for error reporting.
+ * @param line_number Source line number for error reporting.
+ * @exception std::runtime_error if err_code is not CURLE_OK.
+ */
+inline void check_curl_easy(CURLcode err_code, char const* filename, int line_number)
```
madsbk: Move to `.cpp` ?

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * @brief Check a libcurl multi interface return code and throw on error.
+ *
+ * @param err_code The CURLMcode to check.
+ * @param filename Source filename for error reporting.
+ * @param line_number Source line number for error reporting.
+ * @exception std::runtime_error if err_code is not CURLM_OK.
+ */
+inline void check_curl_multi(CURLMcode err_code, char const* filename, int line_number)
```
madsbk: Move to `.cpp` ?

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * @brief Manages a rotating set of bounce buffers for overlapping network I/O with H2D transfers.
+ *
+ * This class implements k-way buffering, rotating through buffers circularly: while one buffer
+ * receives data from the network, previously filled buffers can be asynchronously copied to device
+ * memory. When all buffers have been used, the class synchronizes the CUDA stream before reusing
+ * buffers.
+ */
+class BounceBufferManager {
```
madsbk, line 23: Would it make sense to use this everywhere we need a bounce buffer? If so, I think it would be good to move `BounceBufferManager` out of this PR and submit a separate PR that introduces `BounceBufferManager` and uses it consistently across the KvikIO?

kingcrimsontianyu: Yes. I agree. This will generalize #520 (double buffering), and benefit existing local and remote I/O handles based on pread/io_uring/easy handle.

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
kingcrimsontianyu: Done. Separated the declaration and definition, and moved them to `cpp/include/kvikio/detail/remote_handle.hpp/.cpp` considering it's not specific to the poll based backend.

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
kingcrimsontianyu: Done. Same as above.

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
```diff
@@ -0,0 +1,133 @@
...
+   * @param num_bounce_buffers Number of bounce buffers to allocate from the pool.
+   */
+  BounceBufferManager(std::size_t num_bounce_buffers);
+
+  /**
+   * @brief Get a pointer to the current bounce buffer's data.
+   *
+   * @return Pointer to the current buffer's memory.
+   */
+  void* data() const noexcept;
```
madsbk, line 37: If it doesn't require too many casts, it would be good to use `std::byte*` instead `void*` throughout. 
```suggestion
  std::byte* data() const noexcept;
```

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
```diff
@@ -0,0 +1,133 @@
...
+   */
+  RemoteHandlePollBased(RemoteEndpoint* endpoint, std::size_t max_connections);
+
+  /**
+   * @brief Destructor that cleans up libcurl multi resources.
+   *
+   * Removes all easy handles from the multi handle and performs cleanup. Errors during cleanup are
+   * logged but do not throw.
+   */
+  ~RemoteHandlePollBased();
```
madsbk:
```suggestion
  ~RemoteHandlePollBased() noexcept;
```
?

cpp/src/detail/remote_handle_poll_based.cpp
```diff
@@ -0,0 +1,251 @@
...
+    _bounce_buffer_idx = 0;
+    CUDA_DRIVER_TRY(cudaAPI::instance().StreamSynchronize(stream));
+  }
+}
+
+RemoteHandlePollBased::RemoteHandlePollBased(RemoteEndpoint* endpoint, std::size_t max_connections)
+  : _max_connections{max_connections}, _transfer_ctxs(_max_connections), _endpoint{endpoint}
+{
+  KVIKIO_EXPECT(defaults::task_size() <= defaults::bounce_buffer_size(),
+                "bounce buffer size cannot be less than task size.");
```
madsbk, line 124: What happens if the user uses `set_task_size()` or `set_bounce_buffer_size()` after `RemoteHandlePollBased` has been construed? Do we need checks later as well?

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
kingcrimsontianyu: Good catch. Added.

cpp/include/kvikio/detail/remote_handle_poll_based.hpp
kingcrimsontianyu: Agreed.
Currently we have quite a number of places where `void*` has been used for buffers from function parameters and class data members. There are other places where `char*` is used to facilitate interaction with libcurl API. Perhaps for the public interface we still use `void*`, but internally wherever applicable, we stick to `std::byte*`. Let's defer this to a future, modernization PR.

kingcrimsontianyu:
Superseded by https://github.com/rapidsai/kvikio/pull/960

PR #916, opened 2026-01-30
[WIP] Optimize easy-handle remote I/O using bounce buffer ring

kingcrimsontianyu:
This PR depends on #913


kingcrimsontianyu:
Will reimplement based on the reactor structure.

PR #919, opened 2026-02-02
Implement CUDA event pool to minimize runtime resource allocation overhead (libcurl multi poll-based backend 2/n)

kingcrimsontianyu:
## Summary

This PR adds a CUDA event pool, a thread-safe singleton that caches reusable CUDA events organized per CUDA context. This eliminates the per-acquisition `cuEventCreate` / `cuEventDestroy` cost in hot paths that need short-lived events (e.g. bounce-buffer recycling).

This PR facilitates the implementation of device buffer support for the remote I/O multiplexing backend.

## Related PR
Depends on #917
Addresses part of #914


cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,117 @@
...
+      it->second.pop_back();
+    }
+  }
+
+  if (event == nullptr) {
+    // Create an event outside the lock to improve performance.
+    // The pool is not updated here; the returned Event object will automatically return the event
+    // to the pool when it goes out of scope
+    CUDA_DRIVER_TRY(cudaAPI::instance().EventCreate(&event, CU_EVENT_DISABLE_TIMING));
+  }
```
wence-, line 98: *question*: Should we have a max size on this event pool? Since we never destroy events, could the pool unboundedly grow in a long-running application? I suppose it will depend on the maximum concurrency of threads issuing reads?

cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  if (event == nullptr) { return; }
+
+  try {
+    std::lock_guard const lock(_mutex);
+    _pools[context].push_back(event);
+  } catch (...) {
+    // If returning to pool fails, destroy the event
+    cudaAPI::instance().EventDestroy(event);
```
wence-: *question*: Not checking the error code because this is called from `~Event()`?

I think we should at least log an error in that case.

cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,117 @@
...
+  std::lock_guard const lock(_mutex);
+  auto it = _pools.find(context);
+  return (it != _pools.end()) ? it->second.size() : 0;
+}
+
+std::size_t EventPool::total_free_events() const
+{
+  std::lock_guard const lock(_mutex);
+  std::size_t total{0};
+  for (auto const& [ctx, events] : _pools) {
```
wence-:
```suggestion
  for (auto const& [_, events] : _pools) {
```

cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+std::size_t EventPool::num_free_events(CUcontext context) const
+{
+  std::lock_guard const lock(_mutex);
+  auto it = _pools.find(context);
+  return (it != _pools.end()) ? it->second.size() : 0;
+}
+
+std::size_t EventPool::total_free_events() const
```
wence-: *question*: Do you need these for correct usage, or are they just going to be introspection facilities?

cpp/src/detail/event.cpp
kingcrimsontianyu: I think it is probably fine to let it grow unbounded. The idea I'm trying to implement in #921 is that each `pread()` builds a "pread context" that gets a `num_threads` number of events from the pool, i.e. a single event for each thread. Each 4-MiB chunked `read()` originating from a specific `pread()` performs the following in sequence:

- I/O (file->bounce buffer ring (#916))
- H2D copy async (bounce buffer ring->device) on a per-thread, per-context stream
- Enqueue the per-`pread`, per-thread, per-context event on the stream

This event is _reused_ for all the chunks on the same thread originating from the same `pread()` call. So the space complexity overall is `O(num_threads * num_concurrent_pread)`, which I think is not likely to blow up the RAM in a long running application.

But I do think that in the future a limitation on the resource pools (currently we have bounce buffer pool, this event pool, and libcurl easy handle pool) is a good feature.

cpp/src/detail/event.cpp
kingcrimsontianyu: Thanks. Agreed. I've added the error logging.

cpp/src/detail/event.cpp
kingcrimsontianyu: Thanks. Done.

cpp/src/detail/event.cpp
kingcrimsontianyu: They are just going to be introspection facilities. I plan to include them in the unit test.

KyleFromNVIDIA (approved):
Approved trivial CMake changes

cpp/src/detail/event.cpp
kingcrimsontianyu: The exception message is now properly logged.

cpp/src/detail/event.cpp
kingcrimsontianyu: These utility functions are now being used in the unit test.

cpp/include/kvikio/detail/event.hpp
```diff
@@ -0,0 +1,192 @@
...
+ * goes out of scope (RAII).
+ *
+ * @note The destructor intentionally does NOT call `cuEventDestroy` on cached events.
+ * `EventPool::instance()` is a function-local static destructed after `main` returns, and
+ * making CUDA driver API calls from a static object's destructor at that point is undefined
+ * behavior per the CUDA programming guide:
+ * https://docs.nvidia.com/cuda/cuda-programming-guide/02-basics/intro-to-cuda-cpp.html#runtime-initialization
+ * The OS reclaims process memory at exit regardless.
+ */
+class EventPool {
```
madsbk: Consider naming them `CudaEventPool` and `CudaEvent`.

cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+void EventPool::put(CUevent event, CUcontext cuda_context) noexcept
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  if (event == nullptr) { return; }
+
+  try {
+    std::lock_guard const lock(_mutex);
+    _pools[cuda_context].push_back(event);
```
madsbk, line 110: This note may not hold for pooled events:

```c++
/**
 * @note An event that has never been recorded reports `is_done() == true`,
 * since CUDA's `cuEventQuery` returns `CUDA_SUCCESS` when there is no
 * captured work.
 */
```

If an `Event` goes out of scope without `synchronize()` being called, it is returned to the pool while its recorded work may still be pending on the stream. A subsequent `pool.get()` can then hand that same event to a new caller, for whom `is_done()` may report `false` even though they have never called `record()` themselves.

Either tighten the note to acknowledge that pooled events may retain pending work from a previous user, or make `put()` reject (or synchronize) non-idle events before returning them to the pool.

madsbk (commented):
Looks good. Do you see any performance gain? Do you have a sense of the overhead of event creation/destruction?

kingcrimsontianyu:
I haven't done a direct comparison between event pool vs on-the-fly event creation/destruction, but Mark (@harrism) did this benchmark before: https://github.com/harrism/cuda_event_benchmark

cpp/include/kvikio/detail/event.hpp
kingcrimsontianyu: Sure! Done!

cpp/src/detail/event.cpp
kingcrimsontianyu: That makes sense. I've updated the comment.

madsbk (approved):
Looks good. Please report back when you get some performance numbers

kingcrimsontianyu:
Sure. Thanks for the review!

PR #954, opened 2026-04-14
Use CUDA batch memory copy API wherever possible

kingcrimsontianyu:
The use of CUDA batched memory copy for general CPU-GPU copy is recommended by the driver team, as it avoids certain limitations in the traditional memory copy API, such as unexpected device-wide synchronizations. This PR replaces `cuMemcpyHtoDAsync` and `cuMemcpyDtoHAsync` with `cuMemcpyBatchAsync`.

### DIFF
--- a/cpp/include/kvikio/shim/cuda.hpp
+++ b/cpp/include/kvikio/shim/cuda.hpp
@@ -117,6 +117,23 @@ class cudaAPI {
   void operator=(cudaAPI const&) = delete;
 
   KVIKIO_EXPORT static cudaAPI& instance();
+
+  /**
+   * @brief Copy memory asynchronously, preferring the batched API when available.
+   *
+   * Uses `cuMemcpyBatchAsync` with stream-ordered source access when the CUDA driver supports it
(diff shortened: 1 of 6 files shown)


cpp/include/kvikio/shim/cuda.hpp
```diff
@@ -86,6 +86,7 @@ class cudaAPI {
   decltype(cuMemHostUnregister)* MemHostUnregister{nullptr};
   decltype(cuMemcpyHtoDAsync)* MemcpyHtoDAsync{nullptr};
   decltype(cuMemcpyDtoHAsync)* MemcpyDtoHAsync{nullptr};
```
vuule, line 88: are these still needed?

vuule (approved):
Looks good, just one question

cpp/include/kvikio/shim/cuda.hpp
kingcrimsontianyu: They will not be used, but since this is currently part of a public API, I hope they can stay.

PR #957, opened 2026-05-01
fix(ci): resolve all zizmor findings and add zizmor pre-commit checks

gforsyth:
Similar to upstream changes in `shared-workflows`, this PR cleans up and annotates all of the workflows and adds the `zizmor` linter to make sure changes are checked.

Part of https://github.com/rapidsai/build-planning/issues/275

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -108,6 +108,10 @@ repos:
     rev: v0.10.0.1
     hooks:
       - id: shellcheck
+  - repo: https://github.com/zizmorcore/zizmor-pre-commit
+    rev: v1.24.1
+    hooks:
+      - id: zizmor
 
 
 default_language_version:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
     hooks:
       - id: cython-lint
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.3.0'
+    rev: v1.20.2
     hooks:
       - id: mypy
         additional_dependencies: [types-cachetools, zarr]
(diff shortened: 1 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
     hooks:
       - id: cython-lint
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.3.0'
+    rev: v1.20.2
```
gforsyth, line 28: This was failing in CI on the old version and I couldn't reproduce locally, so I just bumped to the latest release.  The only required change was marking an implicitly optional arg as explicitly optional in `numpy.py`

kingcrimsontianyu (approved):
Thanks for the update. Approved the Python change.

.pre-commit-config.yaml
jakirkham: It's probably a good idea for us to bump these periodically. Thank you for doing that Gil! 🙏

Is there a good way for us to check for outdated versions in CI and error. Or better still open both PRs to update these?

jakirkham (approved):
Thank you Gil for working on this PR! And thank you Tianyu for reviewing 🙏

.pre-commit-config.yaml
gforsyth: There's beta support for having `renovate` bump versions in `pre-commit` -- I can open a `build-planning` issue to investigate.

gforsyth: https://github.com/rapidsai/build-planning/issues/276

jakirkham: That sounds great! Thanks Gil 🙏

PR #960, opened 2026-05-12
Create a new remote I/O backend: libcurl multi poll-based (1/n: host support)

kingcrimsontianyu:
This PR starts to implement `MULTI_POLL` as a second backend for `RemoteHandle::pread()`, alongside the existing thread-pool-based `EASY_THREADPOOL`. The backend is selected by an environment variable at process start, and the public `pread()` API is unchanged. Only host buffer is supported in this PR. 

Three new environment variables are introduced, all read once at process start (values case-insensitive):

- `KVIKIO_REMOTE_IO_BACKEND`: The remote I/O backend. Default `EASY_THREADPOOL`. Allowed values: `EASY_THREADPOOL`, `MULTI_POLL`.
- `KVIKIO_REMOTE_IO_NUM_REACTORS`: Number of reactor threads, only used when `MULTI_POLL` is selected. Default 1.
- `KVIKIO_REMOTE_IO_REACTOR_DISPATCH`: Sub-range dispatch pattern. Default `PER_CHUNK`. Allowed values: `PER_CHUNK` round-robins sub-ranges across reactors. `PER_PREAD` keeps all sub-ranges of one `pread()` on the same reactor. 

See comments below for performance results.

Supersede #896


kingcrimsontianyu:
# Performance

The benchmark reads the TPC-H SF-3K lineitem dataset, consisting of 1133 parquet files of ~620 MB each, from AWS S3 into host memory. Files are processed in fixed order, 4 at a time. Each batch issues 4 concurrent `RemoteHandle::pread()` calls and awaits all 4 futures before submitting the next batch. Each configuration runs for 60 seconds of cumulative read wall time.

```
CPU% = (user + sys) / wall * 100 %
+----------+--------+----------------------------------+----------------------------------+----------------------------------+
| backend  | thread |              warmup              |           steady state           |             overall              |
|          |        +--------+---------+---------------+--------+---------+---------------+--------+---------+---------------+
|          |        |   bw   |  cpu%   |  cpu%/(GiB/s) |   bw   |  cpu%   |  cpu%/(GiB/s) |   bw   |  cpu%   |  cpu%/(GiB/s) |
+----------+--------+--------+---------+---------------+--------+---------+---------------+--------+---------+---------------+
| easy     |   1    | 0.0540 |  4.91   |    90.9040    | 0.0546 |  4.65   |    85.1122    | 0.0543 |   4.78  |    88.0076    |
| easy     |   16   | 0.6737 | 191.42  |   284.1294    | 0.7251 |  59.75  |    82.4046    | 0.7221 |  67.59  |    93.6095    |
| easy     |   128  | 0.7061 | 2912.84 |   4125.1773   | 4.7381 | 535.94  |   113.1130    | 4.5083 | 671.42  |   148.9290    |
| easy     |   256  | 0.3813 | 3019.73 |   7919.9805   | 6.1546 | 692.25  |   112.4763    | 5.5462 | 937.54  |   169.0427    |
|multi-poll|   1    | 0.2042 |   19.71 |     96.5227   | 1.4822 |  93.32  |    62.9614    | 1.2397 |  79.35  |    64.0100    |
|multi-poll|   4    | 0.4429 |   71.38 |    161.1680   | 3.9316 | 357.79  |    91.0033    | 3.6153 | 331.82  |    91.7828    |
|multi-poll|   8    | 2.8957 |  757.97 |    261.7604   | 6.1265 | 633.36  |   103.3808    | 6.0816 | 635.10  |   104.4295    |
+----------+--------+--------+---------+---------------+--------+---------+---------------+--------+---------+---------------+
```

- EASY threads=1 delivers 0.0546 GiB/s. MULTI_POLL threads=1 delivers 1.48 GiB/s, 27x higher due to multiplexing.
- EASY threads=256 reaches 6.15 GiB/s steady, and MULTI_POLL threads=8 reaches 6.13 GiB/s steady. The multi backend can achieve the same throughput with 32x fewer threads, and with 8% higher CPU efficiency (CPU%).
- MULTI_POLL’s warmup CPU is 4-150x lower than EASY, presumably because much fewer threads require much fewer TLS-handshake.

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -0,0 +1,225 @@
...
+        lock.lock();
+      }
+    }
+
+    // (2) Drive transfers in a non-blocking way.
+    int running_handles   = 0;
+    auto const perform_mc = curl_multi_perform(_curl_multi, &running_handles);
+    KVIKIO_EXPECT(perform_mc == CURLM_OK,
+                  std::string("curl_multi_perform: ") + curl_multi_strerror(perform_mc),
+                  std::runtime_error);
```
madsbk: I think we need to handle these errors more gracefully. Currently, a `KVIKIO_EXPECT`
throw escapes `io_thread_main()` and propagates into `std::thread`'s top-level
frame, which calls `std::terminate()` and kills the process mid-transaction.

I'd suggest something like:

* failing all currently in-flight aggregates via `on_subrange_failed()`, so
  callers observe a real exception through the future, and
* then either exiting the loop cleanly or rethrowing internally.

cpp/include/kvikio/detail/multi_poll_reactor.hpp
```diff
@@ -0,0 +1,183 @@
...
+  void submit_pread(std::vector<std::unique_ptr<RemoteMultiTransfer>> transfers);
+
+ private:
+  MultiReactorPool();
+  ~MultiReactorPool() noexcept;
+
+  std::vector<std::unique_ptr<MultiPollReactor>> _reactors;
+  RemoteReactorDispatch _dispatch;
+  std::atomic<std::size_t> _per_pread_counter{0};  // Round-robin counter for PER_PREAD mode.
+  std::atomic<std::size_t> _per_chunk_counter{0};  // Round-robin counter for PER_CHUNK mode.
```
madsbk: Would it be confusing to just have one counter since they are never used together?

madsbk (commented):
Overall looks very good, but I think we need to consider more graceful error handling.

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Great catch. Since more than one reactor can be used on the I/O threads, I think we may let the whole reactor pool fail, and any subsequent call to `fut.get()` from the non-I/O threads will get the propagated exception. I will come up with an implementation of this.

cpp/include/kvikio/detail/multi_poll_reactor.hpp
kingcrimsontianyu: That makes sense. I will combine them into a single `_next_reactor_counter`.

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Thanks for the reminder of exception handling in `std::thread`. This has been addressed by the new changes.

A note on the failure behavior. If a certain transfer for a sub-range of a certain file fails for some reason (such as bad connection), the error is reflected in the return value of `curl_multi_info_read` instead of `curl_multi_perform`, and the caller's `future.get()` call for that failed transfer will throw, whereas other `future`s with successful transfer will return size normally. This was previously well handled in this PR.

What was not well handled is the `KVIKIO_EXPECT` macro for those libcurl API calls inside the I/O thread. Under normal circumstances these expectations should never fail. But if things go badly somehow, the thrown exception would go uncaught. The idea in the new changes is that, if one or more threads encounter exceptions, the first thread will record the exception, the entire reactor pool be killed, all reactors queues be drained, all ongoing and future `pread` calls have a `future.get()` that will throw the same exception.

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -0,0 +1,314 @@
...
+MultiPollReactor::~MultiPollReactor() noexcept
+{
+  // Intentionally empty body. `MultiReactorPool` is a leaked-pointer singleton, so its
+  // `_reactors` vector and the `std::unique_ptr<MultiPollReactor>` elements inside it
+  // are never destroyed. We declare this dtor so the type is complete and usable in
+  // `std::unique_ptr`. Running it would call dtor on an unjoined thread and call
+  // `std::terminate()`.
+}
+
+void MultiPollReactor::wakeup() noexcept { (void)curl_multi_wakeup(_curl_multi); }
```
madsbk:
```suggestion
void MultiPollReactor::wakeup() noexcept {std::ignore = curl_multi_wakeup(_curl_multi); }
```

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -0,0 +1,314 @@
...
+}
+
+std::future<std::size_t> RemoteMultiAggregateContext::get_future() { return _promise.get_future(); }
+
+MultiPollReactor::MultiPollReactor(MultiReactorPool* pool) : _pool{pool}
+{
+  KVIKIO_EXPECT(
+    _pool != nullptr, "MultiPollReactor requires a non-null pool", std::invalid_argument);
+  // Force LibCurl global init before we create the multi handle.
+  (void)LibCurl::instance();
```
madsbk:
```suggestion
  std::ignore = LibCurl::instance();
```

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+  // In-flight is touched only by the I/O thread, which is us, so no lock needed.
+  for (auto& [easy, transfer] : _in_flight) {
+    // Best-effort removal.
+    // TODO: Known issue: if curl_multi_remove_handle fails (rarely happens), the easy handle
+    // remains attached to the multi handle. _in_flight.clear() below then destroys the transfer's
+    // CurlHandle, which unconditionally returns the easy handle to the LibCurl free pool. A future
+    // caller that pulls this handle would operate on a handle that libcurl still considers
+    // attached, which is undefined behavior.
+    (void)curl_multi_remove_handle(_curl_multi, easy);
```
madsbk:
```suggestion
    std::ignore = curl_multi_remove_handle(_curl_multi, easy);
```

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -0,0 +1,314 @@
...
+    transfer->aggregate->on_subrange_failed(eptr);
+  }
+  _in_flight.clear();
+}
+
+MultiReactorPool::MultiReactorPool() : _dispatch{defaults::remote_io_reactor_dispatch()}
+{
+  // Force LibCurl global init before any reactor opens a multi handle.
+  (void)LibCurl::instance();
```
madsbk:
```suggestion
  std::ignore = LibCurl::instance();
```

cpp/src/remote_handle.cpp
```diff
@@ -810,23 +778,92 @@ std::future<std::size_t> RemoteHandle::pread(void* buf,
...
-  };
-  return detail::parallel_io(
-    task,
-    buf,
-    size,
-    file_offset,
-    task_size,
-    0,
-    {.thread_pool = thread_pool, .call_idx = call_idx, .nvtx_color = nvtx_color});
+
```
madsbk, line 782: To avoid overflow in `setup_range_request_impl`, we should guard against zero. Something like:

```suggestion
  if (size == 0) {
    std::promise<std::size_t> p;
    p.set_value(0);
    return p.get_future();
  }
```

madsbk (commented):
It would be good with a end-to-end integration test for `MULTI_POLL`.

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Done

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Done

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Done

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Done.
Just a note that using `std::ignore` in place of `(void)` to suppress the no discard warning is a bit controversial, because it requires including `<tuple>` and seems to be specially designed for tuple. At lower optimization level it generates more instructions than `(void)` or `[[maybe_unused]] auto _=` (confirmed in compiler explorer). But at higher optimization level it converges with `(void)`.

cpp/src/remote_handle.cpp
kingcrimsontianyu: Agreed. Updated using an existing utility function `make_ready_future`.

kingcrimsontianyu:
Because backend choice is made via environment variable in this PR, the unit test was done merely by setting `export KVIKIO_REMOTE_IO_BACKEND=MULTI_POLL` locally. I think we may support runtime backend selection in the future for the unit tests.

cpp/src/detail/multi_poll_reactor.cpp
madsbk: Ah, didn't know that. Good to know. I still prefer `std::ignore` for readability, but I'll leave it up to you :)

madsbk (approved):
Thanks @kingcrimsontianyu

PR #965, opened 2026-05-24
Implement per-(thread, CUDA context) bounce buffer cache singleton (libcurl multi poll-based backend 3/n)

kingcrimsontianyu:
This PR introduces a per-(thread, CUDA context) cache of bounce buffers with asynchronous, stream-ordered recycling. This establishes the foundation for the device-side concurrency control for the MULTI_POLL remote I/O backend. It bounds pinned-memory pressure per reactor while letting completed H2D transfers return their buffers to a free list without blocking the reactor thread.


cpp/include/kvikio/bounce_buffer.hpp
```diff
@@ -199,8 +199,10 @@ class BounceBufferPool {
    *
    * @param buffer Pointer to memory to return
    * @param size Size of the buffer in bytes
+   *
+   * @note noexcept: any failure during the underlying push or deallocation is caught and logged.
    */
-  void put(void* buffer, std::size_t size);
+  void put(void* buffer, std::size_t size) noexcept;
```
kingcrimsontianyu, line 205: This PR is marked breaking since this function is made `noexcept` now.

cpp/src/detail/bounce_buffer_cache.cpp
```diff
@@ -0,0 +1,150 @@
...
+BounceBufferCachePerThreadAndContext<Allocator>&
+BounceBufferCachePerThreadAndContext<Allocator>::instance()
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  static auto* _instance = []() {
+    auto const max_total = defaults::remote_io_max_concurrent_requests();
+    auto const n         = defaults::remote_io_num_reactors();
+    std::optional<std::size_t> const per_reactor_max =
+      (max_total == 0) ? std::nullopt : std::optional{std::max<std::size_t>(max_total / n, 1)};
+    return new BounceBufferCachePerThreadAndContext(per_reactor_max);
```
kingcrimsontianyu, line 156: I think this bounce buffer cache here will be useful for the io_uring implementation in the future (in place of the bounce buffer ring approach, which enforces usage ordering and likely hurts performance). But for now let it be specific to the MULTI_POLL backend. In the future perhaps we can make `BounceBufferCachePerThreadAndContext` templated on an internal enum to differentiate local I/O from remote I/O.

cpp/src/detail/bounce_buffer_cache.cpp
```diff
@@ -0,0 +1,150 @@
...
+  return *it->second;
+}
+
+template <typename Allocator>
+std::optional<typename BounceBufferCachePerThreadAndContext<Allocator>::Buffer>
+BounceBufferCachePerThreadAndContext<Allocator>::try_get(CUcontext ctx)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  auto& shard = get_shard(ctx);
+
```
madsbk:
```suggestion
```

cpp/src/detail/bounce_buffer_cache.cpp
```diff
@@ -0,0 +1,150 @@
...
+  try {
+    std::lock_guard const lock(data->shard->mutex);
+    data->shard->free.push_back(std::move(data->buffer));
+    --data->shard->in_flight;
+  } catch (std::exception const& e) {
+    KVIKIO_LOG_ERROR(std::string("BounceBufferCachePerThreadAndContext::recycle_callback: ") +
+                     e.what());
+  } catch (...) {
+    KVIKIO_LOG_ERROR("BounceBufferCachePerThreadAndContext::recycle_callback: unknown exception");
+  }
```
madsbk, line 143: i think there is two issues here if `shard.free.push_back()` throws `std::bad_alloc` during vector reallocation:

1. `in_flight` counter leak: `--data->shard->in_flight` never executes, so the shard permanently loses one cap slot.
2. CUDA API call from inside the callback: `data` destructs with `_buffer != nullptr`, since the buffer was never moved out. That calls `BounceBufferPool::put()` → `cuMemFreeHost()`, but `cuLaunchHostFunc` prohibits CUDA API calls from the callback.

Both can be fixed by pre-reserving space in the free vector in `recycle_after()`, before enqueuing the callback.

cpp/src/detail/bounce_buffer_cache.cpp
```diff
@@ -0,0 +1,150 @@
...
+    std::lock_guard const lock(shard.mutex);
+    --shard.checked_out;
+    ++shard.in_flight;
+  }
+
+  KVIKIO_CUDA_DRIVER_TRY(cudaAPI::instance().LaunchHostFunc(stream, &recycle_callback, data.get()));
+  // Ownership of the heap payload transfers to the callback.
+  // It is okay if at this point, the callback has already be executed, and the heap payload
+  // destroyed, in which case, `release` here would simply return nullptr instead of a pointer to
+  // the managed object.
```
madsbk: Nit: I don't think that's accurate. `release()` always returns the stored pointer. The code itself looks correct, the comment just needs updating to say that `release()` relinquishes ownership, making the `unique_ptr` destructor a no-op.

cpp/src/detail/bounce_buffer_cache.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  auto& shard = get_shard(ctx);
+  auto data   = std::make_unique<RecycleCallbackData>(RecycleCallbackData{&shard, std::move(buf)});
+
+  // Phase A (`checked_out`) ends and Phase B (`in_flight`) starts.
+  {
+    std::lock_guard const lock(shard.mutex);
+    --shard.checked_out;
+    ++shard.in_flight;
```
madsbk, line 108: The state transition (`--checked_out; ++in_flight`) happens before the `cuLaunchHostFunc` call. If `cuLaunchHostFunc` throws, `data` is correctly destroyed and returns the buffer to the pool, but `in_flight` is never decremented, permanently losing one capacity slot. Maybe undo the state transition before rethrowing?

madsbk (changes requested):
Overall looks good

cpp/include/kvikio/detail/bounce_buffer_cache.hpp
```diff
@@ -0,0 +1,168 @@
...
+   * `defaults::remote_io_num_reactors()`, giving device transfers the same per-reactor budget as
+   * host transfers. The singleton is intentionally heap-allocated and never deleted. Each template
+   * instantiation (different `Allocator`) has its own singleton.
+   *
+   * @return Reference to the singleton instance.
+   */
+  KVIKIO_EXPORT static BounceBufferCachePerThreadAndContext& instance();
+
+  /**
+   * @brief Get the per-(thread, CUDA context) buffer cap this instance was constructed with.
```
bdice: Is this a number of buffers or a number of bytes (total bytes or bytes per buffer)?

bdice (approved):
A bit of docs clarity on “cap” would be nice. Otherwise this is awesome! Should be good to go once @madsbk’s comments are addressed.

cpp/src/detail/bounce_buffer_cache.cpp
kingcrimsontianyu: Very interesting thought.

For 1, I've moved the prefix decrement before `push_back()`.

For 2, the issue is that the callback is executed in the CUDA stream order, which may happen way after `recycle_after()` returns. So it is possible that we have several `recycle_after()` calls in sequence, and none of their callbacks have fired yet. In this case, inside `recycle_after()`, for the `free.reserve(cur_size + 1)` call, the `cur_size` will be the same, so some `push_back` calls in the callback would perform allocation anyway.

I've changed the design a bit, so that the free list is pre-reserved at the shard's construction, if the cap is limited. I've documented the unhandled edge case where no cap (unlimited) + push_back OOM + buffer change would cause CUDA API to be called in the callback.

cpp/src/detail/bounce_buffer_cache.cpp
kingcrimsontianyu: That's right. I've revised the comment here.

cpp/include/kvikio/detail/bounce_buffer_cache.hpp
kingcrimsontianyu: Thanks. Comments revised. The cap refers to the number of bounce buffers per-(thread, CUDA context).

cpp/src/detail/bounce_buffer_cache.cpp
kingcrimsontianyu: Good idea. Done.

madsbk (approved):
Thanks @kingcrimsontianyu, looks very good!

PR #966, opened 2026-05-24
Support device buffer for the new libcurl multi poll-based backend (5/n)

kingcrimsontianyu:
This PR supports device buffer for the multi poll-based remote I/O backend. In addition, this PR makes 5 improvements:
- Refactor `ConcurrentRequestLimiter` with RAII Slot. `try_acquire()` now returns a move-only `Slot` that auto-releases on destruction. Acquire/release pairing is now structurally enforced.
- Add a new `CurlMultiAttachment` RAII guard, that removes the easy handle from the multi handle on destruction.
- Batch the transfer submission to reduce the number of `curl_multi_wakeup`  calls from the initial "per sub-range" to the current "per reactor, per pread".
- Copy `_inbox` (shared by the submission thread and reactor thread) to a reactor-private `_pending` deque, to decouple the submission latency from reactor's admission walk.
- Use AI to significantly condense the comments and remove incorrect, unclear or redundant comments.


kingcrimsontianyu:
Sending this PR to draft. The reactor thread logic is very messy and can potentially be improved by RAII. Refactoring to have a cleaner code now.

kingcrimsontianyu:
## Performance results
- 2026.07.24 cudf-polars TPC-H queries. "v2" is simply a rerun. 
```
t: number of threads
r: number reactors
c: connections
+----------------------+----------+------------+--------------+
| Case                 | Date     | Total (s)  | Δ from base  |
+----------------------+----------+------------+--------------+
| easy_256t (baseline) | 07.25    |  278.6069  |     ——       |
| easy_256t (v2)       | 07.25    |  272.7131  |    -2.12%    | --> best result
| easy_256t            | 07.24    |  284.9486  |    +2.28%    |
| easy_256t (v2)       | 07.24    |  293.2409  |    +5.25%    |
+----------------------+----------+------------+--------------+
| mp_16r/512c          | 07.25    |  294.6391  |    +5.75%    |
| mp_16r/512c (v2)     | 07.25    |  291.7514  |    +4.72%    |
| mp_16r/512c          | 07.24    |  302.0218  |    +8.40%    |
| mp_16r/512c (v2)     | 07.24    |  300.1998  |    +7.75%    |
| mp_16r/256c          | 07.25    |  282.7292  |    +1.48%    |
| mp_16r/256c (v2)     | 07.25    |  280.0991  |    +0.54%    | --> best result
+----------------------+----------+------------+--------------+
```
The results above show that: (1) For EC2-S3, there exists certain level of network bandwidth variance. (2) Comparing the best results, the new multi-poll backend is ~3% slower than the existing easy thread pool backend, with the benefit of thread usage cut by a factor of 16.

kingcrimsontianyu:
## Event pool vs on-the-fly creation/destruction

On a temporary branch, the event pool was modified such that the `get()` method creates a CUDA event and `put()` method destroys it. Comparing use of event pool vs on-the-fly event creation/destruction on cudf-polars TPC-H SF-1k, **no** statistically significant difference in performance was observed.
- Event pool runs: total time ranges 280~285 seconds.
- On-the-fly creation/destruction: total time ranges 284~287 seconds.

cc @madsbk

cpp/include/kvikio/detail/io_event_barrier.hpp
```diff
@@ -0,0 +1,74 @@
...
+ * thread id) after every `cuMemcpyAsync`, re-recording the same event on later calls. Once all
+ * sub-ranges report completion, the caller calls `sync_all_events()` to block until each thread's
+ * last H2D has drained.
+ */
+class IoEventBarrier {
+ public:
+  /**
+   * @brief Construct a barrier carrying `cuda_context` as metadata.
+   *
+   * @param cuda_context The CUDA context that pred's H2Ds will land in. Stored only for callers
```
bdice:
```suggestion
   * @param cuda_context The CUDA context that pread's H2Ds will land in. Stored only for callers
```

cpp/include/kvikio/detail/multi_poll_reactor.hpp
```diff
@@ -48,6 +51,11 @@ class RemoteMultiAggregateContext {
    */
   explicit RemoteMultiAggregateContext(std::size_t num_subranges);
 
+  /**
+   * @brief Per-pread event watermark for the device-buffer path.
```
bdice: Is "watermark" a common term of art here? I'm not familiar with what this means, it might be made up by an LLM?

cpp/src/detail/concurrent_request_limiter.cpp
```diff
@@ -1,53 +1,81 @@
...
+}
+
+ConcurrentRequestLimiter::Slot::~Slot() noexcept { reset(); }
+
+ConcurrentRequestLimiter::Slot::Slot(Slot&& o) noexcept
+  : _limiter{std::exchange(o._limiter, nullptr)}
+{
+}
+
+ConcurrentRequestLimiter::Slot& ConcurrentRequestLimiter::Slot::operator=(Slot&& o) noexcept
```
bdice: Let's use `other` instead of `o`. One character variables are generally not favored in our codebases.

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -8,22 +8,85 @@
...
+
+CurlMultiAttachment::~CurlMultiAttachment()
+{
+  // Best-effort detach on the reactor I/O thread. If curl_multi_remove_handle fails (rare), the
+  // handle stays attached and the owning CurlHandle still returns it to the LibCurl pool, which is
+  // undefined behavior in libcurl. A destructor has no better recovery.
+  detach_from_multi(_multi, _easy);
+}
+
+CurlMultiAttachment::CurlMultiAttachment(CurlMultiAttachment&& o) noexcept
```
bdice: Let's use `other` here and throughout instead of `o`.

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -8,22 +8,85 @@
...
+  : _multi{std::exchange(o._multi, nullptr)}, _easy{std::exchange(o._easy, nullptr)}
+{
+}
+
+CurlMultiAttachment& CurlMultiAttachment::operator=(CurlMultiAttachment&& o) noexcept
```
bdice: Do we need to explicitly `= delete;` the copy constructor and copy assignment? (Trying to remember the rules.)

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -80,133 +142,205 @@ MultiPollReactor::MultiPollReactor(MultiReactorPool* pool,
...
+          std::swap(_pending, _inbox);
+        } else {
+          while (!_inbox.empty()) {
+            _pending.push_back(std::move(_inbox.front()));
+            _inbox.pop_front();
+          }
+        }
+      }
+
+      // Admission walk over the reactor-private _pending. Each entry is either admitted to libcurl
```
bdice: "Admission walk over the reactor-private _pending." What does this mean? See if you can rewrite this.

cpp/include/kvikio/detail/io_event_barrier.hpp
kingcrimsontianyu: Fixed.

cpp/include/kvikio/detail/multi_poll_reactor.hpp
kingcrimsontianyu: No jargon used here. The per-pread, per-thread event from the `IoEventBarrier` is simply used repeatedly this way:
```
H2D (from pread 1) -> record event 1 -> 
H2D (from pread 1) -> record the same event (from pread 1) ->
H2D (from pread 2), record event 2"
...
```
Anyway I've removed the mention of "watermark" to avoid confusion.

cpp/src/detail/concurrent_request_limiter.cpp
kingcrimsontianyu: Done. Also replaced other uses of `o` with `other`.

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Done.

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: The move functions cause the copy functions to be deleted, but [the rule of five](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all) recommends explicit deletion for readability.

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Simplified a bit.

cpp/src/remote_handle.cpp
```diff
@@ -896,7 +915,15 @@ std::future<std::size_t> RemoteHandle::pread(void* buf,
...
   // One pool call per pread(). The pool consults the captured dispatch policy internally.
   detail::MultiReactorPool::instance().submit_pread(std::move(transfers));
-  return fut;
+
+  if (is_host_mem) { return fut; }
+
+  return std::async(std::launch::deferred,
+                    [fut = std::move(fut), io_event_barrier]() mutable -> std::size_t {
+                      auto const n = fut.get();
+                      io_event_barrier->sync_all_events();
```
vuule, line 924: `sync_all_events()` is documented as context-agnostic, but it synchronizes events created in the pread’s CUDA context. Make `IoEventBarrier::sync_all_events()` push `_cuda_context` before synchronizing its events, and add coverage where the future is consumed on a thread with no current CUDA context.

vuule (commented):
looks good, just one potential concern

cpp/src/remote_handle.cpp
kingcrimsontianyu: So this lambda with deferred policy is evaluated at the `future.get()` time, where the CUDA context is allowed to be different from pread's. I've updated the doc to clarify what context-agnostic means.

I've added a unit test for `sync_all_events()` which is called when (1) there is no context current on the calling thread, (2) there is a different context.

bdice (approved):

vuule (approved):

PR #974, opened 2026-06-01
PERF: Avoid duplicate HEAD requests for file size with S3 endpoint

### LINKED ISSUE #973: Duplicate`get_file_size` / HEAD request for S3 endpoints
I have a simple pylibucdf script that reads a table from S3:

```python
import pylibcudf as plc


sources = plc.io.types.SourceInfo(["s3://rapids-tpch/scale-10/nation/part.0.parquet"])
options = plc.io.parquet.ParquetReaderOptions.builder(sources).build()
table = plc.io.parquet.read_parquet(options)
```

Running that under nsys with `nsys profile --python-backtrace=cuda --python-sampling=true -o simple -f true --trace=cuda,osrt,nvtx python simple.py` shows we make two calls to `S3Endpoint::get_file_size()`:

Image

We should be able to do that just once.

Related to https://github.com/rapidsai/cudf/issues/22662

TomAugspurger:
Our S3 implementaiton made two HEAD requests on S3 urls in "AUTO" mode (the only mode that's exposed to libcudf users).

1. A first HEAD request to test S3 credentials
2. A second HEAD request to get the file size

We can avoid the second HEAD request by reusing the response to the first HEAD request when we go to actually open the file.

Closes #973

Here's an screenshot of an nsys profile showing the single `get_file_size()` / `HEAD` request on this script:

```python
import pylibcudf as plc


sources = plc.io.types.SourceInfo(["s3://rapids-tpch/scale-10/nation/part.0.parquet"])
options = plc.io.parquet.ParquetReaderOptions.builder(sources).build()
table = plc.io.parquet.read_parquet(options)
```

image

### DIFF
--- a/cpp/src/remote_handle.cpp
+++ b/cpp/src/remote_handle.cpp
@@ -672,7 +672,9 @@ RemoteHandle RemoteHandle::open(std::string url,
   }
 
   if (nbytes.has_value()) { return RemoteHandle(std::move(endpoint), nbytes.value()); }
-  if (probed_nbytes.has_value()) { return RemoteHandle(std::move(endpoint), probed_nbytes.value()); }
+  if (probed_nbytes.has_value()) {
+    return RemoteHandle(std::move(endpoint), probed_nbytes.value());
+  }
   return RemoteHandle(std::move(endpoint));
 }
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
I'm looking into the final HTTP head request. This comes from the `AUTO` endpoint type, which is the only thing used by libcudf, to check whether or not we should fall back to the public endpoint thing. I suspect this will be best solved in libcudf, but I'm not quite sure on that.

madsbk (approved):
Looks good

PR #978, opened 2026-06-13
Limit concurrency to counterintuitively improve overall performance for host buffer (libcurl multi poll-based backend 4/n)

kingcrimsontianyu:
[Our previous result](https://github.com/rapidsai/kvikio/pull/960#issuecomment-4469200925) showed that when reading a sequence of ~620 MB files into host buffer, the multi poll backend achieved the same level of aggregate bandwidth with the easy thread pool backend, on the order of 6.1 GiB/s, on an EC2 instance with theoretical bandwidth of 11.64 GiB/s.

However, a follow-up study on SOL shows that when reading a large 8 GiB file, the easy thread pool backend is able to saturate the bandwidth, but the multi poll backend falls far behind, even lower than 6.1 GiB/s. It is found that:

- This slowdown is NOT the well-documented [S3 throttling](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html), as we did not receive the HTTP 503 (slow down) response.
- When the slowdown happened, the kernel received fewer incoming TPC packets per second, indicating that this might be related to an unknown server-side throttling mechanism.

While the exact reason is currently not well understood, an effective workaround has been found: do NOT leave the concurrency unbounded, but instead impose a limit. This PR introduces the setting `KVIKIO_REMOTE_IO_MAX_CONCURRENT_REQUESTS` that caps the global concurrency for the host buffer read. Each reactor is allowed to have `KVIKIO_REMOTE_IO_MAX_CONCURRENT_REQUESTS / num_reactors` concurrent requests, and if the limit is reached, the request stays in the inbox until admitted requests complete. This method improves the performance effectively:

```
host buffer, 8 GiB data read repeatedly
Meaning of configs: t: threads, c: connections, r: reactors
In main, the multi poll concurrency is unbounded
+------------+------------+----------+------------+---------------+
|  branch    |  backend   |  config  | bw (GiB/s) |  cpu%/(GiB/s) |
+------------+------------+----------+------------+---------------+
|            |    easy    |   128t   |   9.81     |     165.1     |
|   main     | multi-poll |    8r    |   5.87     |     129.0     |
|            | multi-poll |   16r    |   5.13     |     167.0     |
+------------+------------+----------+------------+---------------+
| this PR    | multi-poll | 16r/128c |   9.68     |     115.1     |
+------------+------------+----------+------------+---------------+
```
With 16 reactors and a cap of 128 connections, the multi poll backend can have higher bandwidth than the same backend on the main branch (`9.68 > 5.87`), with more efficient CPU usage than the easy thread pool backend (`165.1 > 115.1`).

### DIFF
--- a/cpp/include/kvikio/detail/concurrent_request_limiter.hpp
+++ b/cpp/include/kvikio/detail/concurrent_request_limiter.hpp
@@ -1,0 +1,71 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <atomic>
+#include <cstddef>
+
+namespace kvikio::detail {
+
+/**
+ * @brief Non-blocking counting limiter that bounds the number of concurrent in-flight requests.
+ *
+ * Used by the `MULTI_POLL` remote I/O backend to cap the total number of HTTP range requests
+ * (libcurl easy handles) that are simultaneously attached to the reactors' multi handles, across
+ * all reactor threads. Without such a bound, one `pread()` of a large file fans out to one request
+ * per sub-range (e.g. thousands at once), which overwhelms the network path and collapses
+ * throughput. A single process-wide instance is owned by `MultiReactorPool`.
+ *
+ * The limiter never blocks. A caller that cannot acquire a slot is expected to defer its work and
+ * retry later, so the reactor thread is free to keep driving the requests it already admitted
+ * (which is what eventually frees slots). `try_acquire()` and `release()` are thread-safe and may
+ * be called from different threads, including a release from a completion callback running on
+ * another thread.
+ *
+ * A `_max` of 0 means unlimited. To keep the acquire/release pairing uniform at every call site,
+ * the limiter still counts in the unlimited case (`try_acquire()` always succeeds), so callers
+ * always pair a successful `try_acquire()` with exactly one `release()` regardless of
+ * configuration.
+ */
+class ConcurrentRequestLimiter {
+ public:
+  /**
+   * @brief Construct a limiter with the given ceiling.
+   *
+   * @param max_concurrent_requests Maximum number of slots that may be held at once. 0 means
+   * unlimited.
+   */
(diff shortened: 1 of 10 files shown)


cpp/include/kvikio/detail/concurrent_request_limiter.hpp
```diff
@@ -0,0 +1,68 @@
...
+   * @return `true` if a slot was reserved, and the caller must later call `release()` exactly once,
+   * or `false` if the limiter is already at its ceiling. Always returns `true` when unlimited.
+   */
+  [[nodiscard]] bool try_acquire() noexcept;
+
+  /**
+   * @brief Return one previously reserved slot. Must be paired one-to-one with a successful
+   * `try_acquire()`.
+   */
+  void release() noexcept;
```
kingcrimsontianyu, line 57: In this PR, `ConcurrentRequestLimiter` is created per reactor, and the `try_acquire()` and `release()` methods are called on the same reactor thread. But it does not hurt to generalize the feature by making these methods thread-safe.

cpp/include/kvikio/detail/multi_poll_reactor.hpp
```diff
@@ -107,8 +108,11 @@ class MultiPollReactor {
    * @param pool Non-owning back-pointer to the pool that owns this reactor. Used to observe and
    * propagate pool-wide death state. The pool must outlive the reactor, which is guaranteed because
    * the pool is a leaked singleton that owns this reactor by `unique_ptr`.
+   * @param max_concurrent_requests This reactor's private share of the total concurrent-request
+   * budget (the global cap divided across reactors). 0 means unlimited. Each reactor enforces its
+   * own share against its own inbox.
    */
-  explicit MultiPollReactor(MultiReactorPool* pool);
+  MultiPollReactor(MultiReactorPool* pool, std::size_t max_concurrent_requests);
```
madsbk: Consider using `std::optional<>` instead of zero as the `unlimited` value.

madsbk (approved):
Looks very good

cpp/include/kvikio/detail/multi_poll_reactor.hpp
kingcrimsontianyu: Thanks @madsbk for the quick review. Done as suggested.

PR #980, opened 2026-06-15
Handle zero-size remote reads

fallintoplace:
## Summary
- return early from synchronous remote reads when `size == 0`
- reject zero-size HTTP range construction in the shared range helper
- add a regression test that verifies `RemoteHandle::read(buf, 0, 0)` returns without configuring curl or a range request

## Notes
`RemoteHandle::pread()` already treated zero-size reads as a no-op, but the synchronous `read()` path went on to build a range string from `file_offset + size - 1`. For `size == 0`, that underflows and produces an invalid range.

## Validation
- `git diff --check`

C++ tests were not run locally because this machine does not have CUDA tooling/GPU access (`nvcc` and `nvidia-smi` are unavailable).


kingcrimsontianyu (approved):
Thanks for the improvement. Looks good to me.

PR #981, opened 2026-06-15
Validate remote read byte counts

fallintoplace:
## Summary
- validate successful easy-curl transfers against the callback byte count
- reset callback byte accounting before each easy-curl retry attempt
- reset the existing device bounce state before retrying device reads
- fail easy and multi-poll remote reads when the callback receives fewer bytes than requested

## Why
A remote read could return success after libcurl completed even if the write callback received fewer bytes than requested. The easy retry path also reused the same callback offset between attempts, so a partial timed-out attempt could make the next attempt append at the wrong offset.

This keeps the existing host and device callback paths, restarts per-attempt byte state on retries, and rejects successful transfers that did not deliver the requested byte count.

## Validation
- `pre-commit run --files cpp/include/kvikio/detail/remote_callback.hpp cpp/src/detail/remote_callback.cpp cpp/include/kvikio/shim/libcurl.hpp cpp/src/shim/libcurl.cpp cpp/src/remote_handle.cpp cpp/src/detail/multi_poll_reactor.cpp python/kvikio/tests/test_http_io.py`
- `python3 -m py_compile python/kvikio/tests/test_http_io.py`
- `git diff --check`

Not run locally: the KvikIO Python/C++ test suite, because this checkout does not have an importable `kvikio` package and CUDA tooling/GPU access are unavailable (`nvcc` and `nvidia-smi` are not present).


cpp/src/remote_handle.cpp
```diff
@@ -774,6 +765,23 @@ std::size_t RemoteHandle::read(void* buf, std::size_t size, std::size_t file_off
...
+    std::size_t copied = 0;
+    while (copied < size) {
+      std::size_t const chunk_size = std::min(bounce_buffer.size(), size - copied);
+      std::memcpy(bounce_buffer.get(), staging_buffer.data() + copied, chunk_size);
+      KVIKIO_CUDA_DRIVER_TRY(cudaAPI::cuda_memcpy_async(
+        dev_ptr + copied, convert_void2deviceptr(bounce_buffer.get()), chunk_size, stream));
+      KVIKIO_CUDA_DRIVER_TRY(cudaAPI::instance().StreamSynchronize(stream));
+      copied += chunk_size;
+    }
+  }
```
wence-: This change seems to needlessly introduce another level of intermediate bounce buffer. Why is it not possible to just use the existing setup and check the relevant error condition as part of curl.perform?

wence- (changes requested):

cpp/src/remote_handle.cpp
fallintoplace: Thank you for your attention, I have amended.

cpp/include/kvikio/detail/remote_callback.hpp
```diff
@@ -34,6 +34,26 @@ struct CallbackContext {
...
 
+/**
+ * @brief Reset per-attempt byte accounting before retrying a transfer.
+ */
+void reset_callback_context(CallbackContext& ctx) noexcept;
+
+/**
+ * @brief Whether a transfer delivered exactly the requested number of bytes.
+ */
+[[nodiscard]] bool callback_context_complete(CallbackContext const& ctx) noexcept;
```
vuule:
```suggestion
[[nodiscard]] bool is_callback_context_complete(CallbackContext const& ctx) noexcept;
```

vuule (commented):
C++ changes look good, just one suggestion

vuule:
@fallintoplace the new tests fail: https://github.com/rapidsai/kvikio/actions/runs/30305017333/job/90111441444?pr=981

fallintoplace:
I will push ASAP. @vuule

cpp/src/remote_handle.cpp
```diff
@@ -780,17 +789,28 @@ std::size_t RemoteHandle::read(void* buf, std::size_t size, std::size_t file_off
...
     } else {
       PushAndPopContext c(get_context_from_pointer(buf));
       // We use a bounce buffer to avoid many small memory copies to device. Libcurl has a
       // maximum chunk size of 16kb (`CURL_MAX_WRITE_SIZE`) but chunks are often much smaller.
       detail::BounceBufferH2D bounce_buffer(detail::StreamCachePerThreadAndContext::get(), buf);
       ctx.bounce_buffer = &bounce_buffer;
-      curl.perform();
+      curl.set_before_perform_attempt([&ctx, &bounce_buffer] {
+        detail::reset_callback_context(ctx);
+        bounce_buffer.reset();
```
wence-, line 806: I am concerned here. This is before we're calling `curl.perform()`. How can we get a bounce buffer that is not reset?

cpp/src/remote_handle.cpp
```diff
@@ -780,17 +789,28 @@ std::size_t RemoteHandle::read(void* buf, std::size_t size, std::size_t file_off
...
+      });
+      try {
+        curl.perform();
+      } catch (...) {
+        bounce_buffer.reset();
+        throw;
```
wence-, line 812: OK, so if we get an exception then we should reset the bounce buffer before returning it to the cache?

It seems like that should happen in the dtor for `bounce_buffer`. Rather than having to remember to do this manually.

cpp/src/remote_handle.cpp
```diff
@@ -780,17 +789,28 @@ std::size_t RemoteHandle::read(void* buf, std::size_t size, std::size_t file_off
   }
   detail::CallbackContext ctx{buf, size};
   curl.setopt(CURLOPT_WRITEDATA, &ctx);
+  curl.set_after_successful_perform([&ctx] { detail::expect_callback_context_complete(ctx); });
 
   try {
     if (is_host_mem) {
+      curl.set_before_perform_attempt([&ctx] { detail::reset_callback_context(ctx); });
```
wence-, line 796: I kind of think we should just inline the definition of these functions here. I don't think it buys us anything to introduce these detail functions.

Also, again, how are we managing to make a `CallbackContext` that is not in a reset state before the `curl.perform()` call? That seems like a bigger issue that we shouldn't paper over by calling `reset` before `perform`.

wence- (changes requested):

cpp/src/remote_handle.cpp
fallintoplace: Thanks, agreed. I updated this:

CallbackContext starts in the reset state by default.
Resetting now happens only through curl.perform(on_retry), before a retry, not before the first attempt.
The byte-count checks are inlined in the easy and multi-poll paths.
The device bounce buffer is reset on failure before its destructor runs, so partial data is not flushed.
I also removed the extra callback setter/helper functions.

PR #982, opened 2026-06-22
Add APIs for inferring remote endpoint type from URL.

TomAugspurger:
In https://github.com/rapidsai/cudf/pull/22739, cudf / cudf-polars needed to infer the remote endpoint type kvikio would use for some URL, *without* making the HTTP request that'd come from `RemoteFile::open` with the default `AUTO` mode.

To do this, I've added a new public `infer_remote_endpoint_type` method, along with a python binding.

### DIFF
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -10,6 +10,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <vector>
 
 #include <kvikio/defaults.hpp>
 #include <kvikio/error.hpp>
@@ -329,6 +330,17 @@ class S3EndpointWithPresignedUrl : public RemoteEndpoint {
    */
   static bool is_url_valid(std::string const& url) noexcept;
 };
+
+/**
+ * @brief Infer remote endpoint type from URL.
+ *
+ * This function follows the same endpoint-selection order as `RemoteHandle::open()` in
+ * `RemoteEndpointType::AUTO` mode, but only infers the endpoint type and does not create a handle.
+ *
+ * @param url The URL of the remote file.
+ * @return The inferred endpoint type.
+ */
+RemoteEndpointType infer_remote_endpoint_type(std::string url);
 
 /**
  * @brief Handle of remote file.
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -340,7 +340,7 @@ class S3EndpointWithPresignedUrl : public RemoteEndpoint {
  * @param url The URL of the remote file.
  * @return The inferred endpoint type.
  */
-RemoteEndpointType infer_remote_endpoint_type(std::string url);
+RemoteEndpointType infer_remote_endpoint_type(std::string const& url);
 
 /**
  * @brief Handle of remote file.
@@ -428,7 +428,7 @@ class RemoteHandle {
    *   );
    *   @endcode
    */
-  static RemoteHandle open(std::string url,
+  static RemoteHandle open(std::string const& url,
                            RemoteEndpointType remote_endpoint_type = RemoteEndpointType::AUTO,
                            std::optional<std::vector<RemoteEndpointType>> allow_list = std::nullopt,
                            std::optional<std::size_t> nbytes = std::nullopt);
--- a/cpp/include/kvikio/remote_handle.hpp
+++ b/cpp/include/kvikio/remote_handle.hpp
@@ -336,6 +336,9 @@ class S3EndpointWithPresignedUrl : public RemoteEndpoint {
  *
  * This function follows the same endpoint-selection order as `RemoteHandle::open()` in
  * `RemoteEndpointType::AUTO` mode, but only infers the endpoint type and does not create a handle.
+ * Note that this function will not return `RemoteEndpointType::S3_PUBLIC`, because disambiguating
+ * between a URL that's accessible only with authorization or only anonymously is not possible
+ * without making an HTTP request.
  *
  * @param url The URL of the remote file.
  * @return The inferred endpoint type.
--- a/cpp/src/remote_handle.cpp
+++ b/cpp/src/remote_handle.cpp
@@ -14,6 +14,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <tuple>
 
 #include <kvikio/bounce_buffer.hpp>
 #include <kvikio/defaults.hpp>
@@ -276,15 +277,10 @@ std::unique_ptr<RemoteEndpoint> create_endpoint_from_type(std::string const& url
   }
 }
 
-struct InferredEndpoint {
-  RemoteEndpointType remote_endpoint_type{};
-  std::unique_ptr<RemoteEndpoint> endpoint;
-  std::optional<std::size_t> probed_nbytes;
-};
-
-InferredEndpoint infer_endpoint_impl(std::string const& url,
-                                     std::vector<RemoteEndpointType> const& allow_list,
-                                     bool probe_s3_connectivity)
+std::pair<std::unique_ptr<RemoteEndpoint>, std::optional<std::size_t>> infer_endpoint_impl(
+  std::string const& url,
+  std::vector<RemoteEndpointType> const& allow_list,
+  bool probe_s3_connectivity)
 {
   auto const scheme =
     detail::UrlParser::extract_component(url, CURLUPART_SCHEME, CURLU_NON_SUPPORT_SCHEME);
@@ -301,15 +297,14 @@ InferredEndpoint infer_endpoint_impl(std::string const& url,
         // RemoteHandle::open to avoid a second HEAD request.
         probed_nbytes = endpoint->get_file_size();
       }
-      return InferredEndpoint{type, std::move(endpoint), probed_nbytes};
+      return {std::move(endpoint), probed_nbytes};
     } catch (...) {
       // If the credential-based S3 endpoint cannot be used to access the URL, try using
       // S3 public endpoint instead when it is in the allowlist.
       if (type == RemoteEndpointType::S3 &&
           std::find(allow_list.begin(), allow_list.end(), RemoteEndpointType::S3_PUBLIC) !=
             allow_list.end()) {
-        return InferredEndpoint{
-          RemoteEndpointType::S3_PUBLIC, std::make_unique<S3PublicEndpoint>(url), std::nullopt};
+        return {std::make_unique<S3PublicEndpoint>(url), std::nullopt};
       }
       throw;
     }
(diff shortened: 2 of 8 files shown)


TomAugspurger:
/ok to test https://github.com/rapidsai/kvikio/pull/982/commits/2ca7f251bf7c48650373494ecb9050d3b100fa6b

cpp/src/remote_handle.cpp
```diff
@@ -231,6 +232,89 @@ std::string encode_special_chars_in_path(std::string const& url)
...
+  for (auto const& type : allow_list) {
+    try {
+      auto endpoint = create_endpoint_from_type(url, scheme.value(), type);
+      if (endpoint == nullptr) { continue; }
+
+      std::optional<std::size_t> probed_nbytes = std::nullopt;
+      if (probe_s3_connectivity && type == RemoteEndpointType::S3) {
+        // Check connectivity for the credential-based S3 endpoint and reuse this size in
+        // RemoteHandle::open to avoid a second HEAD request.
+        probed_nbytes = endpoint->get_file_size();
```
kingcrimsontianyu, line 298: Since the `open()` function is modified again, can we just cache the file size inside the endpoint as I suggested? That is, the first time the endpoint's file size gets called, HEAD (or whatever method necessary for the endpoint in question) is performed, result cached in a `std::optional<std::size_t> _file_size`, and subsequent file size query just reuses the cached value.

This way, this PR alone will reduce the cudf-polars per-file HEAD request to 1, if I'm not missing anything.

TomAugspurger: Personally, I'm not comfortable making something like `size` part of the data model for `RemoteHandle` without also including something like an etag to ensure that we aren't operating on a stale file. By using `size`, cudf-polars is accepting that risk, but it's unavoidable for now; and it's at least no than what would currently happen if a file changes mid-query. But putting it into the kvikio API is a bit riskier I think.

And, at least from the cudf-polars side, we'll temporarily use kvikio to get the file size exactly once per file. But in the medium term we'll be able to get that information from cudf-polars; either way, we'll be able to provide it the size in all subsequent operations.

Finally, I don't think this would directly help cudf-polars since in our current design we don't store `RemoteHandle` instances. We'll just store the `plc.io.types.SourceInfo` and `FileMetadata`.

cpp/src/remote_handle.cpp
```diff
@@ -575,87 +659,27 @@ bool S3EndpointWithPresignedUrl::is_url_valid(std::string const& url) noexcept
   }
 }
 
+RemoteEndpointType infer_remote_endpoint_type(std::string url)
```
madsbk:
```suggestion
RemoteEndpointType infer_remote_endpoint_type(std::string const& url)
```

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -330,6 +331,17 @@ class S3EndpointWithPresignedUrl : public RemoteEndpoint {
...
+/**
+ * @brief Infer remote endpoint type from URL.
+ *
+ * This function follows the same endpoint-selection order as `RemoteHandle::open()` in
+ * `RemoteEndpointType::AUTO` mode, but only infers the endpoint type and does not create a handle.
+ *
+ * @param url The URL of the remote file.
+ * @return The inferred endpoint type.
+ */
+RemoteEndpointType infer_remote_endpoint_type(std::string url);
```
```suggestion
RemoteEndpointType infer_remote_endpoint_type(std::string const& url);
```

madsbk (approved):
Looks good

cpp/src/remote_handle.cpp
TomAugspurger: Does this sound OK @kingcrimsontianyu? I'd like to merge this today if possible, so we can use it in https://github.com/rapidsai/cudf/pull/22739/

cpp/src/remote_handle.cpp
```diff
@@ -231,6 +232,89 @@ std::string encode_special_chars_in_path(std::string const& url)
...
 }
+
+std::vector<RemoteEndpointType> get_default_allow_list()
+{
+  return {RemoteEndpointType::S3,
+          RemoteEndpointType::S3_PUBLIC,
+          RemoteEndpointType::S3_PRESIGNED_URL,
+          RemoteEndpointType::WEBHDFS,
+          RemoteEndpointType::HTTP};
+}
```
kingcrimsontianyu:
```suggestion
std::vector<RemoteEndpointType> const& get_default_allow_list()
{
  static std::vector const res = {RemoteEndpointType::S3,
          RemoteEndpointType::S3_PUBLIC,
          RemoteEndpointType::S3_PRESIGNED_URL,
          RemoteEndpointType::WEBHDFS,
          RemoteEndpointType::HTTP};
  return res;
}
```
Typed in a browser. May need reformatting. `res` type may just be `std::vector` for simplicity, due to [CTAD](https://en.cppreference.com/cpp/language/class_template_argument_deduction).

kingcrimsontianyu: Also since `get_default_allow_list` is not a public API and only used in this source file, we may put it in an unnamed namespace:
```cpp
namespace {
  std::vector<RemoteEndpointType> const& get_default_allow_list() {
  ...
  }
}
```

TomAugspurger: IIUC, we're already inside a `namespace {` here.

cpp/src/remote_handle.cpp
```diff
@@ -231,6 +232,90 @@ std::string encode_special_chars_in_path(std::string const& url)
...
+
+    default: return nullptr;
+  }
+}
+
+struct InferredEndpoint {
+  RemoteEndpointType remote_endpoint_type{};
+  std::unique_ptr<RemoteEndpoint> endpoint;
+  std::optional<std::size_t> probed_nbytes;
+};
```
kingcrimsontianyu: The data member `remote_endpoint_type` ~~is unused, and it~~ is a duplicate since `endpoint->remote_endpoint_type()` can retrieve the type already. So I think we may remove this struct, and let `infer_endpoint_impl` return a `std::pair`.

kingcrimsontianyu: i.e.
```cpp
auto infer_endpoint_impl(...)
```
and in this function we return `std::pair{endpoint, probed_nbytes}`.

cpp/include/kvikio/remote_handle.hpp
```diff
@@ -330,6 +331,17 @@ class S3EndpointWithPresignedUrl : public RemoteEndpoint {
...
 
+/**
+ * @brief Infer remote endpoint type from URL.
+ *
+ * This function follows the same endpoint-selection order as `RemoteHandle::open()` in
+ * `RemoteEndpointType::AUTO` mode, but only infers the endpoint type and does not create a handle.
+ *
+ * @param url The URL of the remote file.
+ * @return The inferred endpoint type.
+ */
```
kingcrimsontianyu, line 345: I think we need to document the difference between this utility function and the `open` function, that given a URL valid for both S3 private and public endpoints, this function will always return the private endpoint type, and that it is not possible for us to disambiguate S3 private and public endpoints from a URL alone. 
Then in the cudf PR https://github.com/rapidsai/cudf/pull/22739, we may document that known file size + URL without S3 credential is not supported.

cpp/src/remote_handle.cpp
```diff
@@ -4,6 +4,7 @@
  */
 
 #include <algorithm>
+#include <array>
```
kingcrimsontianyu, line 7: Good catch

cpp/src/remote_handle.cpp
kingcrimsontianyu: I misread. `remote_endpoint_type` is used, but it can be replaced by the member function, i.e. in `infer_remote_endpoint_type`:
```cpp
auto [endpoint, _] = infer_endpoint_impl(url, get_default_allow_list(), false);
return endpoint->remote_endpoint_type();
```

TomAugspurger: Done in 37b9ddbdf2cd690bcfd63ec8bfff5fffb354e39d. Codex did some strange things with `std::void` since we don't use the second member of the returned tuple. I had it use `std::ignore`, which also looks kinda weird. I didn't want it using `[[maybe_unused]]`, since that applied to the whole tuple and we do use the first member. I dunno.

cpp/include/kvikio/remote_handle.hpp
TomAugspurger: I added a note in b752f5abe158ac792328d9d828053f44d7a7c552.

But I'd like to get away from "public" vs. "private" URLs entirely. It'd much prefer to just deal with some URL and an authorization method (possibly a chained authorization method that tries multiple).

cpp/src/remote_handle.cpp
```diff
@@ -575,87 +655,29 @@ bool S3EndpointWithPresignedUrl::is_url_valid(std::string const& url) noexcept
   }
 }
 
-RemoteHandle RemoteHandle::open(std::string url,
+RemoteEndpointType infer_remote_endpoint_type(std::string const& url)
+{
+  KVIKIO_NVTX_FUNC_RANGE();
+  std::unique_ptr<RemoteEndpoint> endpoint;
+  std::tie(endpoint, std::ignore) = infer_endpoint_impl(url, get_default_allow_list(), false);
```
kingcrimsontianyu, line 662: Nit: using `std::tie()` is correct, but it requires first default-constructs a `unique_ptr` of `endpoint`, and then move-assigns the returned endpoint to `endpoint`.

Using _structured binding_ is cleaner because of the guaranteed copy/move elision, i.e. `endpoint` _is_ the returned result, without move assignment.

If we worry about compile warning on unused variables, we may just do the following:
```cpp
  [[maybe_unused]] auto [endpoint, ignored] =
    infer_endpoint_impl(url, get_default_allow_list(), false);
```

PS: The one I suggested earlier would be the recommended way in C++26 where `_` is specifically used to absorb unused variables. But this _might_ trigger clang error now since we are using C++20.
```cpp
auto [endpoint, _] = infer_endpoint_impl(url, get_default_allow_list(), false);
```

cpp/src/remote_handle.cpp
kingcrimsontianyu: It is valid to use a variable that is marked `[[maybe_unused]]`.

kingcrimsontianyu (approved):
Lgtm!
Nit: suggested using structured binding in place of tie.

cpp/src/remote_handle.cpp
TomAugspurger: Yeah, it seemed weird to say `[[maybe_unused]]` for a tuple, where you definitely do use one of the fields though. But I can do whatever you want here.

TomAugspurger: Just saw https://github.com/rapidsai/kvikio/pull/982#discussion_r3463959618. I definitely prefer the C++26 style :)

That seemed to build locally without an error / warning so let's try that.

TomAugspurger: That passed CI, thanks!

cpp/src/remote_handle.cpp
TomAugspurger: Ah, I messed up how I applied this in https://github.com/rapidsai/kvikio/pull/982/changes/e91aed6bbe31ed0a599582262410cbb3fd725394. I'll make another PR quick.

TomAugspurger: https://github.com/rapidsai/kvikio/pull/990 has that.

PR #983, opened 2026-06-23
Close resources before move assignment

fallintoplace:
## Summary
- close existing `FileWrapper` and `CUFileHandleWrapper` resources before move assignment takes ownership
- close an existing `FileHandle` before move assignment so cuFile deregistration happens before fd teardown
- add fd-based regression coverage for wrapper and file-handle move assignment

## Validation
- `pre-commit run --files cpp/src/file_utils.cpp cpp/src/file_handle.cpp cpp/tests/test_file_utils.cpp`
- `git diff --check`

Targeted CTest was not run locally because no installed/devcontainer KvikIO test build was present at the paths searched by `ci/run_ctests.sh -R FILE_UTILS_TEST`.

### DIFF
--- a/cpp/src/file_handle.cpp
+++ b/cpp/src/file_handle.cpp
@@ -115,13 +115,16 @@ FileHandle::FileHandle(FileHandle&& o) noexcept
 
 FileHandle& FileHandle::operator=(FileHandle&& o) noexcept
 {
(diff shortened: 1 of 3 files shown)


kingcrimsontianyu (approved):
Good catch. Thanks for the fix.

PR #985, opened 2026-06-23
Avoid overflow in remote read bounds checks

fallintoplace:
## Summary
- avoid overflow in remote read bounds checks by comparing `file_offset` before subtracting from `_nbytes`
- use the same bounds helper in direct `RemoteHandle::read()` and the `MULTI_POLL` `pread()` setup path
- add a regression test for a wrapping `file_offset + size` range that must fail before range setup

## Validation
- `git diff --check`


vyasr:
Needs style fixes before CI can run properly.

vuule (approved):

PR #986, opened 2026-06-23
Check remote range end arithmetic

fallintoplace:
## Summary
- reject remote range requests whose inclusive end offset would overflow
- compute the `CURLOPT_RANGE` end offset only after the checked arithmetic passes
- add a regression test that exercises the shared HTTP/S3 range setup helper without performing a transfer

## Validation
- `git diff --check`

Not run locally: `ci/run_ctests.sh -R REMOTE_HANDLE_TEST` because no installed/devcontainer KvikIO test build was present at the paths searched by the script.


cpp/src/remote_handle.cpp
```diff
@@ -193,8 +194,11 @@ std::size_t get_file_size_using_head_impl(RemoteEndpoint& endpoint, std::string
 void setup_range_request_impl(CurlHandle& curl, std::size_t file_offset, std::size_t size)
 {
   KVIKIO_EXPECT(size > 0, "cannot create a zero-size range request", std::invalid_argument);
-  std::string const byte_range =
-    std::to_string(file_offset) + "-" + std::to_string(file_offset + size - 1);
+  KVIKIO_EXPECT(size <= std::numeric_limits<std::size_t>::max() - file_offset,
```
vuule, line 197: FWIW, the check is off by one. size + offset can be max + 1, while having all bytes in addressable space.

```suggestion
  KVIKIO_EXPECT(size - 1 <= std::numeric_limits<std::size_t>::max() - file_offset,
```

cpp/src/remote_handle.cpp
```diff
@@ -193,8 +194,11 @@ std::size_t get_file_size_using_head_impl(RemoteEndpoint& endpoint, std::string
...
+                "cannot create a range request with overflowing end offset",
```
vuule, line 198: Can this name the offending values, the way `expect_read_in_bounds` formats `file_offset` and `size` into its message further down?

cpp/tests/test_remote_handle.cpp
```diff
@@ -189,6 +190,17 @@ TEST_F(RemoteHandleTest, read_overflowing_range_throws_without_range_request)
...
 }
 
+TEST_F(RemoteHandleTest, range_request_rejects_overflowing_end_offset)
+{
+  kvikio::HttpEndpoint endpoint{"http://example.com/test"};
+  auto curl = create_curl_handle();
+
+  auto constexpr file_offset = std::numeric_limits<std::size_t>::max() - 1;
+  auto constexpr size        = std::size_t{4};
+  EXPECT_THAT([&] { endpoint.setup_range_request(curl, file_offset, size); },
```
vuule, line 200: This case overflows by a wide margin, so it passes whether the guard is exact or a unit strict, and the boundary stays uncovered either way. Would be better to test the two edge cases.

vuule (commented):
few small suggestions

PR #990, opened 2026-06-24
Fix unused variable handling in `infer_remote_endpoint_type`

TomAugspurger:
This fixes how we handle the unused variable in the call to `infer_endpoint_impl`, mentioned in https://github.com/rapidsai/kvikio/pull/982/changes#r3463959618

### DIFF
--- a/cpp/src/remote_handle.cpp
+++ b/cpp/src/remote_handle.cpp
@@ -14,7 +14,6 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
-#include <tuple>
 
 #include <kvikio/bounce_buffer.hpp>
 #include <kvikio/defaults.hpp>
--- a/cpp/src/remote_handle.cpp
+++ b/cpp/src/remote_handle.cpp
@@ -658,8 +658,7 @@ bool S3EndpointWithPresignedUrl::is_url_valid(std::string const& url) noexcept
 RemoteEndpointType infer_remote_endpoint_type(std::string const& url)
 {
   KVIKIO_NVTX_FUNC_RANGE();
-  auto [endpoint, _]              = infer_endpoint_impl(url, get_default_allow_list(), false);
-  std::tie(endpoint, std::ignore) = infer_endpoint_impl(url, get_default_allow_list(), false);
+  auto [endpoint, _] = infer_endpoint_impl(url, get_default_allow_list(), false);
   return endpoint->remote_endpoint_type();
 }
 


cpp/src/remote_handle.cpp
```diff
@@ -658,8 +658,7 @@ bool S3EndpointWithPresignedUrl::is_url_valid(std::string const& url) noexcept
 RemoteEndpointType infer_remote_endpoint_type(std::string const& url)
 {
   KVIKIO_NVTX_FUNC_RANGE();
-  auto [endpoint, _]              = infer_endpoint_impl(url, get_default_allow_list(), false);
-  std::tie(endpoint, std::ignore) = infer_endpoint_impl(url, get_default_allow_list(), false);
+  auto [endpoint, _] = infer_endpoint_impl(url, get_default_allow_list(), false);
```
kingcrimsontianyu, line 660: While this is the right approach in C++26, considering we are building KvikIO with C++20 standard, the `[[maybe_unused]]` should be added, because the unused structured binding `_` may trigger compile warning.
It's a bit surprising the compiler hasn't complaint. Perhaps `-Wunused` is missing in the CMake file.

TomAugspurger: Yeah... our CMakeLists doesn't have that set. I added this

```
set(KVIKIO_CXX_FLAGS "")
if(CMAKE_COMPILER_IS_GNUCXX)
  list(APPEND KVIKIO_CXX_FLAGS -Wall -Werror -Wno-unknown-pragmas -Wno-error=deprecated-declarations)
endif()
```

to kvikio/cpp/CMakeLists.txt.

Interestingly, the `auto [endpoint, _] = infer_endpoint_impl(url, get_default_allow_list(), false);` isn't flagged, so maybe we're OK there?

There are other errors, so I'll open an issue for that.

TomAugspurger: https://github.com/rapidsai/kvikio/issues/991 has that.

kingcrimsontianyu: This is more of a nit. The standard states:

>Recommended practice: ... For a structured binding declaration not marked maybe_unused, implementations should not emit such a warning unless all of its structured bindings are unused.

Compilers do not have to follow the _recommended practice_, and `[[maybe_unused]]` ensure compilers that don't, won't complain about `_` under C++20. I tested GCC and Clang (that's what we mainly care about), and they do follow this recommended practice well. So it's fine to not have the attribute here.

kingcrimsontianyu (approved):

PR #992, opened 2026-06-24
java: use GCS mirror for Maven, other small changes

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/297

RAPIDS projects using NVIDIA-hosted GitHub Actions runners have been getting rate-limited by Maven Central. Similar to https://github.com/NVIDIA/cuvs/pull/2253, this proposes fixing that by using the same read-only Maven mirror that Apache Orc, Lucene, Spark and others use in their builds (https://storage-download.googleapis.com/maven-central/index.html).

Other changes:

* adding default `mvn` options to request packages more slowly and wait longer between retries
* updating RAPIDS `pre-commit` hooks
* adding `check-xml` hook to validate that `pom.xml` are valid XML docs
* reformatting READMEs per https://yihui.org/en/2021/06/markdown-breath/
* other minor `pom.xml` maintenance (see comments)

### DIFF
--- a/java/.mvn/maven.config
+++ b/java/.mvn/maven.config
@@ -1,0 +1,6 @@
+-e
+-B
+-Daether.connector.basic.downstreamThreads=1
+-Daether.transport.http.retryHandler.count=5
+-Daether.transport.http.retryHandler.interval=10000
+-Dmaven.wagon.http.retryHandler.count=5
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
+SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 SPDX-License-Identifier: Apache-2.0
 -->
 <project  xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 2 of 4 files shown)


java/.mvn/maven.config
```diff
@@ -0,0 +1,6 @@
+-e
+-B
+-Daether.connector.basic.downstreamThreads=1
+-Daether.transport.http.retryHandler.count=5
+-Daether.transport.http.retryHandler.interval=10000
+-Dmaven.wagon.http.retryHandler.count=5
```
jameslamb, line 6: See https://github.com/NVIDIA/cuvs/pull/2253#discussion_r3462121538 for an explanation of these flags.

I believe comments are not allowed in this file because its contents are passed literally to `mvn`.

java/pom.xml
```diff
@@ -58,16 +112,9 @@ SPDX-License-Identifier: Apache-2.0
...
-        <plugin>
-          <artifactId>maven-exec-plugin</artifactId>
-          <version>1.6.0</version>
-        </plugin>
         <plugin>
           <artifactId>maven-clean-plugin</artifactId>
           <version>3.1.0</version>
-          <configuration>
-            <createDirs>true</createDirs>
-          </configuration>
```
jameslamb, line 70: `createDirs` is not a valid property for `maven-clean-plugin`: https://github.com/search?q=repo%3Aapache%2Fmaven-clean-plugin%20%22createDirs%22&type=code

java/pom.xml
```diff
@@ -88,7 +135,7 @@ SPDX-License-Identifier: Apache-2.0
             <dependency>
               <groupId>org.junit.jupiter</groupId>
               <artifactId>junit-jupiter-engine</artifactId>
-              <version>5.4.2</version>
+              <version>${junit.version}</version>
```
jameslamb, line 138: This version is stored a few lines up:

https://github.com/rapidsai/kvikio/blob/37790a19b4d6cfce4c1f91049c3bbe2565d033cd/java/pom.xml#L26

Reusing that value reduces the risk of inconsistencies.

java/pom.xml
```diff
@@ -104,14 +151,6 @@ SPDX-License-Identifier: Apache-2.0
...
           <version>2.8.2</version>
         </plugin>
-        <plugin>
-          <artifactId>maven-site-plugin</artifactId>
-          <version>3.7.1</version>
-        </plugin>
-        <plugin>
-          <artifactId>maven-project-info-reports-plugin</artifactId>
-          <version>3.0.0</version>
-        </plugin>
```
jameslamb, line 114: These don't have entries in `<plugins>` so don't need entries in `<pluginManagement>`.

java/.mvn/maven.config
gforsyth: I don't know if it's worth a CI re-run, but one option (maybe) would be to document the options in a `README` inside the `.mvn` directory.  But that shouldn't block this.

gforsyth (approved):
Just the one thought, but looks good.  Thanks @jameslamb !

java/.mvn/maven.config
jameslamb: Yeah fair. I went back and forth on that and settled on not doing it because I thought the git blame pointing back to this PR + the mostly informative names was probably enough to communicate the intent.

I'm gonna treat that "I don't know" and "(maybe)" as you also being unsure about how to handle this 😂 

Think I'll just merge it as-is.

PR #999, opened 2026-07-06
Develop an Nsight System plugin to profile network bandwidth

kingcrimsontianyu:
When developing and optimizing remote I/O, a commonly asked question is: how do I know the achieved bandwidth of my application? While nsys does offer network-related features, limitations exist:

- `--nic-metrics`: Requires OFED/InfiniBand dependency that cannot be fulfilled on many systems, including `g7e.8xlarge`.
- `--enable=network_interface`: Shows raw counter instead of bandwidth value. Does not register signals and therefore always returns non-zero exit code.

This PR develops an nsys plugin to allow flexible, generic NIC bandwidth profiling without the above limitations.


kingcrimsontianyu:
## Sample UI

Below is an example of the NIC bandwidth profile in nsys UI.

image

kingcrimsontianyu:
## Sample documentation

Screenshot 2026-07-07 at 01-43-09 Profiling — kvikio 26 08 00 documentation

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
```diff
@@ -0,0 +1,381 @@
...
+  NicCounterReader _reader;
+  nvtxDomainHandle_t _domain;
+  std::vector<std::uint64_t> _counter_ids;
+  std::uint64_t _total_counter_id{0};
+};
+
+}  // namespace
+
+// C language linkage (extern) to be standard conformant.
+// Also internal linkage (static) as a good practice.
```
wence-:
```suggestion
```

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
```diff
@@ -0,0 +1,381 @@
...
+ *
+ * The domain is separate from libkvikio's own domain because the plugin runs its own process.
+ */
+struct kvikio_nic_domain {
+  static constexpr char const* name{"KvikIO NIC"};
+};
+
+// Set from a signal handler for a clean shutdown.
+volatile std::sig_atomic_t g_stop = 0;
+
```
wence-, line 59: I think this can be:

```
std::atomic_flag g_stop = ATOMIC_FLAG_INIT;
```

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
```diff
@@ -0,0 +1,381 @@
...
+}  // namespace
+
+// C language linkage (extern) to be standard conformant.
+// Also internal linkage (static) as a good practice.
+// nsys stops the collector by sending SIGTERM (then SIGKILL after a grace period). Catch it so the
+// loop breaks and the process exits cleanly with code 0 instead of an abnormal termination. Also
+// catch SIGINT to have the same clean exit for Ctrl-C.
+extern "C" {
+static void kvikio_nic_handle_signal(int /*signum*/) { g_stop = 1; }
```
wence-: And this:

```
g_stop.test_and_set();

File: cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp [hunk 4]
```diff
@@ -0,0 +1,381 @@
...
+    }
+    _total_counter_id = register_counter(_domain, "total", schema_id);
+  }
+
+  /**
+   * @brief Run the sampling loop until @p stop becomes nonzero.
+   *
+   * @param stop Stop flag, set asynchronously by the signal handler.
+   */
+  void run(volatile std::sig_atomic_t const& stop) const
```
wence-: And this `void run() const`

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
```diff
@@ -0,0 +1,381 @@
...
+                 "kvikio_nic: sampling %zu interface(s) every %lld us\n",
+                 interfaces.size(),
+                 static_cast<long long>(_interval.count()));
+
+    using clock        = std::chrono::steady_clock;
+    auto prev_counters = _reader.read();
+    auto prev_time     = clock::now();
+    auto next_deadline = prev_time;
+
+    while (stop == 0) {
```
wence-:
```
while (!g_stop.test()) {
```

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
```diff
@@ -0,0 +1,381 @@
...
+  if (interfaces.empty()) {
+    std::fprintf(stderr, "kvikio_nic: no matching network interfaces to monitor.\n");
+    return constants::exit_no_interfaces;
+  }
+
+  std::signal(SIGTERM, kvikio_nic_handle_signal);
+  std::signal(SIGINT, kvikio_nic_handle_signal);
+
+  BandwidthCollector const collector{std::move(interfaces), config.interval};
+  collector.run(g_stop);
```
wence-:
```suggestion
  collector.run();
```

wence- (commented):
flushing some small comments

kingcrimsontianyu:
@wence- Thanks! These are great suggestions. I didn't know the modern C++ way of doing signal handling, so this is a good learning experience.

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
kingcrimsontianyu: Removed

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
kingcrimsontianyu: [Ref](https://en.cppreference.com/cpp/atomic/ATOMIC_FLAG_INIT) says the following. So I've left it default constructed.
> This macro is no longer needed since default constructor of std::atomic_flag initializes it to clear state. It is kept for the compatibility with C.

kingcrimsontianyu: Great to learn about `std::atomic_flag`.

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
kingcrimsontianyu: Done. Passed the relaxed memory ordering.

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
kingcrimsontianyu: Done. Updated the comment too.

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
kingcrimsontianyu: Done, and passed the relaxed memory ordering.

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
kingcrimsontianyu: Done.

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
```diff
@@ -0,0 +1,381 @@
...
+{
+  if (!config.device_filter.has_value()) { return default_interfaces(); }
+  std::vector<std::string> result;
+  std::error_code ec;
+  std::filesystem::directory_iterator it{
+    std::filesystem::path{kvikio::nsys_plugin::constants::sysfs_net}, ec};
+  if (ec) { return result; }
+  for (auto const& entry : it) {
+    auto name = entry.path().filename().string();
+    if (std::regex_match(name, config.device_filter.value())) { result.push_back(std::move(name)); }
```
madsbk: Consider `regex_search` or state that this "must match the entire interface name" in nsys-plugin.yaml, print_help_and_exit, and profiling.rst.

cpp/CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ option(KvikIO_BUILD_BENCHMARKS "Configure CMake to build benchmarks" OFF)
 option(KvikIO_BUILD_EXAMPLES "Configure CMake to build examples" ON)
 option(KvikIO_BUILD_TESTS "Configure CMake to build tests" ON)
 option(KvikIO_REMOTE_SUPPORT "Configure CMake to build with remote IO support" ON)
+option(KvikIO_BUILD_NSYS_PLUGIN "Configure CMake to build the Nsight Systems NIC plugin" ON)
```
madsbk, line 55: Do we want to build this by default? I guess, yes?

madsbk (approved):
LGTM, thanks @kingcrimsontianyu

cpp/nsys_plugins/nic/kvikio_nic_nsys_plugin.cpp
kingcrimsontianyu: Good idea. I'm sticking to the `regex_match` and have improved the doc and error message.

cpp/CMakeLists.txt
kingcrimsontianyu: Yes. I hope to have it built by default

python/libkvikio/libkvikio/nsys.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+
+
+def nsys_plugin_search_dir() -> str:
```
jameslamb: I think putting this in the `libkvikio` (instead of `kvikio`) packages is ok. Looks like it is less than 1MB uncompressed. And I could understand how any library built against libkvikio might also want this file at runtime.

If it were larger or pulled in other dependencies, I'd say it should be split out into its own `libkvikio-tools` or `libkvikio-devel` package or something.

But for now, this is much simpler. Let's leave it.

(no action needed, just want you to know how I thought through this)

msarahan:
small suggestion for the CMake code (via codex):

[P2] Link the plugin explicitly against Threads::Threads. The executable uses std::thread, but its [link declaration](https://github.com/rapidsai/kvikio/blob/496acd854f0e49cf212e6c6c6ed2e8c945650b20/cpp/nsys_plugins/nic/CMakeLists.txt#L20) only includes NVTX and ${CMAKE_DL_LIBS}. The Rocky Linux CI linker command confirms no -pthread; it currently succeeds through an indirect runtime dependency, which is toolchain-dependent. Since the parent already finds Threads, use:target_link_libraries(
  kvikio_nic_nsys_plugin
  PRIVATE nvtx3::nvtx3-cpp Threads::Threads ${CMAKE_DL_LIBS}
)

kingcrimsontianyu:
Thanks. Done. @msarahan

wence- (approved):

Matt711 (approved):
Just noting that I've also tested with it and it's helped tremendously, so thank you!

conda/recipes/libkvikio/recipe.yaml
```diff
@@ -112,6 +112,8 @@ outputs:
     tests:
       - script:
           - test -f $PREFIX/include/kvikio/file_handle.hpp
+          - test -x $PREFIX/share/kvikio/nsys-plugins/kvikio_nic/kvikio_nic_nsys_plugin
+          - test -f $PREFIX/share/kvikio/nsys-plugins/kvikio_nic/nsys-plugin.yaml
```
bdice, line 116: Thanks, this is really important to verify.

bdice (approved):
This is really nicely done. The packaging is tricky for something like this but I think this is a sensible solution.

msarahan (approved):

PR #1000, opened 2026-07-06
fix(deps): add explicit `cufile` dependency and load cufile

gforsyth:
Discovered while testing on a smaller wheel testing image in https://github.com/rapidsai/kvikio/pull/998.

Some streaming tests were failing because they can't load `cufile` -- this adds an explicit dependency on `nvidia-cufile-cu*` and then uses `cuda.pathfinder` to load it.

xref rapidsai/build-planning#143

Representative failure here: https://github.com/rapidsai/kvikio/actions/runs/28811101361/job/85441199077?pr=998#step:13:2073


gforsyth:
oh, there's a `cufile` extra for `cuda-toolkit` metapackage. let me use that instead

dependencies.yaml
```diff
@@ -284,6 +285,18 @@ dependencies:
...
+      - output_types: pyproject
+        matrices:
+          - matrix:
+              cuda: "12.*"
+            packages:
+              - nvidia-cufile-cu12
+          # fallback to CUDA 13 versions if 'cuda' is '13.*' or not provided
+          - matrix:
+            packages:
+              - nvidia-cufile-cu13
```
jameslamb: I think for this we'll want, the `cuda-toolkit[cufile]` extra, not `nvidia-cufile-cu13` directly.

That way:

1. this testing will actually test Minimum Version Compatibility guarantees (e.g. the ability to use `kvikio-cu12` built w/ 12.9 on CTK 12.2) https://github.com/rapidsai/kvikio/blob/abe0ccacf4fdfc4dd4e3a7bee6e8ad306bfe6fd1/dependencies.yaml#L237-L250
2. we're guaranteed to get an `nvidia-cufile-cu12` that's compatible with whatever other `nvidia-{thing}` are pulled into the environment

When you do that, please also add a `use_cuda_wheels` guard, to keep this out of builds where we prefer the system CTK like devcontainers and DLFW builds.

`cudf`'s treament of its NVRTC requirement would be a useful guide: https://github.com/rapidsai/cudf/blob/fb23b0d200fb3c21b8be6c3e7df33fb170fcfe14/dependencies.yaml#L955

jameslamb (changes requested):
Find with the idea of this `load_nvidia_dynamic_lib()` usage, but please see my suggestions on how the dependency is spelled.

dependencies.yaml
jameslamb: ha looks like you got there while I was typing: https://github.com/rapidsai/kvikio/pull/1000#issuecomment-4897035000

jameslamb (approved):
Congrats on getting PR 1000!! 🥳

gforsyth:
Ugh, we're getting a `cufile` related timeout on one of the tests -- I already restarted that job once and it stalled out in the same spot.

https://github.com/rapidsai/kvikio/actions/runs/28819409144/job/85476907339?pr=1000#step:13:1517

jameslamb:
@gforsyth I see the problem. Notice `cuda-toolkit[cufile]` isn't showing up in the diff in `libkvikio/pyproject.toml` and I don't see `nvidia-cufile` getting installed in that job you linked.

You need to add an entry like this in `[tool.rapids-build-backend]`:

```toml
[tool.rapids-build-backend]
matrix-entry = "cuda_suffixed=true;use_cuda_wheels=true"
```

Without that, the wheel doesn't have `cuda-toolkit[cufile]` as a requirement, because `rapids-build-backend` is falling back to the empty matrix at the end of the new list you added.

Example from `cudf`: https://github.com/rapidsai/cudf/blob/35e7cd6874ef1d2e87701556d41cf26d66c9c677/python/libcudf/pyproject.toml#L70

gforsyth:
Bleh, still deadlocking on that one `cufile` test

gforsyth:
Ok, I've tested this out locally on an L4 arm64 machine and I can reproduce the failure.

After getting the same timeout, I tried loading `libkvikio` in a Python interpreter and I get this:

```
>>> import libkvikio
>>> libkvikio.load_library()
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_dl_linux.py", line 218, in load_with_abs_path
    handle = _load_lib(desc, found_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_dl_linux.py", line 155, in _load_lib
    return ctypes.CDLL(filename, cdll_mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.12.13/lib/python3.12/ctypes/__init__.py", line 379, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: /pyenv/versions/3.12.13/lib/python3.12/site-packages/nvidia/cu13/lib/libcufile.so.0: undefined symbol: shm_open

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/libkvikio/load.py", line 44, in load_library
    load_nvidia_dynamic_lib("cufile")
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 303, in load_nvidia_dynamic_lib
    return _load_lib_no_cache(libname)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 175, in _load_lib_no_cache
    return LOADER.load_with_abs_path(desc, find.abs_path, find.found_via)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_dl_linux.py", line 220, in load_with_abs_path
    raise RuntimeError(f"Failed to dlopen {found_path}: {e}") from e
RuntimeError: Failed to dlopen /pyenv/versions/3.12.13/lib/python3.12/site-packages/nvidia/cu13/lib/libcufile.so.0: /pyenv/versions/3.12.13/lib/python3.12/site-packages/nvidia/cu13/lib/libcufile.so.0: undefined symbol: shm_open
```

so there's a symbol missing in the `libcufile` wheel for cuda 13.0.3 on arm64

I don't imagine a fix will get backported to that, yeah?  So probably we want to both: 

- skip the test on arm64 and cuda 13.0.3
- don't try to use `cuda.pathfinder` to dynamically load `libcufile.so` from the wheel on arm64 and cuda 13.0.3, since the test was previously passing using the cuda-devel images (although I will try to confirm that now)

python/libkvikio/libkvikio/load.py
```diff
@@ -38,8 +38,19 @@ def load_library():
...
+        try:
+            load_nvidia_dynamic_lib("cufile")
+        except RuntimeError as e:
+            # cufile wheels on arm64 on 13.0.3 are missing symbols (specifically `shm_open`)
+            # so we ignore loading errors related to that particular symbol.
+            # Any other RuntimeError gets raised to the user
+            if "undefined symbol: shm_open" in repr(e):
+                pass
+            else:
+                raise e
```
gforsyth: Open to other means of doing this, but after some consideration, this seemed like the most straightforward way.  There's a particular symbol missing, if that's what the error message says, then move-on

gforsyth:
Ok, the `cufile` wheel for arm64 is built incorrectly and isn't loading `librt.so.1` -- if we load it manually, then `cufile` loads properly:

```sh
[root@942dc5f00ed4 work]# python
Python 3.12.13 (main, Jul  1 2026, 20:40:39) [GCC 8.5.0 20210514 (Red Hat 8.5.0-28)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ctypes
>>> ctypes.CDLL("librt.so.1", mode=ctypes.RTLD_GLOBAL)
<CDLL 'librt.so.1', handle 342d23b0 at 0xefed3f4b8b90>
>>> from cuda.pathfinder import load_nvidia_dynamic_lib
>>> load_nvidia_dynamic_lib('cufile')
LoadedDL(abs_path='/pyenv/versions/3.12.13/lib/python3.12/site-packages/nvidia/cu13/lib/libcufile.so.0', was_already_loaded_from_elsewhere=False, _handle_uint=875329216, found_via='site-packages')
```

I'm working on reporting this upstream to `cufile`, but do we want to stick the `librt.so.1` load into our `load.py` for `libkvikio` to make this work?

bdice:
Yes, I'm fine with the proposed workaround. Ideally we can cite an nvbug number for this in a code comment.

bdice:
Should we file a bug with the `cuda.pathfinder` module too? Even if the wheel is incorrect, we could upstream this workaround into `load_nvidia_dynamic_lib` -- though I wouldn't block this PR on that.

gforsyth:
> Should we file a bug with the `cuda.pathfinder` module too? Even if the wheel is incorrect, we could upstream this workaround into `load_nvidia_dynamic_lib` -- though I wouldn't block this PR on that.

I'll file the bug and see what they say

python/libkvikio/libkvikio/load.py
```diff
@@ -37,9 +37,17 @@ def load_library():
...
+        import ctypes
+
         import rapids_logger
+        from cuda.pathfinder import load_nvidia_dynamic_lib
 
         rapids_logger.load_library()
+
+        # We manually load librt.so.1 here to workaround a bad dynamic link
+        # in nvidia_cufile-1.15.1.6-py3-none-manylinux_2_27_aarch64.whl
+        ctypes.CDLL("librt.so.1", mode=ctypes.RTLD_GLOBAL)
```
jameslamb, line 48: I'm ok with the workaround.

I looked and `librt.so.1` is part of `libc` / `glibc`:

* https://packages.debian.org/cgi-bin/search_contents.pl?word=librt.so.1&searchmode=searchfiles&case=insensitive
* https://pkgs.org/download/librt.so.1()(64bit)

It's safe to assume it'll exist.

From https://docs.oracle.com/cd/E86824_01/html/E54772/librt-3lib.html it sounds like it might be.... a compatibility layer for Solaris support? Maybe, based on that, if cufile is built against a new-enough `glibc` then it could drop `-lrt` (if it has it) completely?

jameslamb (approved):

gforsyth:
~/merge~

oops, need `kvikio-python-codeowners`

python/libkvikio/libkvikio/load.py
```diff
@@ -37,9 +37,18 @@ def load_library():
     try:
         # librapids_logger must be loaded before libkvikio because libkvikio
         # references it.
+        import ctypes
+
         import rapids_logger
+        from cuda.pathfinder import load_nvidia_dynamic_lib
```
kingcrimsontianyu, line 41: Do we need to explicitly declare `cuda.pathfinder` as a dependency?

python/libkvikio/libkvikio/load.py
```diff
@@ -37,9 +37,18 @@ def load_library():
...
 
         rapids_logger.load_library()
+
+        # We manually load librt.so.1 here to workaround a bad dynamic link
+        # in nvidia_cufile-1.15.1.6-py3-none-manylinux_2_27_aarch64.whl
+        # https://nvbugspro.nvidia.com/bug/6425783
+        ctypes.CDLL("librt.so.1", mode=ctypes.RTLD_GLOBAL)
+        load_nvidia_dynamic_lib("cufile")
     except ModuleNotFoundError:
```
kingcrimsontianyu, line 50: From the doc https://nvidia.github.io/cuda-python/cuda-pathfinder/latest/generated/cuda.pathfinder.load_nvidia_dynamic_lib.html, it seems `load_nvidia_dynamic_lib("cufile")` may throw some other types of exceptions. Should we catch those?

python/libkvikio/libkvikio/load.py
```diff
@@ -37,9 +37,18 @@ def load_library():
     try:
         # librapids_logger must be loaded before libkvikio because libkvikio
         # references it.
+        import ctypes
```
kingcrimsontianyu: `ctypes` is imported at the module level, so I think it is probably no longer needed here.

gforsyth: Yep! removing that now

python/libkvikio/libkvikio/load.py
gforsyth: I'll add it so it's available in the event someone only installs `libkvikio`

python/libkvikio/libkvikio/load.py
gforsyth: This is a very good question -- I think the answer is no.

If `rapids_logger` wheel IS installed, then `rapids_logger.load_library()` will succeed. After that, `cufile` is required whether the wheel is installed or it's already available on the system, and `load_nvidia_dynamic_lib` covers both of those cases.

If `rapids_logger` wheel ISN'T installed, then we assume we're in a system library scenario and we fall back to a system-installed `cufile` and `rapids_logger`.

In the event that `rapids_logger` IS installed AND there is no `libcufile.so` anywhere on the system, I think that we want to raise an error there.  

If that last point is incorrect, then we can refactor this to allow for a missing `cufile` install.

gforsyth:
Ok, so the `cufile` extra for `cuda-toolkit` was added in version 12.6.3 -- tests are passing here because `libcufile` is still present in the `citestwheel` image. I'll fix up the dependencies

python/kvikio/pyproject.toml
```diff
@@ -39,6 +39,7 @@ classifiers = [
 test = [
     "boto3>=1.21.21",
     "cuda-core>=0.5.1,<2",
+    "cuda-toolkit[cufile]",
```
gforsyth, line 42: This is added here for cuda-toolkit tested with CUDA >=12.8 or thereabouts, where the `cufile` extra is available, but we can't specify it as an extra via `$PIP_CONSTRAINTS`.

For versions <12.8, the `cufile` extra won't be available, but this only produces a non-blocking warning message and won't impact installation.

For those cases, we explicitly add `nvidia-cufile-cu12` to the test requirements.

dependencies.yaml
```diff
@@ -243,11 +248,13 @@ dependencies:
               use_cuda_wheels: "true"
             packages:
               - cuda-toolkit==12.2.*
+              - nvidia-cufile-cu12
```
gforsyth, line 252: This is a cuda-version where the `cufile` extra isn't available, so we specify the extra dependency

dependencies.yaml
```diff
@@ -243,11 +248,13 @@ dependencies:
...
           - matrix:
               cuda: "12.5"
               use_cuda_wheels: "true"
             packages:
               - cuda-toolkit==12.5.*
+              - nvidia-cufile-cu12
```
gforsyth, line 258: This is a cuda-version where the `cufile` extra isn't available, so we specify the extra dependency

kingcrimsontianyu (approved):

PR #1003, opened 2026-07-09
fix(deps): install cufile for cuda12 tests, add cuda-pathfinder dep

gforsyth:
This is a followup to #1000 -- after some additional testing on #998 we still weren't getting the right `cufile` dependencies on CUDA 12 tests, but with these changes we are.

xref rapidsai/build-planning#143


dependencies.yaml
```diff
@@ -485,6 +482,24 @@ dependencies:
...
+        packages:
+          - cuda-toolkit[cufile]
+    specific:
+      # The `cufile` extra was added to the `cuda-toolkit` metapackage in 12.6.3
+      # Adding an explicit test dependency on the cufile package here for
+      # testing on versions before 12.6.3
+      - output_types: pyproject
+        matrices:
+          - matrix:
+              cuda: "12.*"
```
bdice, line 497: I think this also works. Either is fine.
```suggestion
              cuda: "12.[012345]"
```

bdice (approved):

dependencies.yaml
gforsyth: That's good to know about!  I'm fine with the `*` for this one, but that will come in handy elsewhere

PR #1013, opened 2026-07-22
Fix cuFile dependency and enable CI wheel tests

bdice:
A few critical packaging/CI fixes:
- Wheel tests were not being run in `test.yaml`. This adds that job.
- The cuFile dependency wasn't being specified properly, so CUDA < 12.6 didn't have a runtime dependency on cuFile wheels, and `libkvikio.load_library()` would fail.
- The wheel script now explicitly loads libkvikio after installation (and without the `[test]` extras), so any missing runtime libraries will trigger a failure in CI.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -55,6 +55,21 @@ jobs:
       date: ${{ inputs.date }}
       script: ci/test_python.sh
       sha: ${{ inputs.sha }}
+  wheel-tests:
+    permissions:
+      actions: read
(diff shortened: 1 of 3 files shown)


bdice:
I verified that tests fail as expected here: https://github.com/rapidsai/kvikio/actions/runs/30014350579/job/89234334674?pr=1013

Now I will apply a fix -- and then rebase/retarget to `release/26.08`.

msarahan (approved):

PR #1014, opened 2026-07-22
Name KvikIO thread-pool workers for profiling

vyasr:
Give KvikIO's IO thread-pool workers a real OS thread name (`kvikio-<idx>` for the default pool, `kvikio-dev-<idx>` for per-block-device pools) via `BS::thread_pool`'s `init_task` hook. This makes them recognizable in profilers like nsys, `py-spy`, and `/proc/<pid>/task/<tid>/comm` instead of showing up as the parent process name.


cpp/src/file_handle.cpp
```diff
@@ -80,7 +80,8 @@ ThreadPool* get_thread_pool_per_block_device(std::string const& file_path)
     }
 
     // First file on this block device: create a new dedicated thread pool
-    auto thread_pool = std::make_shared<ThreadPool>(defaults::num_threads());
+    auto thread_pool = std::make_shared<ThreadPool>(defaults::num_threads(),
+                                                    make_thread_pool_init_task("kvikio-dev"));
```
kingcrimsontianyu: Could we call it "kvikio-block-device" / "kvikio-block-dev" or just "kvikio", to avoid confusion with "development"?

cpp/include/kvikio/threadpool_wrapper.hpp
```diff
@@ -1,17 +1,58 @@
...
+ */
+[[nodiscard]] inline std::function<void()> make_thread_pool_init_task(std::string prefix)
+{
+  auto counter = std::make_shared<std::atomic<unsigned int>>(0);
+  return [counter = std::move(counter), prefix = std::move(prefix)]() {
+    unsigned int const idx = counter->fetch_add(1, std::memory_order_relaxed);
+    // Linux comm limit is 15 chars + NUL.
+    std::array<char, 16> name{};
+    std::snprintf(name.data(), name.size(), "%s-%u", prefix.c_str(), idx);
+    pthread_setname_np(pthread_self(), name.data());
```
kingcrimsontianyu, line 54: NVTX has a function dedicated to thread renaming: [`nvtxNameOsThread`](https://nvidia.github.io/NVTX/doxygen/group___r_e_s_o_u_r_c_e___n_a_m_i_n_g.html#gaeb7d5b25e1147fc4aaf1f3acb8e719d0) . But it only worked with global NVTX domain, not with kvikio's custom domain ( https://nvidia.slack.com/archives/C07722DM4MA/p1739200643850559). What is interesting is that its documentation has the following statement:

> Tools expect thread ID to be a number that uniquely identifies the thread at the time of the call. Note that a thread's ID can be reused after it is destroyed. Tools may choose how to handle aliasing of thread IDs.
POSIX pthread_t type returned by pthread_self() may not comply with these expectations. Please use OS-specific thread ID instead of pthread_t.

I was not sure if the "use OS-specific thread ID instead of pthread_t" recommendation applied to `nvtxNameOsThread` specifically, or to the profiling tools in general (including nsys) that want to consume NVTX.

But I think whatever works in nsys-UI is the one we should pick, so the method in this PR should be good.

kingcrimsontianyu (approved):
Looks good! Just a minor suggestion.

cpp/src/file_handle.cpp
vyasr: Renamed to `kvikio-blk`. pthreads has a 15 char limit, but I also didn't want to just use `kvikio` since that would collide with the default pool, so this seemed like the best compromise and avoids `dev` altogether.

cpp/include/kvikio/threadpool_wrapper.hpp
vyasr: My read of the NVTX docs is that the "use OS-specific thread ID instead of pthread_t" note is specifically about the *thread identifier* argument passed to `nvtxNameOsThread` (which takes a `uint32_t` tid, not a `pthread_t`) i.e. it's warning against passing `pthread_self()` as the ID because `pthread_t` is opaque and can be reused. It's not a recommendation against `pthread_setname_np` as a naming mechanism in general.

PR #1021, opened 2026-08-06
Multi-poll retry support (1/n): Fix a bug in request retry

kingcrimsontianyu:
Currently with the easy thread pool backend, if a transfer of a subrange times out, the retry logic will kick in and redo the transfer. However, the offset counters in the `BounceBufferH2D` and `CallbackContext` are not reset prior to the retry, which will cause the new transfer to start at the previous offset, eventually triggering the data overflow exception.

This PR fixes this bug by resetting the offset counters before retry attempts.


cpp/tests/utils/hdfs_helper.cpp
```diff
@@ -22,7 +22,7 @@ namespace {
  * @brief Helper struct that wraps a buffer view and tracks how many data have been processed via an
  * offset value.
  */
-struct tracked_buffer_t {
+struct TrackedBuffer {
```
kingcrimsontianyu, line 25: Need to use CamelCase on structs, per KvikIO convention.

python/kvikio/tests/test_http_io.py
```diff
@@ -91,6 +92,55 @@ def do_HEAD(self) -> None:
...
+        max_stall_count: int = 1,
+        stall_duration: int = 5,
+        **kwargs,
+    ):
+        self.request_counter = request_counter
+        self.max_stall_count = max_stall_count
+        self.stall_duration = stall_duration
+        super().__init__(*args, directory=directory, **kwargs)
+
+    def copyfile(self, source, outputfile) -> None:
```
kingcrimsontianyu, line 127: Learned about this internal function of `SimpleHTTPRequestHandler` from Claude.

python/kvikio/tests/test_http_io.py
```diff
@@ -260,6 +310,24 @@ def test_retry_timeout_ok(tmpdir):
                 f.read(b)
 
 
+def test_retry_after_partial_body(tmpdir, xp):
```
kingcrimsontianyu, line 313: This test will be a fail on `main` and pass on this PR.

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -142,10 +147,27 @@ class CurlHandle {
...
    * See <https://curl.se/libcurl/c/curl_easy_perform.html>.
+   *
+   * @exception std::runtime_error if the transfer fails with a non-retryable error, or if it
+   * exhausts its attempt budget.
    */
   void perform();
 
+  /**
+   * @brief Perform a blocking network transfer and execute an on_retry callback to roll back to
+   * pre-transfer state.
```
vuule: adjust the docs to make it clear that `on_retry` is only called when an attempt fails.

vuule (approved):
just one suggestion, looks great otherwise

cpp/include/kvikio/shim/libcurl.hpp
kingcrimsontianyu: Thanks. Done!

bdice (approved):

PR #1022, opened 2026-08-07
Multi-poll retry support (2/n): Refactor and extract the retry logic

kingcrimsontianyu:
Functionally this PR adds nothing new: it refactors and then extracts the core retry logic to be shared by the easy thread pool and multi-poll backends.

## Related PR
Depends on #1021

PR #1023, opened 2026-08-07
Multi-poll retry support (3/n): Use KvikIO logger to display retry notices instead of the standard output

kingcrimsontianyu:
This small PR replaces the use of `std::cout` with `KVIKIO_LOG_WARN` to display the retry notices.

Depends on #1022


wence- (approved):
Looks good.

PR #1027, opened 2026-08-10
Fix multi poll backend deadlock

kingcrimsontianyu:
This PR does the following 2 things:
1. Previously, the multi poll backend occasionally hangs (to my experience, the frequency of hang was low for cudf-polars TPC-H SF-1K benchmarks, about 1 out of 30 runs). Interestingly, switching to `CU_STREAM_NON_BLOCKING` makes query 2 hang definitively, manifesting the bug. The hang happens due to this pattern: (1) In reactor's I/O thread, a shard's mutex in `BounceBufferCachePerThreadAndContext::try_get()` is used to acquire a buffer, which may call `cuMemHostAlloc`, (2) In `cuLaunchHostFunc`, the same mutex is used to return the buffer to the pool. Apparently `cuLaunchHostFunc` holds some internal mutex that blocks the progress of `cuMemHostAlloc`. The effective way to break the deadlock is to move CUDA call in `BounceBufferCachePerThreadAndContext::try_get()`  out of the locked region.
2. This PR also makes similar changes to `StreamCachePerThreadAndContext` and `BounceBufferCachePerThreadAndContext` to make CUDA calls in non-locked region.

Testing with cudf-polars TPC-H SF-1K shows that this PR effectively eliminates the hang.


cpp/src/detail/bounce_buffer_cache.cpp
```diff
@@ -57,29 +57,38 @@ BounceBufferCachePerThreadAndContext<Allocator>::try_get(CUcontext ctx)
...
 
-  // No buffer available on the free list. Allocate if under cap (or if cap is unlimited).
-  auto const total = shard.free.size() + shard.checked_out + shard.in_flight;
-  if (_cap.has_value() && total >= _cap.value()) { return std::nullopt; }
-
-  auto buf = BounceBufferPool<Allocator>::instance().get();
-  ++shard.checked_out;
-  return buf;
+  try {
+    return BounceBufferPool<Allocator>::instance().get();
```
kingcrimsontianyu, line 96: This single line is the key to the solution: get the buffer outside of the lock.

wence-: This seems reasonable. I think one has to assume, for reasoning about lock ordering, that any cuda driver call grabs the driver lock for the duration of the call. So if we are also holding a lock then we have to make sure we drop it, or are always holding it when calling in to the driver.

cpp/src/detail/bounce_buffer_cache.cpp
```diff
@@ -57,29 +57,38 @@ BounceBufferCachePerThreadAndContext<Allocator>::try_get(CUcontext ctx)
...
+  {
+    std::lock_guard const lock(shard.mutex);
+
+    // Discard free buffers whose size no longer matches the current bounce_buffer_size. Their
+    // destructors route through BounceBufferPool::put, which deallocates wrong-size buffers.
+    auto const current_size = defaults::bounce_buffer_size();
+    while (!shard.free.empty() && shard.free.back().size() != current_size) {
+      shard.free.pop_back();
+    }
+
```
wence-, line 74: issue: This can call into the driver if the bounce buffer if backed by either a `CudaPinnedAllocator` or `CudaPageAlignedPinnedAllocator`.

I suspect you need:
```
{
  std::unique_lock lock(shard.mutex);
  std::vector<...> to_free;
  while (!shard.free.empty() && ...) {
     to_free.push_back(std::move(shard.free.back());
     shard.free.pop_back();
  }
  lock.unlock();
  to_free.clear();
  lock.lock();
  if (!shard.free.empty()) {
     auto buf = std::move(shard.free.back());
     ..
     return buf;
}
```

wence- (changes requested):

cpp/src/detail/bounce_buffer_cache.cpp
kingcrimsontianyu: Thanks @wence- . That's a good catch, and takes more efforts to fix than I had thought for the cleanup path.

kingcrimsontianyu:
On a second thought, I have reverted the use of non-blocking stream in this PR. I will do that in a separate PR and reevaluate the default stream semantics we have maintained for the `FileHandle`. This will minimize the risks.

cpp/src/detail/bounce_buffer_cache.cpp
kingcrimsontianyu: The cleanup path in both `BounceBufferCachePerThreadAndContext` and `BounceBufferPool` has been improved (CUDA call moved out of lock).

wence- (approved):
Thanks!

PR #1032, opened 2026-08-13
Do not reset kvikio thread pool when nthreads is unchanged

Matt711:
When `set_thread_pool_nthreads`, we reset the thread pool which drains and detroys the previous one. We avoid that if the number of threads we are setting is unchanged.

See

```cpp
void defaults::set_thread_pool_nthreads(unsigned int nthreads)
{
  KVIKIO_EXPECT(
    nthreads > 0, "number of threads must be a positive integer", std::invalid_argument);
  thread_pool().reset(nthreads, make_thread_pool_init_task("kvikio"));
}
```

xref https://github.com/NVIDIA/cudf/pull/23634


madsbk (approved):

kingcrimsontianyu:
Thanks for the improvement.

PR #1016, opened 2026-07-28
Add structured, physical read traces

TomAugspurger:
This updates our logging setup to optionally emit structured, physical traces.

There are two main sets of changes:

1. Log *physical* reads at a `TRACE` level. This takes the "logical" reads from the user's program and traces the actual reads done by kvikio, which might be smaller thanks to kvikio's read splitting
2. Optionally output to a structured ndjson format, which is easier for downstream tools to parse.

Here's an example program:

```
import kvikio
import cupy as cp


remote_file = kvikio.RemoteFile.open_s3_url("s3://rapids-tpch/tpch-rs/scale-1/nation/part.0.parquet")
buf = cp.empty(remote_file.nbytes(), dtype=cp.uint8)

size = 1024

remote_file.read(buf, size, 0)
remote_file.read(buf, remote_file.nbytes() - size, size)

print("done")
```

When run with `KVIKIO_LOG_LEVEL=TRACE KVIKIO_LOG_FORMAT=JSON KVIKIO_LOG_FILE=trace.ndjson python test_remote.py` that writes 

```
$ cat trace.ndjson 
{"event":"http","level":"trace","source":"https://rapids-tpch.s3.us-east-2.amazonaws.com/tpch-rs/scale-1/nation/part.0.parquet","start":1785255602404710891,"end":1785255602626790456,"threadId":4621804004366526777,"method":"HEAD","status":"ok","purpose":"metadata"}
{"event":"read","level":"trace","source":"https://rapids-tpch.s3.us-east-2.amazonaws.com/tpch-rs/scale-1/nation/part.0.parquet","start":1785255603191754437,"end":1785255603272868402,"offset":0,"size":1024,"threadId":6135855871742540183,"bytesRead":1024,"backend":"remote","status":"ok","isDeviceBuffer":true,"requestId":1,"method":"GET"}
{"event":"read","level":"trace","source":"https://rapids-tpch.s3.us-east-2.amazonaws.com/tpch-rs/scale-1/nation/part.0.parquet","start":1785255603273011822,"end":1785255603349365176,"offset":1024,"size":1226,"threadId":6135855871742540183,"bytesRead":1226,"backend":"remote","status":"ok","isDeviceBuffer":true,"requestId":2,"method":"GET"}
```

Closes https://github.com/rapidsai/kvikio/issues/967

PR #1033, opened 2026-08-14
Add an I/O observation facility

madsbk:
This PR introduces a hook into monitoring KvikIO operations, with the goal of building statistics, Quent timelines, and whatever else wants to know what the I/O layer is doing. This PR is the base. A follow-up introduces a concrete `Monitor` that makes statistics easy to get.

The hook reports whole KvikIO operations (the logical level) so a `pread()` is a single observation however many reads the thread pool issued underneath. Physical operations can be added along the same path later, which could be the basis of #1016.

Each call produces a `kvikio::Observation`: its span, the offset and size etc. To receive them, derive from `kvikio::Monitor` and register it. A monitor is told when an operation *starts* as well as when it finishes.

```c++
// Example of a Monitor that tracks how many KvikIO operations are in flight at any moment.
class QueueDepth : public kvikio::Monitor {
  void on_start(kvikio::Observation const&) noexcept override { ++_in_flight; }
  void on_finish(kvikio::Observation const&) noexcept override { --_in_flight; }
  std::atomic<int> _in_flight{0};
};

QueueDepth gauge;
auto id = kvikio::register_monitor(&gauge);
```

### Overhead

Measured on my local workstation:

- **~3 ns per call when nobody is observing**, which is a gate check and a branch.
- **~60 ns per call when somebody is**, or 1 % of a 64 KiB `pread()`, and nothing detectable at a megabyte.

Confirmed against a real workload: cudf-polars PDS-H query 1 at scale 10, with and without a monitor attached, showed no difference outside noise.

### What is not observed

The cuFile asynchronous API on a GDS system, and the batch API, complete without KvikIO seeing it, so they emit nothing. Handling those needs a stream-completion callback, which is future work.

### Follow-up: statistics

The next PR adds `kvikio::statistics::SummaryMonitor`, which is a `Monitor` and nothing more:

```python
monitor = kvikio.SummaryMonitor()   # statistics are now on
...
print(monitor.get())
```

```
KvikIO I/O summary
  wall time            1.876 s
  busy time            366.592 ms (19.54 % of the wall time)
  busy bandwidth       8.95 GB/s
  operations           7970
  bytes requested      3.06 GiB
  bytes transferred    3.06 GiB
  errors               0
```
Those are real numbers, from a cudf-polars run, and they show a very useful `busy bandwidth`.  **8.95 GB/s** is the rate while KvikIO actually had work in flight, where dividing the same bytes by the wall clock would have said **1.75 GB/s** and described the query rather than the storage.

A `TimelineMonitor`, for *when* things happened rather than how much, is planned after that.

### DIFF
--- a/cpp/include/kvikio/detail/observation_recorder.hpp
+++ b/cpp/include/kvikio/detail/observation_recorder.hpp
@@ -1,0 +1,166 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <atomic>
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
+
+#include <kvikio/observation.hpp>
+
+namespace kvikio::detail {
+
+/**
+ * @brief Throw if the calling thread is inside a monitor callback.
+ *
+ * @exception std::runtime_error if called from inside a monitor callback.
+ */
+void expect_not_in_monitor();
+
+/**
+ * @brief The reading KvikIO timestamps observations with.
+ *
+ * @return Now, on `kvikio::Clock`.
+ */
+[[nodiscard]] inline TimePoint now() noexcept { return Clock::now(); }
+
+/**
+ * @brief Tell every monitor subscribed to its kind that an operation started.
+ *
+ * Runs the callbacks inline, on the calling thread. Does nothing if no monitor is registered, or
+ * if the calling thread is already inside a monitor callback.
+ *
+ * @param observation The operation, with `start` set and the fields known at submission.
+ */
+void notify_started(Observation const& observation) noexcept;
+
+/**
+ * @brief Tell every monitor subscribed to its kind that an operation finished.
+ *
+ * Runs the callbacks inline, on the thread that completed the operation, which need not be the one
+ * that started it. Does nothing if no monitor is registered, or if the calling thread is already
+ * inside a monitor callback.
+ *
+ * @param observation The completed operation, with `end`, `bytes_transferred` and `ok` set.
+ */
+void notify_finished(Observation const& observation) noexcept;
+
+/**
+ * @brief Whether any monitor is registered.
+ *
+ * @return True if an observation would reach somebody.
+ */
+[[nodiscard]] bool monitoring_enabled() noexcept;
+
+/**
+ * @brief Times one user-facing call and emits its logical observation.
+ *
+ * Construction takes the start timestamp. `finish()` takes the end timestamp and notifies, exactly
+ * once. There are two ways to use it, according to where the call actually ends.
+ *
+ * **Scope-bound**, for a blocking call: leave it on the stack and let the destructor finish it. An
+ * exception leaving the scope marks the observation failed, so error paths are reported without
+ * every call site having to catch.
+ *
+ * **Shared**, for a call whose work outlives the function that started it. `pread()` returns as
+ * soon as the parts are submitted, so a destructor at the end of that function would stop the clock
+ * before the work finished. Hold the recorder by `std::shared_ptr`, hand it to the tasks, and have
+ * the task that completes the work call `finish()`. The end timestamp is then the true completion
+ * time, the observation is delivered before the caller's future becomes ready, and none of it
+ * depends on the caller ever waiting. A failing part calls `finish_with_failure()`, so one failed
+ * part is one failed logical operation rather than a lost one.
+ *
+ * `finish()` is idempotent, so the destructor is a safe backstop under either use. When nobody is
+ * subscribed the whole object collapses to a single relaxed atomic load. A shared user should still
+ * test `monitoring_enabled()` first, to skip the allocation.
+ */
+class LogicalObservationRecorder {
+ public:
+  /**
+   * @brief Start recording a call.
+   *
+   * @param backend The backend.
+   * @param direction The direction.
+   * @param memory_kind The kind of memory the caller's buffer lives in.
+   * @param offset Byte offset into the file or remote object.
+   * @param size Number of bytes requested.
+   * @param http_method HTTP method for a remote call, e.g. `"GET"`. A constructor parameter rather
+   * than a setter because the monitors are told the operation has started before this constructor
+   * returns, and the record they are shown must be as complete as it can be.
+   */
+  LogicalObservationRecorder(IoBackend backend,
+                             TransferDirection direction,
+                             MemoryKind memory_kind,
+                             std::size_t offset,
+                             std::size_t size,
+                             char const* http_method = nullptr) noexcept
+    : _active{monitoring_enabled()}
+  {
+    if (_active) { begin(backend, direction, memory_kind, offset, size, http_method); }
+  }
+
+  /**
+   * @brief Backstop. An operation that was never finished did not complete, so it is recorded as
+   * a failure.
+   *
+   * Every path that completes the work calls `finish()` or `finish_with_failure()`, and this then
+   * does nothing. Reaching here un-emitted means the call threw on its way to the work, or was
+   * abandoned.
+   */
+  ~LogicalObservationRecorder() { finish_with_failure(); }
+
+  LogicalObservationRecorder(LogicalObservationRecorder const&)            = delete;
+  LogicalObservationRecorder& operator=(LogicalObservationRecorder const&) = delete;
+  LogicalObservationRecorder(LogicalObservationRecorder&&)                 = delete;
+  LogicalObservationRecorder& operator=(LogicalObservationRecorder&&)      = delete;
+
+  /**
+   * @brief The call completed. Stamp the end time and emit. Idempotent.
+   *
+   * @param bytes_transferred Number of bytes the call actually moved.
+   */
+  void finish(std::size_t bytes_transferred) noexcept
+  {
+    if (!_active) { return; }
+    if (_emitted.exchange(true, std::memory_order_relaxed)) { return; }
+    _observation.bytes_transferred = bytes_transferred;
+    emit();
+  }
+
+  /**
+   * @brief The call failed. Emit it as an error that moved nothing. Idempotent.
+   *
+   * One failing part fails the whole logical operation, so this is what a fan-out calls when any
+   * of its parts failed.
+   */
+  void finish_with_failure() noexcept
+  {
+    if (!_active) { return; }
+    if (_emitted.exchange(true, std::memory_order_relaxed)) { return; }
+    _observation.ok                = false;
+    _observation.bytes_transferred = 0;
+    emit();
+  }
+
+ private:
+  /// Fill in the record and tell the monitors. Out of line, see the constructor.
+  void begin(IoBackend backend,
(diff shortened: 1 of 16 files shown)


cpp/src/remote_handle.cpp
```diff
@@ -875,14 +932,7 @@ std::future<std::size_t> RemoteHandle::pread(void* buf,
   //
   // Build all N transfers here, then hand them off in a single pool call.
   KVIKIO_EXPECT(task_size > 0, "`task_size` must be positive", std::invalid_argument);
-  if (is_read_out_of_bounds(file_offset, size, _nbytes)) {
-    std::stringstream ss;
-    ss << "cannot read " << file_offset << "+" << size << " bytes into a " << _nbytes
-       << " bytes file (" << _endpoint->str() << ")";
-    KVIKIO_FAIL(ss.str(), std::invalid_argument);
-  }
```
madsbk, line 883: Not removed, just moved to the top of read, before the buffer is classified.

cpp/src/remote_handle.cpp
```diff
@@ -775,20 +776,57 @@ std::size_t callback_device_memory(char* data, std::size_t size, std::size_t nme
...
+  // Before the buffer is classified and before any monitor is told: a call rejected on its
+  // arguments never became an I/O operation.
+  expect_read_in_bounds(size, file_offset);
 
-  if (is_read_out_of_bounds(file_offset, size, _nbytes)) {
-    std::stringstream ss;
-    ss << "cannot read " << file_offset << "+" << size << " bytes into a " << _nbytes
-       << " bytes file (" << _endpoint->str() << ")";
-    KVIKIO_FAIL(ss.str(), std::invalid_argument);
-  }
```
madsbk, line 789: Not removed, just moved to the top of read, before the buffer is classified.

cpp/include/kvikio/observation.hpp
```diff
@@ -0,0 +1,262 @@
...
+  TransferDirection direction{TransferDirection::Read};
+  /// The kind of memory the caller's buffer lives in.
+  MemoryKind memory_kind{MemoryKind::Host};
+  /// False if the operation failed.
+  bool ok{true};
+  /// Identifies this operation, uniquely within the process.
+  std::uint64_t id{};
+
+  /// HTTP method, e.g. `"GET"`. Null for local I/O.
+  char const* http_method{nullptr};
```
wence-: nit: Check the layout of this struct for potentially more padding than necessary

cpp/src/file_handle.cpp
```diff
@@ -181,9 +182,30 @@ std::size_t FileHandle::read(void* devPtr_base,
                              bool sync_default_stream)
 {
   KVIKIO_NVTX_FUNC_RANGE(size);
-  if (get_compat_mode_manager().is_compat_mode_preferred()) {
-    return detail::posix_device_read(
+  detail::expect_not_in_monitor();
```
wence-, line 195: question: Should we make this a no-op unless in debug mode?

cpp/src/observation.cpp
```diff
@@ -0,0 +1,269 @@
...
+  if (monitor_count.load(std::memory_order_acquire) != 0) { notify_started(_observation); }
+}
+
+void LogicalObservationRecorder::emit() noexcept
+{
+  _observation.end = now();
+  // A monitor registered after this operation began never saw its start and ignores the finish,
+  // which is why no flag is latched here.
+  notify_finished(_observation);
+}
```
wence-: I don't understand this comment

cpp/src/remote_handle.cpp
```diff
@@ -775,20 +776,57 @@ std::size_t callback_device_memory(char* data, std::size_t size, std::size_t nme
...
+  detail::expect_not_in_monitor();
+  // Classified once here and passed down, since the query is a CUDA driver call.
```
wence-: Unnecessary added comment?

cpp/tests/test_observation.cpp
```diff
@@ -0,0 +1,493 @@
...
+  // rather than a bare timestamp, because two operations can be stamped in the same nanosecond.
+  class Correlator final : public kvikio::Monitor {
+   public:
+    void on_start(Observation const& o) noexcept override
+    {
+      std::lock_guard const lock{_mutex};
+      _open[o.id] = o.size;
+      // What is known at submission is set. What is not, is not.
+      if (o.size == 0 || o.id == 0 || o.start == kvikio::TimePoint{}) { ++_malformed; }
+      if (o.end != kvikio::TimePoint{} || o.bytes_transferred != 0) { ++_malformed; }
```
wence-: I don't understand this checking of the start against the timepoint. These could match/mismatch validly, no?

wence- (commented):
Broadly looks good, I think.

cpp/tests/test_observation.cpp
madsbk: You are right `TimePoint{}` is the clock epoch, so it is a stand-in for "unset" that a real timestamp could in principle collide with.

Replaced with something that has content: the map now holds the whole `Observation`, and `on_finish()` checks the completion carries the same `size` and the same `start` as the record seen at submission.

cpp/src/remote_handle.cpp
madsbk: yes, removed :)

cpp/src/observation.cpp
madsbk: Fair, it was too compressed. It was trying to say why the recorder does not track which monitors it notified. Rewritten as:
```c++
  // Delivered to whoever is registered now, which need not be who was registered when the
  // operation began. A monitor that missed the start ignores this, as `Monitor::on_finish()`
  // documents, so the recorder does not have to remember which monitors it told.
```

cpp/src/file_handle.cpp
madsbk: I would rather keep it in release. It is one thread-local bool load per user-facing call, and the whole disabled observation path measured at about 3 ns. 

What it buys is a clear error instead of a hang.

cpp/include/kvikio/observation.hpp
madsbk: Good catch, moved `kind` down to join the other small enums and `id` up with the other 8-byte fields. Now
64 bytes with a single 3-byte tail hole.

cpp/include/kvikio/observation.hpp
```diff
@@ -0,0 +1,261 @@
...
+/**
+ * @brief The I/O backend that carried out an operation.
+ */
+enum class IoBackend : std::uint8_t {
+  Posix = 0,   ///< POSIX `pread`/`pwrite`, including the compatibility-mode path.
+  Gds,         ///< cuFile / GPUDirect Storage.
+  Mmap,        ///< Memory-mapped file access.
+  RemoteHttp,  ///< Remote I/O over HTTP(S), including S3.
+  RemoteHdfs   ///< Remote I/O over WebHDFS.
+};
```
kingcrimsontianyu, line 86: Nit: Should we follow cuDF's convention (and also Python's) to use all capital letters for enums? (Precedent in KvikIO: https://github.com/rapidsai/kvikio/blob/main/cpp/include/kvikio/detail/posix_io.hpp#L29). 

Although the C++ core guideline [advises against all capital style to avoid clash with macros](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#renum-caps), I think this style makes our C++/Cython/Python more consistent with each other.

madsbk: Good point, done.

cpp/src/observation.cpp
madsbk: Removed now.

wence- (approved):

vyasr (approved):
Approving CMake (didn't review the C++).

cpp/include/kvikio/detail/observation_recorder.hpp
```diff
@@ -0,0 +1,170 @@
...
+ * @exception std::runtime_error if called from inside a monitor callback.
+ */
+void expect_not_in_monitor();
+
+/**
+ * @brief The reading KvikIO timestamps observations with.
+ *
+ * @return Now, on `kvikio::Clock`.
+ */
+[[nodiscard]] inline TimePoint now() noexcept { return Clock::now(); }
```
kingcrimsontianyu: I think in general we should place all implementations in a corresponding .cpp file, i.e. `src/detail/observation_recorder.cpp`, no matter how small they may look like.

cpp/src/observation.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+std::uint64_t register_monitor(Monitor* monitor, ObservationKind kind)
+{
+  return Registry::instance().add(monitor, kind);
+}
+
+void unregister_monitor(std::uint64_t id) { Registry::instance().remove(id); }
+
+namespace detail {
```
kingcrimsontianyu, line 230: A better location for this section is `src/detail/observation_recorder.cpp`. This would improve consistency.

madsbk:
But register_monitor is the public API

kingcrimsontianyu:
Oh I was commenting on the `namespace detail` part.

madsbk:
But do you really want to split that into two files ?

kingcrimsontianyu:
It's up to you. We do have an existing `src/detail` directory containing code that otherwise goes to the detail namespace. Not a big deal though.

madsbk:
Hmm alright, let me think about this. It might be nice to have a clean
header that only have the public APl. I can also move the small impl to
.cpp.

cpp/include/kvikio/observation.hpp
```diff
@@ -0,0 +1,309 @@
...
+   * @brief Convert a point on `Clock` to wall-clock time.
+   *
+   * @warning The wall clock can be stepped or slewed by NTP or an operator, so an
+   * old anchor may no longer be valid. Take an anchor at each end of a long run
+   * and compare them to detect an in-flight adjustment to the system time.
+   *
+   * @param time The point to convert.
+   * @return The corresponding wall-clock time.
+   */
+  [[nodiscard]] std::chrono::system_clock::time_point to_wall_clock(TimePoint time) const noexcept
```
kingcrimsontianyu, line 66: Nit: Should we drop `noexcept` considering that `operator+/-` and `duration_cast` may potentially throw?

madsbk: It cannot throw, updated the comment:
```c++
  [[nodiscard]] std::chrono::system_clock::time_point to_wall_clock(TimePoint time) const noexcept
  {
    // Chrono is not noexcept because a duration's representation may have throwing arithmetic, and
    // duration_cast is specified in terms of the same operations. Ours are integers, so nothing
    // here can throw.
    static_assert(std::is_integral_v<TimePoint::rep>);
    static_assert(std::is_integral_v<std::chrono::system_clock::rep>);
    return wall + std::chrono::duration_cast<std::chrono::system_clock::duration>(time - steady);
  }
```

cpp/src/observation.cpp
madsbk: done

kingcrimsontianyu (approved):
Thanks. Good design!

cpp/include/kvikio/detail/observation_recorder.hpp
madsbk: done

PR #1034, opened 2026-08-14
Multi-poll backend optimization: configure CURLMOPT_MAXCONNECTS to avoid unnecessary TCP/TLS handshakes

kingcrimsontianyu:
By default libcurl sets `CURLMOPT_MAXCONNECTS` to 4 x the number of easy handles attached to a multi handle. This is recomputed on every transition, and a transient dip in concurrency will cause libcurl to evict warm, reusable connections, and cause unnecessary TCP/TLS handshake. In this PR we pin `CURLMOPT_MAXCONNECTS` to a fixed size to avoid this problem.

Benchmark on cudf-polars TPC-H SF-1K query 5 shows that, on `main` the TCP connection opened is 618, whereas with this PR, the connection number drops to 496, which is a ~20% saving.


cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -136,9 +137,40 @@ MultiPollReactor::MultiPollReactor(MultiReactorPool* pool,
...
+  set_connection_cache_size(max_concurrent_requests);
   _io_thread = std::thread(&MultiPollReactor::io_thread_main, this);
 }
 
+std::optional<long> connection_cache_size(
+  std::optional<std::size_t> max_concurrent_requests) noexcept
+{
+  if (!max_concurrent_requests.has_value()) { return std::nullopt; }
+
+  // libcurl documents this option as taking a `long`, and the value is internally store as an
```
madsbk:
```suggestion
  // libcurl documents this option as taking a `long`, and the value is internally stored as an
```

cpp/tests/test_multi_poll_reactor.cpp
```diff
@@ -11,90 +16,124 @@
...
+
+TEST(ConnectionCacheSize, never_exceeds_what_libcurl_accepts)
+{
+  constexpr auto limit = std::min(static_cast<long>(std::numeric_limits<unsigned>::max()),
+                                  std::numeric_limits<long>::max());
+  std::array ceilings  = {std::numeric_limits<std::size_t>::max(),
+                          std::numeric_limits<std::size_t>::max() / 2,
+                          static_cast<std::size_t>(limit),
+                          static_cast<std::size_t>(limit) / 2 + 1};
+  for (const auto ceiling : ceilings) {
```
madsbk:
```suggestion
  for (auto const ceiling : ceilings) {
```

madsbk (approved):
Thanks @kingcrimsontianyu

PR #1036, opened 2026-08-18
Add I/O statistics: totals of what a run read and wrote

madsbk:
Builds on the observation facility from #1033. That PR gives a callback per user-facing I/O call, and this one gives the answer most people actually want from it, which is what a run did in total.

`SummaryMonitor` registers itself on construction and accumulates while it exists. `Summary` carries the operations, the bytes, the errors, and the time at least one operation was in flight, which is what distinguishes an I/O-bound run from a compute-bound one.

### Using it

```python
with kvikio.SummaryMonitor() as monitor:
    ...
print(monitor.get())
```

In C++ the constructor takes an optional callback, which runs on destruction, so a program can report its own I/O without touching the code that performs it.

```c++
kvikio::statistics::SummaryMonitor const monitor{
  [](kvikio::statistics::Summary const& summary) { std::cout << summary.report(); }};
```

An interval is the difference of two readings, so periodic reporting takes one reading per tick and differences it against the last. Taking the reading and the interval separately would leave a gap that an operation could fall into twice.

```python
baseline = monitor.get()
while running:
    time.sleep(interval)
    now = monitor.get()
    report(now.since(baseline))
    baseline = now
```

### The report

Printing a summary, or `report()`, gives a report meant for a person, always the same shape so two runs can be compared line by line. This is `python/kvikio/examples/hello_world.py`:

```
KvikIO I/O summary
  wall time            251.72 ms
  busy time            5.95 ms (2.36 % of the wall time)
  busy bandwidth       538.15 kB/s
  operations           5 (4 read, 1 write)
  time per operation   1.19 ms mean, 3.19 ms longest
  bytes                3.12 KiB of 3.12 KiB requested (2.34 KiB read, 800 B written)
  errors               0
  backend POSIX        3.12 KiB in 5 ops, 5.95 ms, 538.15 kB/s
  backend GDS          unused
  backend MMAP         unused
  backend REMOTE_HTTP  unused
  backend REMOTE_HDFS  unused
```

Five operations for one write and four reads, however many reads KvikIO issued underneath, since the observations are logical.

**Busy time** is the union of the operations' spans, so overlapping work counts once and the gaps between calls count as idle. **Busy bandwidth** divides by that rather than by the wall time, so a program that reads for 10 ms and then computes for 90 ms is not reported as ten times slower than its storage really is. The **backend** rows are the only place the report says whether a read reached cuFile or fell back to POSIX, which compatibility mode decides per call.

### Overhead

On my local workstation, 32 cores, medians over two million observed operations, so an order of magnitude rather than a specification.

| per operation | |
|---|---|
| no monitor registered | 4.8 ns |
| `SummaryMonitor` | 64 to 79 ns |

This includes everything, both the observation facility from #1033 stamping and dispatching each operation, and the accumulation this PR adds on top. So a monitor costs about 2 % of a 4 KiB `pread()` and 0.25 % of a 1 MiB one. Registering none is the 4.8 ns row, one relaxed atomic load per operation.

Eight threads doing nothing but emitting observations cost 320 ns each, every operation taking the monitor's lock to add itself to the totals. Real work in between makes it disappear.

### Follow-ups

- **What KvikIO spends on itself**, the bounce buffers it allocates, the connections it opens, the time inside the file system. The next PR adds those counters and surfaces them on `Summary`.
- **A record per operation**, which a summary cannot give: a timeline monitor and a sampling monitor.
- **`ObservationKind::PHYSICAL`**. Everything here is logical, one user-facing call being one operation. When physical operations arrive, `Summary` stays one type rather than splitting in two, since the fields mean the same thing at either level.


cpp/examples/basic_io.cpp
```diff
@@ -14,28 +13,10 @@
 #include <kvikio/defaults.hpp>
 #include <kvikio/error.hpp>
 #include <kvikio/file_handle.hpp>
+#include <kvikio/statistics/summary.hpp>
 
 using namespace std;
 
-class Timer {
```
madsbk, line 20: We can now use the new Monitor instead of this homemade `Timer`

cpp/include/kvikio/detail/string_utils.hpp
```diff
@@ -0,0 +1,78 @@
+/*
```
madsbk, line 1: This file and its .cpp is mostly a copy-paste from https://github.com/rapidsai/rapidsmpf/blob/main/cpp/include/rapidsmpf/utils/string.hpp#L73

python/kvikio/kvikio/_lib/statistics.pyx
```diff
@@ -0,0 +1,191 @@
...
+        cpp_ClockDuration duration
+    )
+    cpp_Duration system_to_ns "std::chrono::duration_cast<kvikio::Duration>"(
+        cpp_SystemDuration duration
+    )
+
+
+# In the order of `kvikio::IoBackend`, which is what the arrays above are indexed by.
+_BACKEND_NAMES = ("POSIX", "GDS", "MMAP", "REMOTE_HTTP", "REMOTE_HDFS")
+
```
kingcrimsontianyu, line 43: Would it work if we import the C++ enum and do `for i, name in enumerate(IoBackend)` below since Python enum is iterable?

cpp/include/kvikio/detail/string_utils.hpp
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/**
+ * @brief Format a byte count as a human-readable string using IEC units.
+ *
+ * Converts a byte count into a scaled string representation using binary (base-1024) units such
+ * as KiB, MiB and GiB.
+ *
+ * Negative values are supported and are formatted with a leading minus sign, which is useful when
+ * representing signed byte deltas.
```
kingcrimsontianyu: Is this intentional, or an unwanted comment from `format_rate/duration`?

madsbk: Yes, removed

python/kvikio/kvikio/_lib/statistics.pyx
madsbk: Yes, good point, done

bdice (approved):

cpp/src/statistics/summary.cpp
```diff
@@ -0,0 +1,433 @@
...
+  // part of its span that follows it.
+  _busy.reset(t);
+}
+
+void SummaryMonitor::stop()
+{
+  if (_registration == 0) { return; }
+  // Waits for a notification in progress, so once this returns no thread is inside this object.
+  unregister_monitor(_registration);
+  _registration = 0;
```
kingcrimsontianyu, line 426: Do we expect `SummaryMonitor` to be owned and used by a single thread? If it is shared by two threads, there could be a race condition, where one thread does the write `_registration = 0;` and the other does the read `if (_registration == 0)`, or they both pass the `unregister_monitor` call and do the write at the same time.

kingcrimsontianyu (approved):
Great framework. This opens doors to many statistics we want to collect such as p50/p90/ p99 latency and probably TTFB (time to first byte).

cpp/src/statistics/summary.cpp
madsbk: Good catch, fixed

PR #1041, opened 2026-08-21
[DO NOT MERGE] ci: release catalog canary

msarahan:
**Posted by Codex (GPT-5.6) on behalf of Michael Sarahan; treat this LLM-generated content as a proposed implementation requiring human review.**

## Why

RAPIDS release-platform needs real, immutable build evidence before it can safely assemble a release candidate. This canary lets us validate that evidence against KvikIO’s existing Conda and wheel build jobs before asking every repository to adopt the shared release-catalog contract.

The branch deliberately uses the unmerged shared-workflows release-catalog revision. It must not merge until that shared workflow is approved and its contract is understood.

## Scope

The change only points KvikIO’s existing shared Conda and wheel workflow callers at the canary shared-workflows revision. Those workflows should continue to produce their normal artifacts and additionally emit the single `release-catalog-entries.json` companion per source artifact.

## Success criteria

A fresh build at this PR commit produces release-catalog companions whose source SHA and artifact name match the build. Release-platform will then consume those artifacts directly in an end-to-end collection test; it will not support the deprecated release-build-output contract.

## Validation

- Full applicable pre-commit checks for both changed workflow files
- `git diff --check`

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -54,8 +54,6 @@ jobs:
       node_type: cpu8
       date: ${{ inputs.date }}
       script: ci/build_cpp.sh
-      release-build-output: true
-      release-unit: conda:kvikio-cpp
       sha: ${{ inputs.sha }}
   python-build:
     needs: [build-details, cpp-build]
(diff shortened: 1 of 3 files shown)


msarahan:
/ok to test [1377c1c](https://github.com/rapidsai/kvikio/pull/1041/commits/1377c1ceeb60fb14495ec68e470c19ba1d07751b)

Issue #1045, opened 2026-08-25
Attach retry/backoff outcome to remote-read observations


Matt711:
`HttpRetryPolicy`/`RetryOutcome` (`cpp/include/kvikio/detail/http_retry.hpp`) already computes per-attempt retry decisions and exponential backoff inside `remote_handle.cpp`'s retry loop. None of it (attempt count, backoff time, final HTTP status) reaches `Observation` or `Summary`. Relevant to the cloud push specifically: a query that looks slow from `max_concurrent_io_tasks`'s vantage point could actually be S3 throttling triggering retries, which no current metric distinguishes from ordinary latency. Add `attempt_count`/`final_status` fields to `Observation`, threaded from `HttpRetryPolicy::evaluate()`.

PR #1047, opened 2026-08-27
Add performance and diagnostic counters for remote I/O

madsbk:
This PR introduces always-on performance and diagnostic counters for remote I/O. These counters belong to no single operation, either because KvikIO does not track enough to correlate them with one, libcurl's connection statistics for instance, or because they are shared between many operations.

For a remote read a good deal of the time goes somewhere other than the transfer. The file has to be asked how big it is, a connection may have to be opened and a TLS handshake completed, and the endpoint may turn a request away and make KvikIO wait before trying again.

`kvikio::statistics::counters()` returns running totals for the process, so the cost of an interval is the difference between two readings. That is deliberately the whole design: no subscription, no record per event, no notification path, and nothing to enable. Relaxed atomics on a leaked singleton, read by whoever wants them. `Summary` carries a `Counters` for its span, so one report says what the run read and what reaching the data cost.

### What it looks like

```
KvikIO I/O summary
  wall time            352.71 ms
  busy time            4.77 ms (1.35 % of the wall time)
  busy bandwidth       335.52 kB/s
  operations           2 (2 read, 0 write)
  bytes                1.56 KiB of 1.56 KiB requested (1.56 KiB read, 0 B written)
  backend REMOTE_HTTP  1.56 KiB in 2 ops, 4.77 ms, 335.52 kB/s
  http size probes     1 probes, 2.93 ms
  http handshake       3 connections, 49 us dns, 163 us tcp, 0 s tls
  http retries         0 retries, 0 s backoff
```

### What it costs

Nothing to collect. libcurl measures the connection phases whether or not anybody asks, and reports them cumulatively from the start of the transfer, so one `curl_easy_getinfo` call after a transfer completes reads them all back out and differences them. It runs after every attempt rather than only after the one that succeeded, since an attempt that opened a connection and then failed would otherwise go uncounted. The size probe and the retry backoff are one line each at sites that already cost milliseconds.

### Follow-ups

- **More counters.** The struct takes anything that belongs to no single operation, and adding one is a field, a recorder and a row. Twenty-six were written and measured against cudf-polars on PDS-H before being cut to keep this reviewable, covering buffers, the file system, cuFile, and the work deferred behind KvikIO's own limits.
- **Per interval rather than per run.** An interval is a `Summary`, so a sampling monitor carries these counters too. That is where a burst shows up that a run total flattens into a number.
- **Beyond remote I/O.** Nothing in the design is specific to remote reads, only the instrumentation is.
- **Per operation rather than per run.** https://github.com/rapidsai/kvikio/issues/1045 asks for the attempt count and the final status on the `Observation` itself. The retry counters here answer whether a run was throttled and for how long, not which read it was, so that issue stays open.

### DIFF
--- a/cpp/include/kvikio/shim/libcurl.hpp
+++ b/cpp/include/kvikio/shim/libcurl.hpp
@@ -211,6 +211,23 @@ __attribute__((noinline)) inline std::string fix_conda_file_path_hack(std::strin
 }
 }  // namespace detail
 
+namespace detail {
+
+/**
+ * @brief Record what opening a connection cost a finished transfer.
+ *
+ * What resolving, connecting and shaking hands took, which a transfer that reused a connection
+ * paid nothing for.
+ *
+ * libcurl measures this whether or not anybody asks, and reports the phases cumulatively from the
+ * start of the transfer, so they are differenced here.
+ *
+ * @param easy The handle the transfer ran on, after it completed.
+ */
+void count_http_connection_of(CURL* easy) noexcept;
+
+}  // namespace detail
+
 /**
  * @brief Create a new curl handle.
  *
--- a/cpp/include/kvikio/statistics/counters.hpp
+++ b/cpp/include/kvikio/statistics/counters.hpp
@@ -1,0 +1,165 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <string>
+#include <utility>
+
+#include <kvikio/detail/observation_recorder.hpp>
+#include <kvikio/observation.hpp>
+#include <kvikio/shim/utils.hpp>
+
+namespace KVIKIO_EXPORT kvikio {
+namespace statistics {
+
+/**
+ * @brief Which rows `report()` prints.
+ */
+enum class ReportRows : std::uint8_t {
+  /// Only the backends and subsystems the run used.
+  USED,
+  /// Every row, including the ones the run never used.
+  ALL,
+};
+
+/**
+ * @brief Totals for work that belongs to no single operation.
+ *
+ * An `Observation` records one operation, and part of what I/O costs does not fit there, either
+ * because nothing correlates it with one operation or because it is shared between many.
+ *
+ * These counters count from the moment the process starts, whether or not anybody is watching, so
+ * there is nothing to enable and nothing to switch off. Reading them is `counters()`, and the cost
+ * of an interval is the difference between two readings.
+ */
+struct Counters {
+  /// Remote file sizes asked for, which is an HTTP round trip each.
+  std::uint64_t remote_size_probes{};
+  /// Time spent waiting for them.
+  Duration remote_size_probing{};
+
+  /// Connections libcurl opened, as opposed to reused. Against the requests a run made, this is
+  /// whether connections are being reused at all, which against a TLS endpoint dominates
+  /// everything else.
+  std::uint64_t http_connections{};
+  /// Time those connections spent resolving, connecting, and shaking hands. libcurl measures
+  /// these whether or not anybody asks, so reading them costs nothing.
+  Duration http_dns{};
+  Duration http_tcp{};
+  Duration http_tls{};
+
+  /// Requests the endpoint turned away with a retryable error, and the time spent sleeping before
+  /// trying again. A request that was retried twice counts twice. Nothing else records this, since
+  /// a retry that eventually succeeds is reported as a success.
+  std::uint64_t http_retries{};
+  Duration http_retry_backoff{};
+
+  /**
+   * @brief The difference between this reading and an earlier one.
+   *
+   * @param previous An earlier reading.
+   * @return What was spent in between, saturating at zero.
+   */
+  [[nodiscard]] Counters since(Counters const& previous) const noexcept;
+
+  /**
+   * @brief Whether anything was counted at all.
+   *
+   * @return True if every counter is zero.
+   */
+  [[nodiscard]] bool empty() const noexcept;
+
+  /**
+   * @brief Serialise to JSON.
+   *
+   * @return A JSON object as a string.
+   */
+  [[nodiscard]] std::string to_json() const;
+
+  /**
+   * @brief Format a human-readable report.
+   *
+   * Grouped by subsystem, and a group the run never touched is left out, so an empty reading
+   * formats as an empty string.
+   *
+   * @code
+   *   http size probes     12 probes, 600 ms
+   *   http handshake       128 connections, 40 ms dns, 900 ms tcp, 1.90 s tls
+   * @endcode
+   *
+   * @param rows Which rows to print.
+   * @return The report, newline-terminated, or empty.
+   */
+  [[nodiscard]] std::string report(ReportRows rows = ReportRows::USED) const;
+};
+
+/**
+ * @brief Every counter as it stands now.
+ *
+ * @return The running totals.
+ */
+[[nodiscard]] Counters counters() noexcept;
+
+}  // namespace statistics
+
+namespace detail {
+
+/**
+ * @brief Times a scope and records the duration however the scope is left.
+ *
+ * @code
+ * detail::ScopedTimer const probe{detail::count_remote_size_probe};
+ * curl.perform();  // Counted whether it returns or throws.
+ * @endcode
+ *
+ * @tparam Record What to hand the duration to. A `count_*()` below, or a lambda for one that
+ * takes more than a duration.
+ */
+template <typename Record>
+class ScopedTimer {
+ public:
+  explicit ScopedTimer(Record record) noexcept : _record{std::move(record)}, _started{now()} {}
+
+  ~ScopedTimer() noexcept { _record(now() - _started); }
+
+  ScopedTimer(ScopedTimer const&)            = delete;
+  ScopedTimer& operator=(ScopedTimer const&) = delete;
+
+ private:
+  Record _record;
+  TimePoint _started;
+};
+
+/**
+ * @brief Record asking a remote endpoint how big a file is.
+ *
+ * @param probing How long the round trip took.
+ */
+void count_remote_size_probe(Duration probing) noexcept;
+
+/**
+ * @brief Record what a finished HTTP transfer spent getting connected.
+ *
+ * @param connections Connections opened, which is zero when one was reused.
+ * @param dns Time resolving the name.
+ * @param tcp Time establishing the transport connection.
+ * @param tls Time shaking hands, or zero without TLS.
+ */
+void count_http_connection(std::uint64_t connections,
+                           Duration dns,
+                           Duration tcp,
+                           Duration tls) noexcept;
+
+/**
+ * @brief Record an HTTP request that hit a retryable error and will be tried again.
+ *
+ * @param backoff How long the next attempt waits before it goes out.
+ */
+void count_http_retry(Duration backoff) noexcept;
+
+}  // namespace detail
+}  // namespace KVIKIO_EXPORT kvikio
--- a/cpp/include/kvikio/statistics/summary.hpp
+++ b/cpp/include/kvikio/statistics/summary.hpp
@@ -15,6 +15,7 @@
 
 #include <kvikio/observation.hpp>
 #include <kvikio/shim/utils.hpp>
+#include <kvikio/statistics/counters.hpp>
 
 /**
  * @brief KvikIO namespace.
@@ -76,6 +77,14 @@ struct Summary {
   /// What each backend carried, indexed by `IoBackend`. Compatibility mode decides per call
   /// whether a read reaches cuFile or falls back to POSIX, so this is where that shows.
   std::array<BackendTotals, num_io_backends> by_backend{};
+
+  /**
+   * @brief The work in the span that belongs to no single operation.
+   *
+   * The counters run for the life of the process, and this is the part of them that falls inside
+   * the span.
+   */
+  Counters counters{};
 
   /// The operations' durations added up, every operation counted.
   Duration total_duration{};
@@ -188,9 +197,11 @@ struct Summary {
    *   ...
    * @endcode
    *
+   * @param rows Which rows to print. Under `ReportRows::USED` a backend the run never reached and a
+   * counter group it never touched are left out.
    * @return The report, one field per line, newline-terminated.
    */
-  [[nodiscard]] std::string report() const;
+  [[nodiscard]] std::string report(ReportRows rows = ReportRows::USED) const;
 };
 
 /**
@@ -354,6 +365,11 @@ class SummaryMonitor final : private kvikio::Monitor {
   /// Serializes `stop()`, which cannot use `_mutex`, since `unregister_monitor()` waits for
   /// notifications that take it.
   std::mutex _stopping;
+
+  /// The internal counters as they stood when the span began, so the reading is a difference,
+  /// and as they stood when it ended, so a stopped summary does not keep growing.
+  Counters _counters_at_start{};
+  Counters _counters_at_stop{};
 
   /**
    * @brief What every reading is a copy of.
--- a/cpp/src/detail/multi_poll_reactor.cpp
+++ b/cpp/src/detail/multi_poll_reactor.cpp
@@ -28,6 +28,7 @@
 #include <kvikio/remote_handle.hpp>
 #include <kvikio/shim/cuda.hpp>
 #include <kvikio/shim/libcurl.hpp>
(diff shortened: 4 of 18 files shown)


cpp/src/shim/libcurl.cpp
```diff
@@ -122,6 +124,31 @@ std::string CurlHandle::error_message() const
...
+  if (curl_easy_getinfo(easy, CURLINFO_NUM_CONNECTS, &connections) != CURLE_OK) { return; }
+  // Zero means the connection was reused, so nothing was paid here.
+  if (connections <= 0) { return; }
+
+  curl_off_t namelookup{0};
+  curl_off_t connect{0};
+  curl_off_t appconnect{0};
+  curl_easy_getinfo(easy, CURLINFO_NAMELOOKUP_TIME_T, &namelookup);
+  curl_easy_getinfo(easy, CURLINFO_CONNECT_TIME_T, &connect);
+  curl_easy_getinfo(easy, CURLINFO_APPCONNECT_TIME_T, &appconnect);
```
kingcrimsontianyu: Nit: The doc suggests that:

> The data is stored accordingly and can be relied upon only if this function returns CURLE_OK.

So we may have a macro checking the error code, something like (untested):
```cpp
#define KVIKIO_CHECK_CURL_EASY(error) \
kvikio::detail::check_curl_easy(error, __LINE__, __FILE__)

inline void check_curl_easy(CURLcode error, int line_number, char const* filename)
{
  if (error != CURLE_OK) {
    throw Exception{std::string{"libcurl error at: "} + filename + ":" + std::to_string(line_number) +
                    ": " + std::string(curl_easy_strerror(error))};
  }
}
```
Or for simplicity, just have a function-scope lambda checking the error code, without the line number / file name arguments.

I know in `CurlHandle` we also use a buffer to store more detailed error message, but that could be a bit of overkill for `get_info` purpose.

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+  std::uint64_t remote_size_probes{};
+  /// Time spent waiting for them.
+  Duration remote_size_probing{};
+
+  /// Connections libcurl opened, as opposed to reused. Against the requests a run made, this is
+  /// whether connections are being reused at all, which against a TLS endpoint dominates
+  /// everything else.
+  std::uint64_t http_connections{};
+  /// Time those connections spent resolving, connecting, and shaking hands. libcurl measures
+  /// these whether or not anybody asks, so reading them costs nothing.
```
kingcrimsontianyu, line 50: That's good to know.

cpp/src/shim/libcurl.cpp
```diff
@@ -122,6 +124,31 @@ std::string CurlHandle::error_message() const
   return std::string{_errbuf};
 }
 
+namespace detail {
+
+void count_http_connection_of(CURL* easy) noexcept
+{
+  auto const micros = [](curl_off_t us) { return std::chrono::microseconds{us}; };
```
kingcrimsontianyu: Nit: Alternatively, we could simply have a function-scope `using std::chrono::microseconds`, and use `microseconds{...}` below.

cpp/src/statistics/counters.cpp
```diff
@@ -0,0 +1,185 @@
...
+    .http_tcp            = spent(http_tcp, previous.http_tcp),
+    .http_tls            = spent(http_tls, previous.http_tls),
+    .http_retries        = spent(http_retries, previous.http_retries),
+    .http_retry_backoff  = spent(http_retry_backoff, previous.http_retry_backoff),
+  };
+}
+
+bool Counters::empty() const noexcept
+{
+  return remote_size_probes == 0 && http_connections == 0 && http_retries == 0;
```
kingcrimsontianyu: Should other `Duration` type members be checked too?

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -353,6 +354,7 @@ void MultiPollReactor::io_thread_main()
                       std::runtime_error);
         auto transfer = std::move(it->second);
         _in_flight.erase(it);
+        detail::count_http_connection_of(easy);
```
kingcrimsontianyu: `detail::` can be dropped here.

madsbk: Dropped.

cpp/src/statistics/counters.cpp
madsbk: Yes, and it now does. The docstring already claimed every counter, so the implementation was the wrong one. They cannot be nonzero on their own today, since each recorder writes a count and its duration together, but nothing holds that in place.

cpp/src/shim/libcurl.cpp
madsbk: Done, `using std::chrono::microseconds` instead.

cpp/src/shim/libcurl.cpp
madsbk: Good catch, added a `noexcept` helper.

`KVIKIO_CHECK_CURL_EASY` is worth having where throwing is right, the `setopt` and `perform` paths.

cpp/include/kvikio/shim/libcurl.hpp
```diff
@@ -211,6 +211,23 @@ __attribute__((noinline)) inline std::string fix_conda_file_path_hack(std::strin
...
+
+/**
+ * @brief Record what opening a connection cost a finished transfer.
+ *
+ * What resolving, connecting and shaking hands took, which a transfer that reused a connection
+ * paid nothing for.
+ *
+ * libcurl measures this whether or not anybody asks, and reports the phases cumulatively from the
+ * start of the transfer, so they are differenced here.
+ *
```
wence-: Please reword this docstring, I find it very hard to follow what it means.

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+ */
+enum class ReportRows : std::uint8_t {
+  /// Only the backends and subsystems the run used.
+  USED,
+  /// Every row, including the ones the run never used.
+  ALL,
+};
+
+/**
+ * @brief Totals for work that belongs to no single operation.
```
wence-: nit (grammar): `Totals` is the object of this sentence so:


```suggestion
 * @brief Totals for work that belong to no single operation.
```

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * An `Observation` records one operation, and part of what I/O costs does not fit there, either
+ * because nothing correlates it with one operation or because it is shared between many.
```
wence-: This reads like a non-sequitur because we're not thinking about observations at the moment.

Perhaps something like:


> These `Counters` record aggregate statistics about I/O operations that do not naturally fit the classification required by the `Observation` object, which tracks statistics for a single I/O operation. Here instead, we record information that is not tied to a single `Observation` (perhaps because it is not driven by a single I/O request or because it is shared between many).

?

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * These counters count from the moment the process starts, whether or not anybody is watching, so
+ * there is nothing to enable and nothing to switch off. Reading them is `counters()`, and the cost
+ * of an interval is the difference between two readings.
```
wence-: Please reword. "cost of an interval" is a weird phrasing. `Reading them is counters()` is also very odd.

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+ */
+struct Counters {
+  /// Remote file sizes asked for, which is an HTTP round trip each.
+  std::uint64_t remote_size_probes{};
+  /// Time spent waiting for them.
+  Duration remote_size_probing{};
```
wence-:
```suggestion
  /// Number of remote file size requests.
  std::uint64_t remote_size_probes{};
  /// Time spent waiting for remote file sizes.
  Duration remote_size_probing{};
```

Or something.

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+  /// Connections libcurl opened, as opposed to reused. Against the requests a run made, this is
+  /// whether connections are being reused at all, which against a TLS endpoint dominates
+  /// everything else.
```
wence-: The second sentence is very hard to pass. Perhaps the first one should just be "Number of connections opened by libcurl"?

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+  std::uint64_t http_connections{};
+  /// Time those connections spent resolving, connecting, and shaking hands. libcurl measures
```
wence-: Also reword this to be slightly more standalone.

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+  /// these whether or not anybody asks, so reading them costs nothing.
+  Duration http_dns{};
+  Duration http_tcp{};
+  Duration http_tls{};
```
wence-: Document each of these separately. If we're going to have doxygen docstrings we need to attach the information to each thing.

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+  /// Requests the endpoint turned away with a retryable error, and the time spent sleeping before
+  /// trying again. A request that was retried twice counts twice. Nothing else records this, since
+  /// a retry that eventually succeeds is reported as a success.
+  std::uint64_t http_retries{};
+  Duration http_retry_backoff{};
```
wence-: Split the documentation to the two things, also reword.

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * @brief The difference between this reading and an earlier one.
+   *
+   * @param previous An earlier reading.
+   * @return What was spent in between, saturating at zero.
```
wence-: "spent" is the wrong verb I think. Also, since these counters only ever increase, is "saturating at zero" necessary?

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+   *
+   * @return A JSON object as a string.
+   */
+  [[nodiscard]] std::string to_json() const;
+
+  /**
+   * @brief Format a human-readable report.
+   *
+   * Grouped by subsystem, and a group the run never touched is left out, so an empty reading
+   * formats as an empty string.
```
wence-: First appearance of the phrase "subsystem".

wence-: Also, if the `ReportRows` are `ALL` are empty readings left out?

cpp/include/kvikio/statistics/counters.hpp
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * @brief Times a scope and records the duration however the scope is left.
+ *
+ * @code
+ * detail::ScopedTimer const probe{detail::count_remote_size_probe};
+ * curl.perform();  // Counted whether it returns or throws.
+ * @endcode
+ *
+ * @tparam Record What to hand the duration to. A `count_*()` below, or a lambda for one that
+ * takes more than a duration.
```
wence-: Please document properly. IIUC the `Record` must always be something that models a `noexcept` function `record :: Duration -> void`.

cpp/include/kvikio/statistics/summary.hpp
```diff
@@ -188,9 +197,11 @@ struct Summary {
    *   ...
    * @endcode
    *
+   * @param rows Which rows to print. Under `ReportRows::USED` a backend the run never reached and a
```
wence-: Here we use `backend` before we used `subsystem`?

cpp/include/kvikio/statistics/summary.hpp
```diff
@@ -77,6 +78,14 @@ struct Summary {
   /// whether a read reaches cuFile or falls back to POSIX, so this is where that shows.
   std::array<BackendTotals, num_io_backends> by_backend{};
 
+  /**
+   * @brief The work in the span that belongs to no single operation.
+   *
+   * The counters run for the life of the process, and this is the part of them that falls inside
+   * the span.
+   */
+  Counters counters{};
```
wence-: If the counters run for the process lifetime how can the counters also only be the part that falls inside the span?

cpp/include/kvikio/statistics/summary.hpp
```diff
@@ -355,6 +366,11 @@ class SummaryMonitor final : private kvikio::Monitor {
   /// notifications that take it.
   std::mutex _stopping;
 
+  /// The internal counters as they stood when the span began, so the reading is a difference,
+  /// and as they stood when it ended, so a stopped summary does not keep growing.
+  Counters _counters_at_start{};
+  Counters _counters_at_stop{};
```
wence-: As above: doxygen docstrings must be attached individually to their objects.

cpp/src/shim/libcurl.cpp
```diff
@@ -122,6 +124,39 @@ std::string CurlHandle::error_message() const
...
+  if (curl_easy_getinfo(easy, CURLINFO_NUM_CONNECTS, &connections) != CURLE_OK) { return; }
+  // Zero means the connection was reused, so nothing was paid here.
+  if (connections <= 0) { return; }
+
+  auto const namelookup = timing_of(easy, CURLINFO_NAMELOOKUP_TIME_T);
+  auto const connect    = timing_of(easy, CURLINFO_CONNECT_TIME_T);
+  auto const appconnect = timing_of(easy, CURLINFO_APPCONNECT_TIME_T);
+
+  auto const tcp = connect > namelookup ? microseconds{connect - namelookup} : Duration::zero();
+  auto const tls = appconnect > connect ? microseconds{appconnect - connect} : Duration::zero();
```
wence-: Can we not use `std::max` here?

cpp/src/shim/libcurl.cpp
```diff
@@ -122,6 +124,39 @@ std::string CurlHandle::error_message() const
...
+  using std::chrono::microseconds;
+
+  long connections{0};
+  if (curl_easy_getinfo(easy, CURLINFO_NUM_CONNECTS, &connections) != CURLE_OK) { return; }
+  // Zero means the connection was reused, so nothing was paid here.
+  if (connections <= 0) { return; }
+
+  auto const namelookup = timing_of(easy, CURLINFO_NAMELOOKUP_TIME_T);
+  auto const connect    = timing_of(easy, CURLINFO_CONNECT_TIME_T);
+  auto const appconnect = timing_of(easy, CURLINFO_APPCONNECT_TIME_T);
```
wence-: `timing_of` could return `microseconds` rather than `curl_off_t`?

cpp/src/statistics/counters.cpp
```diff
@@ -0,0 +1,188 @@
...
+namespace kvikio {
+
+namespace {
+
+/**
+ * @brief The counters themselves.
+ *
+ * Relaxed throughout. Each is a running total that nothing else is ordered against, and a reading
+ * taken while another thread is incrementing is a snapshot of a moving target either way.
+ */
```
wence-: This also wants reworded I think

cpp/src/statistics/counters.cpp
```diff
@@ -0,0 +1,188 @@
...
+ * still find somewhere to count.
+ */
+Atomics& atomics() noexcept
+{
+  static auto* instance = new Atomics{};
+  return *instance;
+}
+
+/// Add a duration to a counter of nanoseconds.
+void add(std::atomic<std::int64_t>& counter, Duration duration) noexcept
```
wence-: nit: we should probably mark this as `[[attribute(inline=always)]]` or however you spell it.

cpp/src/statistics/counters.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+}  // namespace
+
+namespace detail {
+
+void count_remote_size_probe(Duration probing) noexcept
+{
+  auto& all = atomics();
+  all.remote_size_probes.fetch_add(1, std::memory_order_relaxed);
+  add(all.remote_size_probing_ns, probing);
```
wence-: Yeah, or just remove `add` and inline the `all.remote_size_probing_ns.fetch_add(probing.count(), std::memory_order_relaxed)` here?

kingcrimsontianyu (approved):
Looks great!

bdice (approved):

madsbk:
Thanks @kingcrimsontianyu and @bdice

PR #1049, opened 2026-08-27
Add runtime setters for MULTI_POLL reactor count, dispatch, and request ceiling

Matt711:
Adds setters for `remote_io_num_reactors`, `remote_io_reactor_dispatch`, and `remote_io_max_concurrent_requests`

- Need for cudf-polars  https://github.com/NVIDIA/cudf/pull/23839

### DIFF
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -488,6 +488,17 @@ class defaults {
   [[nodiscard]] static unsigned int remote_io_num_reactors();
 
   /**
+   * @brief Set the number of reactor threads used by the `MULTI_POLL` remote I/O backend at
+   * runtime, overriding `KVIKIO_REMOTE_IO_NUM_REACTORS`.
+   *
+   * Takes effect only if set before the `MULTI_POLL` reactor pool is first used (the pool, like
+   * the default thread pool, is created lazily on first use and is never rebuilt).
+   *
+   * @param num_reactors The number of reactor threads. Must be a positive integer.
+   */
+  static void set_remote_io_num_reactors(unsigned int num_reactors);
+
+  /**
    * @brief How sub-ranges of one `pread()` are distributed across reactor threads under the
    * `MULTI_POLL` remote I/O backend.
    *
@@ -499,6 +510,17 @@ class defaults {
    * @return The reactor dispatch policy.
    */
   [[nodiscard]] static RemoteReactorDispatch remote_io_reactor_dispatch();
+
+  /**
+   * @brief Set the reactor dispatch policy used by the `MULTI_POLL` remote I/O backend at
+   * runtime, overriding `KVIKIO_REMOTE_IO_REACTOR_DISPATCH`.
+   *
+   * Takes effect only if set before the `MULTI_POLL` reactor pool is first used (the pool, like
+   * the default thread pool, is created lazily on first use and is never rebuilt).
+   *
+   * @param dispatch The reactor dispatch policy.
+   */
+  static void set_remote_io_reactor_dispatch(RemoteReactorDispatch dispatch);
 
   /**
    * @brief Maximum number of concurrent in-flight requests across all reactor threads under the
@@ -519,6 +541,18 @@ class defaults {
    * @return The configured concurrent-request ceiling, or 0 for unlimited.
    */
   [[nodiscard]] static std::size_t remote_io_max_concurrent_requests();
+
+  /**
+   * @brief Set the maximum number of concurrent in-flight requests across all reactor threads
+   * under the `MULTI_POLL` remote I/O backend at runtime, overriding
+   * `KVIKIO_REMOTE_IO_MAX_CONCURRENT_REQUESTS`.
+   *
+   * Takes effect only if set before the `MULTI_POLL` reactor pool is first used (the pool, like
+   * the default thread pool, is created lazily on first use and is never rebuilt).
+   *
+   * @param max_requests The concurrent-request ceiling. 0 means unlimited.
+   */
+  static void set_remote_io_max_concurrent_requests(std::size_t max_requests);
 };
 
 }  // namespace kvikio
--- a/cpp/include/kvikio/defaults.hpp
+++ b/cpp/include/kvikio/defaults.hpp
@@ -491,7 +491,7 @@ class defaults {
    * @brief Set the number of reactor threads used by the `MULTI_POLL` remote I/O backend at
    * runtime, overriding `KVIKIO_REMOTE_IO_NUM_REACTORS`.
    *
-   * The pool, like the default thread pool, is created lazily on first use and is never rebuilt.
+   * The pool is created lazily on first use and is never rebuilt or resized.
    *
    * @param num_reactors The number of reactor threads. Must be a positive integer.
    *
@@ -516,7 +516,7 @@ class defaults {
    * @brief Set the reactor dispatch policy used by the `MULTI_POLL` remote I/O backend at
    * runtime, overriding `KVIKIO_REMOTE_IO_REACTOR_DISPATCH`.
    *
-   * The pool, like the default thread pool, is created lazily on first use and is never rebuilt.
+   * The pool is created lazily on first use and is never rebuilt or resized.
    *
    * @param dispatch The reactor dispatch policy.
    *
@@ -549,7 +549,7 @@ class defaults {
    * under the `MULTI_POLL` remote I/O backend at runtime, overriding
    * `KVIKIO_REMOTE_IO_MAX_CONCURRENT_REQUESTS`.
    *
-   * The pool, like the default thread pool, is created lazily on first use and is never rebuilt.
+   * The pool is created lazily on first use and is never rebuilt or resized.
    *
    * @param max_requests The concurrent-request ceiling. 0 means unlimited.
    *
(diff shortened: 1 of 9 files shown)


cpp/src/defaults.cpp
```diff
@@ -304,13 +304,30 @@ void defaults::set_remote_io_backend(RemoteIOBackend backend)
 
 unsigned int defaults::remote_io_num_reactors() { return instance()->_remote_io_num_reactors; }
 
+void defaults::set_remote_io_num_reactors(unsigned int num_reactors)
+{
+  KVIKIO_EXPECT(
+    num_reactors > 0, "remote_io_num_reactors must be a positive integer", std::invalid_argument);
+  instance()->_remote_io_num_reactors = num_reactors;
+}
+
```
madsbk, line 322: I think it would be helpful if we could throw if the reactor pool has already been started?
Maybe add something like:
```c++
// multi_poll_reactor.hpp
static bool is_instantiated() noexcept;

// multi_poll_reactor.cpp
namespace { std::atomic<bool> _pool_instantiated{false}; }

bool MultiReactorPool::is_instantiated() noexcept
{
  return _pool_instantiated.load(std::memory_order_acquire);
}

MultiReactorPool::MultiReactorPool() : _dispatch{defaults::remote_io_reactor_dispatch()}
{
  ...
  _pool_instantiated.store(true, std::memory_order_release);  // end of ctor
}
```

madsbk (changes requested):
Overall looks good.
 - update the stale sentence at `runtime_settings.rst:82`
 - add missing docs for three newly user-facing settings

cpp/include/kvikio/defaults.hpp
```diff
@@ -487,6 +487,17 @@ class defaults {
    */
   [[nodiscard]] static unsigned int remote_io_num_reactors();
 
+  /**
+   * @brief Set the number of reactor threads used by the `MULTI_POLL` remote I/O backend at
+   * runtime, overriding `KVIKIO_REMOTE_IO_NUM_REACTORS`.
+   *
+   * Takes effect only if set before the `MULTI_POLL` reactor pool is first used (the pool, like
+   * the default thread pool, is created lazily on first use and is never rebuilt).
```
wence-: nit: Like Mads, I think we should throw if this set call would not take effect.

cpp/include/kvikio/defaults.hpp
```diff
@@ -500,6 +511,17 @@ class defaults {
    */
   [[nodiscard]] static RemoteReactorDispatch remote_io_reactor_dispatch();
 
+  /**
+   * @brief Set the reactor dispatch policy used by the `MULTI_POLL` remote I/O backend at
+   * runtime, overriding `KVIKIO_REMOTE_IO_REACTOR_DISPATCH`.
+   *
+   * Takes effect only if set before the `MULTI_POLL` reactor pool is first used (the pool, like
+   * the default thread pool, is created lazily on first use and is never rebuilt).
```
wence-: Same here.

cpp/include/kvikio/defaults.hpp
```diff
@@ -519,6 +541,18 @@ class defaults {
...
    */
   [[nodiscard]] static std::size_t remote_io_max_concurrent_requests();
+
+  /**
+   * @brief Set the maximum number of concurrent in-flight requests across all reactor threads
+   * under the `MULTI_POLL` remote I/O backend at runtime, overriding
+   * `KVIKIO_REMOTE_IO_MAX_CONCURRENT_REQUESTS`.
+   *
+   * Takes effect only if set before the `MULTI_POLL` reactor pool is first used (the pool, like
+   * the default thread pool, is created lazily on first use and is never rebuilt).
```
wence-: And here.

cpp/src/defaults.cpp
```diff
@@ -304,13 +304,30 @@ void defaults::set_remote_io_backend(RemoteIOBackend backend)
...
 RemoteReactorDispatch defaults::remote_io_reactor_dispatch()
 {
   return instance()->_remote_io_reactor_dispatch;
 }
 
+void defaults::set_remote_io_reactor_dispatch(RemoteReactorDispatch dispatch)
+{
+  instance()->_remote_io_reactor_dispatch = dispatch;
```
wence-, line 336: Yeah, I think especially because the getters will return this value (even if it was not the one used for setting up the multi-poll instance), we should throw. Otherwise inspecting the defaults later to report behaviour will be misleading.

wence- (changes requested):

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: I agree with the suggestion. Just want to point out that for the BS thread pool used for local I/O and `easy` backend, we do support changing the number of threads midway at runtime, which would block the calling thread, drain the task queue, destroy the worker threads, and then recreate a new set of worker threads. So I think the phrasing "like the default thread pool" should be removed from the doc string here.

kingcrimsontianyu (approved):

cpp/src/defaults.cpp
```diff
@@ -12,6 +12,7 @@
 
 #include <kvikio/compat_mode.hpp>
 #include <kvikio/defaults.hpp>
+#include <kvikio/detail/multi_poll_reactor.hpp>
```
madsbk, line 16: This need a guard, I don't think it will work with `-DKvikIO_REMOTE_SUPPORT=OFF` ?

madsbk (dismissed):
Since we cannot reset the pool size, the defaults context manager is now broken:
```python
with kvikio.defaults.set("remote_io_num_reactors", 8):
    remote_file.pread(...)
```
It fails when trying to reset `remote_io_num_reactors` on exit.

Matt711:
> ```python
> kvikio.defaults.set("remote_io_num_reactors", 8)
> ```

Can we just not support context managers for these settings? I did that in https://github.com/rapidsai/kvikio/pull/1049/commits/7373370dba5af0e0871a3999c39360700ad7619d

cpp/src/defaults.cpp
Matt711: Done in https://github.com/rapidsai/kvikio/pull/1049/commits/7373370dba5af0e0871a3999c39360700ad7619d

python/kvikio/kvikio/defaults.py
```diff
@@ -25,6 +38,17 @@ def __init__(self, config: dict[str, str]):
...
+            self._old_properties
+        )
+        if process_lifetime_keys:
+            raise ValueError(
+                f"{sorted(process_lifetime_keys)} cannot be set with a `with` block. "
+                "Once the MULTI_POLL reactor pool has started (i.e. the first remote "
+                "I/O has been issued), these properties are fixed for the remaining "
+                "process lifetime and cannot be reverted. Use "
+                "kvikio.defaults.set(...) without a `with` block instead."
+            )
```
wence-: This error message is somewhat misleading. Using `set_` doesn't solve the underlying problem.

Why don't we just not allow introspecting the multipoll setting _at all_ from the context manager? That is, remove the names from the set of permissible options.

python/kvikio/tests/test_defaults.py
```diff
@@ -184,3 +184,60 @@ def test_remote_io_backend():
...
+
+def test_remote_io_max_concurrent_requests():
+    before = kvikio.defaults.get("remote_io_max_concurrent_requests")
+
+    try:
+        kvikio.defaults.set("remote_io_max_concurrent_requests", 128)
+        assert kvikio.defaults.get("remote_io_max_concurrent_requests") == 128
+    finally:
+        kvikio.defaults.set("remote_io_max_concurrent_requests", before)
+    assert kvikio.defaults.get("remote_io_max_concurrent_requests") == before
```
wence-, line 250: All these tests are not quite as good as one might think. They can, incorrectly, pass if `set` doesn't work but `before == 128`.

Shall we ensure that we're actually always modifying `before` and using that modified value for the `set`?

python/kvikio/kvikio/defaults.py
```diff
@@ -7,6 +7,19 @@
...
+# `__exit__` that can no longer be honored, so `ConfigContextManager` refuses to be
+# entered as a `with` block when one of these is among the properties being set.
+_PROCESS_LIFETIME_PROPERTIES = frozenset(
+    {
+        "remote_io_num_reactors",
+        "remote_io_reactor_dispatch",
+        "remote_io_max_concurrent_requests",
+    }
+)
+
```
wence-, line 22: Yeah, as below, I think we can simplify this by just not even advertising these are permissible options for the context manager.

wence- (changes requested):

python/kvikio/kvikio/defaults.py
Matt711: Ugh...are you thinking something like https://github.com/rapidsai/kvikio/pull/1049/commits/a2802022a2e803c72c9718bfa80c3f4e2b05bee4?

python/kvikio/kvikio/defaults.py
Matt711: See https://github.com/rapidsai/kvikio/pull/1049#discussion_r3926495708

wence- (approved):

python/kvikio/tests/test_defaults.py
Matt711: I updated the tests in https://github.com/rapidsai/kvikio/pull/1049/commits/a2802022a2e803c72c9718bfa80c3f4e2b05bee4

Issue #1044, opened 2026-08-25
Expose thread pool queue/running/total task counts


Matt711:
kvikio's thread pool is `BS::thread_pool` (`cpp/include/kvikio/threadpool_wrapper.hpp`), which already implements `get_tasks_queued()`, `get_tasks_running()`, and `get_tasks_total()`. Real-time utilization for the exact resource `kvikio_nthreads` (default 256) controls. None of it is exposed past `kvikio::defaults::thread_pool()` (`cpp/include/kvikio/defaults.hpp`). No C++ wrapper, no Python binding. `python/kvikio/kvikio/_lib/defaults.pyx` only exposes get/set for `thread_pool_nthreads`.

Add a thin C++ wrapper and a `.pyx` binding.

Why it matters: today the only signal is `kvikio.Summary`, a post-hoc aggregate over the whole query. It can't tell apart a saturated pool (raising thread count would help) from a pool with spare capacity (raising thread count won't help, the real constraint is upstream). A `kvikio_nthreads` sweep that shows no improvement is ambiguous between those two cases without a live queued/running count. This is also how we'd tell whether `max_concurrent_io_tasks` or `kvikio_nthreads` is actually the one binding for a given regression, instead of guessing which layer to tune.

Issue #967, opened 2026-05-27
Structured logs for each read request


TomAugspurger:
Kvikio currently emits [some logs](https://docs.rapids.ai/api/kvikio/nightly/runtime_settings/#logging-kvikio-log-level-kvikio-log-file) related to reads. I'm working on optimizing cudf-polars IO, and these logs had a couple of shortcomings:

1. The logs aren't machine readable. The lines are formatted like `FileHandle::pread(buf=..., size=..., ...)`. It didn't have the path being read or the start/end times.
2. Logical reads are logged, rather than physical reads (I think). kvikio automatically breaks large multi-MB reads into smaller chunks. Both are interesting, but I was especially interested in the physical reads.

Have we considered emitting structured logs (in JSON format, say, but this might be configurable)? https://github.com/rapidsai/kvikio/compare/main...TomAugspurger:kvikio:tom/read-log-traces?expand=1 has a POC for that. The lines include

- source (path / URL)
- start, end (int counters)
- offset, size
- threadId
- bytesRead (the bytes actually read, which might differ from size)
- backend (local or remote)
- status
- isDeviceBuffer (maybe this should be a "destination" enum?)
- requestId: correlate with the logical request that induced this physical read (a logical read might be split into multiple physical reads)

It also includes a (private) tool for converting the structured logs to `chrome://tracing` format, which can be loaded into tools like https://ui.perfetto.dev/. https://gist.github.com/TomAugspurger/135c38917a3a4cbe0cfc6eb80b503bff is an example file, shown here:

<img width="1482" height="1266" alt="Image" src="https://github.com/user-attachments/assets/b54541fc-b269-4ef5-aa5f-ae130015f1e0" />

If this is helpful, I can clean things up and submit a pR.

TomAugspurger:
Discussed a bit today and one potential way to improve the POC: we should integrate these new logs into the existing logging framework:

- use the `TRACE` log level to control whether these physical reads are logged. This pairs nicely with the other settings: `=DEBUG` is logical and `=TRACE` is physical (and logical)
- Separate the data logged from the output format with a new parameter. Then we can have all the logs emitted as JSON or the custom strings we've used up till now

PR #1050, opened 2026-08-28
Add physical observations

madsbk:
This PR introduces physical observations. One covers a single transfer occupying a thread or a connection: one thread-pool task locally, one HTTP range request remotely, one per attempt when a request is retried. It starts when a worker picks the task up or when the request goes on the wire, and ends when that chunk is done, so its span is the transfer rather than the call that was waiting for a thread.

`Summary` is unchanged. No new field, no new accessor. `SummaryMonitor` takes an `ObservationKind` and computes the same statistics over whichever kind of observation you asked for.

```python
logical = kvikio.SummaryMonitor(kvikio.ObservationKind.LOGICAL)
physical = kvikio.SummaryMonitor(kvikio.ObservationKind.PHYSICAL)
```

### What is covered

Every byte of a call appears exactly once across its transfers. A monitor subscribed to one kind sees the same totals as a monitor subscribed to the other, so the two are comparable and a physical-only summary is never short.

That is what forces a record on the single-shot paths. `FileHandle::read()/write()`, `MmapHandle::read()` and `RemoteHandle::read()` do the work inline with no queue in front, so a physical record there duplicates the logical one exactly. It is emitted regardless.

| path | one observation is |
|---|---|
| local POSIX and GDS | one thread-pool task |
| mmap | one thread-pool task |
| remote `MULTI_POLL` | one sub-range request, one per attempt |
| remote `EASY_THREADPOOL` | one thread-pool task |

The bounce-buffer loop inside `posix_io.cpp` and the per-chunk device copies are deliberately not subdivided.

### Overhead

`monitor_count` becomes one count per kind, so a program with only a `SummaryMonitor` never builds a physical record. It pays 3.4 ns per task to find that out.

With a monitor subscribed, an observation costs about 60 ns whichever kind it is, since it is the same record through the same registry. On my local workstation, the four states interleaved round by round:

| workload | off (ns/call) | logical | physical | both |
|---|---|---|---|---|
| mmap read, 8 B, 1 task | 378 | +62 | +60 | +124 |
| host `pread`, 4 KiB, 1 task | 3952 | +211 | +60 | +277 |
| host `pread`, 16 MiB, 16 tasks | 1659380 | noise | noise | noise |

The last row is the one where the kinds differ in volume, sixteen physical records against one logical. Neither state separates from `off`: roughly 1 µs of observation on a 1.66 ms call, against a run-to-run spread of ±10 µs.

### Notable

`Observation` gains `parent_id`, a `std::optional<std::uint64_t>` holding the `id` of the logical operation a transfer belongs to. It is empty only when the call started with no monitor registered for logical observations, since there is no logical record to point at. Nothing else leaves it empty, so a consumer subscribed to both kinds can rely on it.

A retried request is one observation per attempt, so a backoff is a gap between two records rather than one long transfer. A failed attempt is recorded with zero bytes.

Within a request there is no progress reporting, so an unsplit GET appears at its completion however long it ran. `pread()` splits at `task_size` on both remote backends, 4 MB by default. `RemoteHandle::read()` does not split, so there a request is as large as the call.

Nothing subscribes to `PHYSICAL` by default, and `SummaryMonitor` still defaults to `LOGICAL`.

### Follow-ups

A trace monitor on the physical observations, writing one record per transfer for a timeline viewer. That is what makes `parent_id` visible to a user, and it covers what rapidsai/kvikio#967 asks for without a logger call on the I/O path.

Per-call queue wait, rapidsai/kvikio#1044, which is a consumer subscribed to both kinds and grouping by `parent_id`.


kingcrimsontianyu (approved):
Looks great overall!

PR #1051, opened 2026-08-28
Share DNS caches among easy handles to prevent DNS storm

kingcrimsontianyu:
This PR only applies to the easy thread pool backend.

Saturating a fast NIC needs a large `KVIKIO_NTHREADS`. Each easy handle resolves the endpoint on its own, and the number of DNS lookups grows with the thread pool. When too many happen at once, some queries go unanswered, and libcurl reports this as `CURLE_COULDNT_RESOLVE_HOST` ("Could not resolve host").

This PR helps to avoid this problem by sharing the DNS cache among easy handles. Two knobs are provided:
- `KVIKIO_REMOTE_SHARE_DNS_CACHE`: Whether to share DNS cache. Default to 1 (enabled).
- `KVIKIO_REMOTE_MAX_THREADS_PER_DNS_CACHE`: How many threads share a DNS cache. Default to 16.

### DIFF
--- a/cpp/include/kvikio/detail/curl_share.hpp
+++ b/cpp/include/kvikio/detail/curl_share.hpp
@@ -1,0 +1,76 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <array>
+#include <mutex>
+
+#include <curl/curl.h>
+
+namespace kvikio::detail {
+
+/**
+ * @brief Process-wide libcurl share holding a single DNS cache for all curl handles.
+ *
+ * An attached share overrides the multi handle's cache. `curl_multi_add_handle()` installs its own
+ * only when the easy handle has none, so `CURLOPT_SHARE` has to be set first. Setting it in the
+ * `CurlHandle` constructor guarantees that order.
+ *
+ * Setting the environment variable `KVIKIO_REMOTE_SHARE_DNS_CACHE` to a false value to disable
+ * cache sharing and prevent this class from construction.
+ */
+class CurlShare {
+ public:
+  /**
+   * @brief Get the process-wide share, creating it on first use.
+   */
+  static CurlShare& instance();
+
+  CurlShare(CurlShare const&)            = delete;
+  CurlShare& operator=(CurlShare const&) = delete;
+  CurlShare(CurlShare&&)                 = delete;
+  CurlShare& operator=(CurlShare&&)      = delete;
+
+  /**
+   * @brief The underlying share handle, for `curl_easy_setopt(CURLOPT_SHARE, ...)`.
+   */
+  [[nodiscard]] CURLSH* handle() const noexcept { return _share; }
+
+ private:
+  CurlShare();
+  ~CurlShare() = default;
+
+  /**
+   * @brief `CURLSHOPT_LOCKFUNC` callback. Takes the mutex guarding @p data.
+   *
+   * libcurl does no locking of its own, so without this callback the shared cache is accessed
+   * unsynchronized.
+   *
+   * @param handle The easy handle libcurl is serving. Unused. The mutex is chosen by @p data alone.
+   * @param data Which shared cache is about to be accessed.
+   * @param access Whether libcurl wants shared or exclusive access. Unused, since the DNS cache is
+   * locked exclusively.
+   * @param userptr The `CurlShare*` registered via `CURLSHOPT_USERDATA`.
+   */
+  static void lock_callback(CURL* handle,
+                            curl_lock_data data,
+                            curl_lock_access access,
+                            void* userptr);
+
+  /**
+   * @brief `CURLSHOPT_UNLOCKFUNC` callback. Releases the mutex taken by `lock_callback`.
+   *
+   * @param handle The easy handle libcurl is serving. Unused.
+   * @param data Which shared cache was accessed.
+   * @param userptr The `CurlShare*` registered via `CURLSHOPT_USERDATA`.
+   */
+  static void unlock_callback(CURL* handle, curl_lock_data data, void* userptr);
+
+  CURLSH* _share{nullptr};
+  // One mutex per shareable data kind in libcurl
+  std::array<std::mutex, CURL_LOCK_DATA_LAST> _mutexes;
+};
+
+}  // namespace kvikio::detail
--- a/cpp/include/kvikio/detail/curl_share.hpp
+++ b/cpp/include/kvikio/detail/curl_share.hpp
@@ -12,35 +12,36 @@
 namespace kvikio::detail {
 
 /**
- * @brief Process-wide libcurl share holding a single DNS cache for all curl handles.
+ * @brief A libcurl share handle holding a DNS cache for a group of curl easy handles.
  *
- * An attached share overrides the multi handle's cache. `curl_multi_add_handle()` installs its own
- * only when the easy handle has none, so `CURLOPT_SHARE` has to be set first. Setting it in the
- * `CurlHandle` constructor guarantees that order.
+ * An attached share handle overrides the multi handle's cache. `curl_multi_add_handle()` installs
+ * its own only when the easy handle has none, so `CURLOPT_SHARE` has to be set first. Setting it in
+ * the `CurlHandle` constructor guarantees that order.
  *
  * Setting the environment variable `KVIKIO_REMOTE_SHARE_DNS_CACHE` to a false value to disable
  * cache sharing and prevent this class from construction.
  */
-class CurlShare {
+class CurlShareHandle {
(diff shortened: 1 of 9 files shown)


cpp/include/kvikio/detail/curl_share.hpp
```diff
@@ -0,0 +1,77 @@
...
+namespace kvikio::detail {
+
+/**
+ * @brief A libcurl share handle holding a DNS cache for a group of curl easy handles.
+ *
+ * An attached share handle overrides the multi handle's cache. `curl_multi_add_handle()` installs
+ * its own only when the easy handle has none, so `CURLOPT_SHARE` has to be set first. Setting it in
+ * the `CurlHandle` constructor guarantees that order.
+ *
+ * Setting the environment variable `KVIKIO_REMOTE_SHARE_DNS_CACHE` to a false value to disable
```
bdice, line 21: Is there a known case where we would not want this? Seems like an optimization that might be better without a knob. Too many knobs is confusing to tune.

cpp/include/kvikio/detail/curl_share.hpp
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * @brief `CURLSHOPT_LOCKFUNC` callback. Takes the mutex guarding @p data.
+   *
+   * libcurl does no locking of its own, so without this callback the shared cache is accessed
+   * unsynchronized.
+   *
+   * @param handle The easy handle libcurl is serving. Unused. The mutex is chosen by @p data alone.
+   * @param data Which shared cache is about to be accessed.
+   * @param access Whether libcurl wants shared or exclusive access. Unused, since the DNS cache is
+   * locked exclusively.
```
bdice: I would think read-only access could be shared?

docs/source/runtime_settings.rst
```diff
@@ -103,6 +103,18 @@ The global budget is divided into an equal private share per reactor (``KVIKIO_R
...
 The even split assumes sub-ranges are spread across reactors, which holds under ``PER_CHUNK``. Under ``PER_PREAD`` all sub-ranges of one large :py:func:`kvikio.RemoteFile.pread` land on a single reactor, so that read is effectively limited to one reactor's share while the others stay idle.
 
+Shared DNS Caches ``KVIKIO_REMOTE_SHARE_DNS_CACHE``, ``KVIKIO_REMOTE_NUM_DNS_CACHES``
+--------------------------------------------------------------------------------------
+
+Let the easy handles in ``EASY_THREADPOOL`` share DNS caches to reduce lookups.
+
+Sharing is enabled by default. Set ``KVIKIO_REMOTE_SHARE_DNS_CACHE`` to ``false``, ``off``, ``no``, or ``0`` (case-insensitive) to disable sharing.
+
+``KVIKIO_REMOTE_NUM_DNS_CACHES`` sets how many DNS caches the process creates. The default value is ``16``, and a value below ``1`` is clamped to ``1``. Each worker thread is assigned one cache round-robin on first use, which means roughly ``KVIKIO_NTHREADS`` divided by this count threads share a cache.
```
bdice: Would it make more sense to have a number of caches that varies with the number of threads? Presumably the tradeoff is given at some level of threads contending for a mutex. So should this be “number of threads per cache”?

cpp/include/kvikio/detail/curl_share.hpp
kingcrimsontianyu: DNS cache sharing lets easy handles from different threads use the same IP address, whereas AWS recommends spreading the addresses (each name resolution gives a list of IP addresses) for throughput considerations. So there is a trade-off between limiting DNS lookups and spreading IP addresses, and DNS cache sharing is not necessarily an optimization if the thread count is low.

> Too many knobs is confusing to tune

Unfortunately there is a lot more to come in an experimental PR to max out throughput on larger instances. I think what we can do is to give a decent default that works in several common cases, and document clearly for users to tune in other cases where default does not apply.

cpp/include/kvikio/detail/curl_share.hpp
kingcrimsontianyu: Updated the doc for a bit of more clarity. Libcurl internally always passes `CURL_LOCK_ACCESS_SINGLE` to the callback for DNS cache, so the intention of lock is always exclusive. So `std::shared_mutex` won't help much here, and therefore `std::mutex` is used.

docs/source/runtime_settings.rst
kingcrimsontianyu: Yes. That makes sense. Reimplemented this part and updated the doc.

cpp/include/kvikio/detail/curl_share.hpp
wence-: If we resolve the list of ips once can we not just locally loadbalance the threads among those ips. Pr does a later lookup return new ips?

kingcrimsontianyu: Good idea.

In this experimental branch ([code](https://github.com/rapidsai/kvikio/pull/1059/changes#diff-af74b952ed1d995a0fc5ecf9b255637da943be8c6859e40f56b5b31a4319cf06R154))  `KVIKIO_REMOTE_DNS_SHUFFLE` is used with the hope of loadbalancing, and I did observe some performance improvement. However, the [libcurl doc](https://curl.se/libcurl/c/CURLOPT_DNS_SHUFFLE_ADDRESSES.html) states that:

> Addresses are not reshuffled if name resolution is completed using the DNS cache.

So with this setting threads sharing a DNS cache are still using the same IP address, and my observation may have just been a matter of statistical fluctuation. 

It is definitely worth exploring how to obtain the list of IPs and establish a pool visible to KvikIO threads, and a quick search seems to point to `getaddrinfo()`.

wence-: Yeah, I guess it's annoying but I had thought if one resolves the IPs up front we could pass curl IP addresses rather than DNS domain names.

cpp/src/detail/curl_share.cpp
```diff
@@ -0,0 +1,115 @@
...
+  auto* share_handle = static_cast<CurlShareHandle*>(userptr);
+  share_handle->_mutexes[static_cast<std::size_t>(data)].lock();
+}
+
+void CurlShareHandle::unlock_callback(CURL* /*handle*/, curl_lock_data data, void* userptr)
+{
+  auto* share_handle = static_cast<CurlShareHandle*>(userptr);
+  share_handle->_mutexes[static_cast<std::size_t>(data)].unlock();
+}
+
```
wence-, line 129: Is there a scenario in which curl gets an error after locking but before unlocking? It seems we could end up in a deadlock state if so. I imagine that if, for example, any cuda copy fails for some reason in the curl data receive callback

wence- (approved):
One query about lock release

cpp/src/shim/libcurl.cpp
```diff
@@ -106,14 +109,26 @@ CurlHandle::CurlHandle(LibCurl::UniqueHandlePtr handle,
   // Make requests time out after `value` seconds.
   setopt(CURLOPT_TIMEOUT, kvikio::defaults::http_timeout());
 
+  // Resolve a hostname once per DNS cache.
+  static bool const share_dns_cache = getenv_or("KVIKIO_REMOTE_SHARE_DNS_CACHE", true);
```
bdice: From my agent:

`KVIKIO_REMOTE_SHARE_DNS_CACHE` is read before checking `use_shared_dns_cache`. Therefore, a `MULTI_POLL` handle constructed with `false` still parses and permanently caches this setting. An invalid value can fail a `MULTI`-only request, and a later switch to `EASY_THREADPOOL` can use a value captured before DNS caches were first used, contrary to the documentation. Initialize the static setting only inside the `use_shared_dns_cache` branch.

bdice (approved):
A couple additional comments from my agent:

The PR body documents `KVIKIO_REMOTE_NUM_DNS_CACHES`, while the implementation and runtime documentation use `KVIKIO_REMOTE_MAX_THREADS_PER_DNS_CACHE` with different semantics. Update the PR description before merging.

No tests were added for enabling/disabling sharing, bypassing it under `MULTI_POLL`, or cache assignment limits. At minimum, the first finding should have a regression test covering a `MULTI` handle before the first `EASY` handle.

cpp/src/detail/curl_share.cpp
kingcrimsontianyu: I had Claude explore libcurl internals. The DNS share lock is not held when a write callback is run (where CUDA memcpy happens), so I think we are good in this regard.
But there will indeed be a deadlock if the DNS cache lock is held and a [debug callback](https://curl.se/libcurl/c/CURLOPT_DEBUGFUNCTION.html) throws. The debug callback is not used in KvikIO.

cpp/src/shim/libcurl.cpp
kingcrimsontianyu: Done. Restructured the code to allow unit tests.

PR #1052, opened 2026-08-30
Remote I/O egress controls and retryable transport failures

karthikeyann:
Draft. Two commits, both driven by measurements while reading a 3 TB TPCH dataset from S3 into GPU
memory on a `g7e.48xlarge` (4 ENA cards, 8 GPUs) through Velox/Presto.

## 1. `feat(remote): egress controls, streaming-store receive, and a retry-accounting fix`

Unchanged from `karthikeyann:s3-max-bandwidth`, rebased onto current `main`. Adds
`KVIKIO_REMOTE_IO_INTERFACE` (`CURLOPT_INTERFACE`, `if!<name>` / `SO_BINDTODEVICE`),
`KVIKIO_REMOTE_IO_DNS_SHUFFLE`, `KVIKIO_REMOTE_IO_NT_COPY`, `KVIKIO_REMOTE_IO_BUFFER_SIZE`,
`KVIKIO_REMOTE_IO_KTLS`, `KVIKIO_REMOTE_IO_DISCARD`, and fixes retry accounting so a partially
completed range transfer rolls the write-callback context back before being retried.

Independent confirmation of the egress control's value, measured end to end through Presto on
SF3000 TPCH: with all four cards on one subnet the kernel picks by route metric, so **every**
connection left through a single NIC.

| | without `KVIKIO_REMOTE_IO_INTERFACE` | with per-NIC binding |
|---|---|---|
| per-NIC receive | `[95.8, 0, 0, 0]` Gbps | `[68.8, 45.4, 81.4, 51.0]` Gbps |
| peak aggregate | 95.8 Gbps | **579.6 Gbps** |

One deployment note that cost some time to find: the interface name must exist in the process's
network namespace, so a containerised reader needs host networking. Without it `SO_BINDTODEVICE`
fails and the failure surfaces from `UrlParser::parse` as a *misleading*
`url.cpp: Unsupported URL scheme` on the original `s3://` URL, because
`encode_special_chars_in_path()` parses without `CURLU_NON_SUPPORT_SCHEME` and outside any
try/catch. Worth improving separately.

## 2. `fix(remote): retry transport-level curl failures (resets, DNS, partial body)`

`HttpRetryPolicy::is_retryable()` currently accepts only `CURLE_OPERATION_TIMEDOUT`, with a `TODO`
asking for more candidates. Every other transport failure is fatal, so a single dropped packet fails
an entire multi-TB table scan. Two failure modes hit repeatedly at these rates:

- `CURLE_COULDNT_RESOLVE_HOST` — the remote path resolves per request with no shared DNS cache, so at
  512 concurrent requests x 8 processes the query rate exceeds the AWS VPC resolver's 1024 packets/s
  per-ENI allowance.
- `CURLE_RECV_ERROR` — S3 recycles connections; observed 6-7 times per 22-query run.

Both were fatal, and in our stack the throwing worker then crashed, so one lost packet cascaded into
unrelated query failures. Measured impact: TPCH runs went from 17/22 and 14/22 queries passing (a
*different* set failing each time) to 22/22 in a single run.

This now also retries `CURLE_COULDNT_CONNECT`, `CURLE_SEND_ERROR`, `CURLE_PARTIAL_FILE` and
`CURLE_GOT_NOTHING`, and stops reporting `Got HTTP code 0` for failures that never carried an HTTP
status. It is only safe on top of commit 1's retry-accounting fix: `perform(on_retry)` rolls the
callback context back, so a retried partial transfer cannot corrupt the destination buffer.

## Not included

`KVIKIO_REMOTE_IO_KTLS` is retained from commit 1 but confirmed still inert: setting it left
`/proc/net/tls_stat` `TlsRxSw` unchanged, consistent with libcurl installing its own BIO while
OpenSSL requires the BIO under the `SSL` object to advertise kTLS on a real socket fd.

## Open questions for reviewers

- Should the retryable-transport-error set be configurable rather than hard-coded?
- Should `encode_special_chars_in_path()` pass `CURLU_NON_SUPPORT_SCHEME` so endpoint-probe failures
  stop masquerading as scheme errors?

### DIFF
--- a/cpp/src/detail/remote_callback.cpp
+++ b/cpp/src/detail/remote_callback.cpp
@@ -3,11 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <algorithm>
 #include <cstddef>
+#include <cstdint>
+#include <cstdlib>
 #include <cstring>
 #include <string>
 
 #include <curl/curl.h>
+#include <immintrin.h>
 
 #include <kvikio/detail/nvtx.hpp>
 #include <kvikio/detail/remote_callback.hpp>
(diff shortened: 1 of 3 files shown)


cpp/src/detail/remote_callback.cpp
```diff
@@ -3,11 +3,15 @@
...
 
+#include <algorithm>
 #include <cstddef>
+#include <cstdint>
+#include <cstdlib>
 #include <cstring>
 #include <string>
 
 #include <curl/curl.h>
+#include <immintrin.h>
```
madsbk, line 14: We would need to guard this to support ARM builds?

PR #1055, opened 2026-09-01
Remote batch API (1/n): Add shim for async batch memcpy

kingcrimsontianyu:
KvikIO provides access to CUDA API `cuMemcpyBatchAsync` through a shim `cuda_memcpy_async`, which permits only a single source and destination. This PR adds a general shim `cuda_memcpy_batch_async` to allow more than one sources and destinations.


madsbk (approved):
Looks good

wence- (approved):

PR #1056, opened 2026-09-01
Remote batch API (2/n): Implement transfer plan that coalesces and splits HTTP requests

kingcrimsontianyu:
This PR adds `detail::build_transfer_plan()`, which turns a batch of remote read requests into the HTTP range requests to issue.

Requests are grouped by file handle and CUDA context, merged when separated by at most `coalesce_max_gap` and fitting in `task_size`, or split into `task_size` chunks otherwise. The plan records where each fetched piece lands and how many transfers touch each request. No caller has been added yet.

The unit test (+477) is created by agent and reviewed by human.

This PR is tracked by https://github.com/rapidsai/kvikio/issues/1074


madsbk (approved):
Looks very good, have you done any benchmarking?

kingcrimsontianyu:
@madsbk Thanks! Not yet with this PR, which only creates a transfer plan and does not perform actual read. I'm working on the 4-th and 5-th PRs (tracked in https://github.com/rapidsai/kvikio/issues/1074), which provides the public C++ and Python API. Once the 5-th PR is prototyped, the plan is to first make microbenchmark in KvikIO, and then adopt the batch API in cudf-polars and Presto GPUs for TPC-H benchmarks.

PR #1057, opened 2026-09-01
Add C++ API reference to Sphinx docs

bdice:
## Summary

Integrate KvikIO’s Doxygen-generated C++ API into the Sphinx documentation using Breathe. This adds the C++ reference to the main documentation navigation and filters implementation-only headers from the generated API.


kingcrimsontianyu:
Verified locally. Looks good!

```bash
# In the devcontainer
(cd ~/kvikio/cpp/doxygen && doxygen Doxyfile) && sphinx-build -b dirhtml ~/kvikio/docs/source ~/kvikio/docs/_html -W
```

image

kingcrimsontianyu:
@bdice  Could you merge with `main` and resolve the conflict?

wence- (approved):

TomAugspurger (approved):

vyasr (approved):

kingcrimsontianyu (approved):

docs/source/python/index.rst
```diff
@@ -1,5 +1,5 @@
-API
-===
+Python API
```
mhaseeb123, line 1: AI pointed out that we need to preserve the existing `/api/` URL. Renaming this document moves the published page to `/python/`, breaking existing bookmarks and external links. Add a compatibility redirect or stub from `/api/` to `/python/`.

mhaseeb123 (approved):

docs/source/python/index.rst
bdice: I would accept the breakage here. We have a ton of URLs that are already breaking with other site migrations and I don't think this one alone is worth adding a dependency.

Docs using intersphinx references will not be broken by this move.

If we get user complaints, I can file a followup to use the `sphinx-reredirects` plugin.

PR #1062, opened 2026-09-08
Add to MULTI_POLL backend the execution-time dispatch pattern SHARED_QUEUE

kingcrimsontianyu:
This PR has 2 major parts:

- A third dispatch mode `KVIKIO_REMOTE_IO_REACTOR_DISPATCH=SHARED_QUEUE` is added to the `MULTI_POLL` backend. In this mode, sub-ranges wait in one pool-wide queue, and a reactor pulls one only after reserving a concurrency slot for it. This way the work binds to a reactor at execution time instead of at submission time. The mode requires a non-zero `KVIKIO_REMOTE_IO_MAX_CONCURRENT_REQUESTS` and falls back to `PER_CHUNK` without one. This mode improves TPC-H performance for both Polars and Presto, shown in the section below https://github.com/rapidsai/kvikio/pull/1062#issuecomment-5747204064.

- The reactor loop has now become quite complex, as it incorporates several key logic in a flat pattern, such as concurrency restriction, device copy, retry mechanism, shared queue dispatch, etc. This PR completely reorganizes the reactor loop for improved readability. Comments are modified by AI agent for better clarity.


kingcrimsontianyu:
## Performance result
- TPC-H SF-1K
- AWS EC2 g7e.12xlarge instance

### cudf-polars
```text
- 64 MiB task size
- Arguments:
  --frontend spmd
  --max-concurrent-io-tasks=8 
  --pinned-memory
  --pinned-initial-pool-size 68719476736 --rapidsmpf-statistics
  --no-print-results --no-collect-traces --no-summarize all
+-----------------+----------------------+--------+--------+------------+------------+
|     backend     |    configuration     | iter-1 | iter-2 | i1 vs easy | i2 vs easy |
+-----------------+----------------------+--------+--------+------------+------------+
| easy_threadpool | 384 threads          |  159.6 |  137.3 |      +0.0% |      +0.0% |
| SHARED_QUEUE    | 24 threads, conn 384 |  153.4 |  137.7 |      -3.9% |      +0.3% |
| PER_CHUNK       | 24 threads, conn 384 |  167.6 |  151.5 |      +5.0% |     +10.3% |
| PER_PREAD       | 24 threads, conn 384 |  166.0 |  150.2 |      +4.0% |      +9.4% |
+-----------------+----------------------+--------+--------+------------+------------+
```

### Presto GPU

```text
- WXD-launch velox and presto branch
- This KvikIO branch
- 2 max drivers
- driver.max-split-preload=6
- 64 MiB task size
+-----------------+----------------------+--------+--------+------------+------------+
|     backend     |    configuration     | iter-1 | iter-2 | i1 vs easy | i2 vs easy |
+-----------------+----------------------+--------+--------+------------+------------+
| easy_threadpool | 384 threads          |  120.2 |  109.4 |      +0.0% |      +0.0% |
| SHARED_QUEUE    | 48 threads, conn 384 |  121.6 |  109.6 |      +1.2% |      +0.2% |
| PER_CHUNK       | 48 threads, conn 384 |  125.7 |  113.5 |      +4.6% |      +3.8% |
| PER_PREAD       | 48 threads, conn 384 |  126.1 |  113.1 |      +4.9% |      +3.4% |
+-----------------+----------------------+--------+--------+------------+------------+
```

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -216,255 +217,307 @@ void MultiPollReactor::submit(std::vector<std::unique_ptr<RemoteMultiTransfer>>
...
+  // The request is on the wire from here, so this is where the transfer's own span starts.
+  // Everything before it was queueing, in the inbox or behind the gates.
+  transfer->physical_recorder.emplace(
+    transfer->physical, transfer->file_offset, transfer->ctx.size);
+  _in_flight.emplace(easy, std::move(transfer));
+  return true;
+}
+
+void MultiPollReactor::admit_from_pool(AdmitPass& pass)
+{
```
kingcrimsontianyu, line 320: This function is the key part of this PR. Under the `SHARED_QUEUE` mode, the work assignment is now done at execution-time instead of submission-time. The queue wants to hand a total of `share=queue-size / reactor-count (rounded up)` works to each reactor, and the actual number of works accepted by a reactor depends collectively on the value of `share`, the reactor-private concurrency cap, and the per-(reactor, cuda context) bounce buffer cap.

cpp/include/kvikio/detail/multi_poll_reactor.hpp
```diff
@@ -381,17 +500,74 @@ class MultiReactorPool {
    */
   void signal_death(std::exception_ptr eptr) noexcept;
 
+  /**
+   * @brief Take one sub-range off the pool-wide queue, or nothing if it is empty. Thread-safe.
```
madsbk, line 504: Anything here not thread-safe? I so , make them as not thread safe instead?

cpp/src/detail/multi_poll_reactor.cpp
```diff
@@ -216,255 +217,307 @@ void MultiPollReactor::submit(std::vector<std::unique_ptr<RemoteMultiTransfer>>
...
+    } catch (...) {
+      // Keep the transfer reachable for `fail_all_pending()` to resolve its aggregate.
+      _pending.push_back(std::move(transfer));
+      throw;
+    }
+    if (!admitted) {
+      // Refused a bounce buffer, and `try_admit` has dropped its slot. Back to the head of the
+      // queue for whichever reactor can start it.
+      _pool->return_to_queue(std::move(transfer));
+      return;
```
madsbk, line 355: With concurrent `pread()` calls of different kinds, this can block the queue even when work the reactor could start sits right behind. One `pread()` is all one kind, so it takes host mixed with device, or two CUDA contexts, to hit.

`return_to_queue()` pushes to the front and `admit_from_pool()` then `return`s. So when the head is a device sub-range whose context has no free bounce buffer on this reactor, the reactor stops pulling for the rest of the pass and everything behind the head waits with it, including host sub-ranges that need no buffer at all.

Could you set refused sub-ranges aside and keep pulling, instead of ending the pass? Or just a comment noting the stall is intentional and bounded.

cpp/include/kvikio/defaults.hpp
```diff
@@ -536,6 +538,12 @@ class defaults {
    * approximate: it rounds down when the value is not a multiple of the reactor count, and up when
    * it is smaller than the reactor count (each reactor is floored to at least 1).
```
madsbk: Stale, we now now spreads the remainder to keep the total exact, right?

madsbk: Same staleness in `cpp/include/kvikio/detail/concurrent_request_limiter.hpp`, I think.

madsbk (commented):
First pass, looks good.

- Now that `SHARED_QUEUE` wins on the benchmarks, would it make sense to retire `PER_CHUNK` and `PER_PREAD` and keep a single dispatch path?
- A dedicated `SHARED_QUEUE` test would be good. Maybe run `test_http_io.py` under `SHARED_QUEUE` in a subprocess with the env var set, since the pool captures the mode at first use?

cpp/include/kvikio/detail/multi_poll_reactor.hpp
kingcrimsontianyu: Done. Removed the thread safe remark from the member function's comment, and instead added it to the class-level comment.

cpp/src/detail/multi_poll_reactor.cpp
kingcrimsontianyu: Yes. That's a legitimate concern. I raised the bounce buffer cap to twice the request slot cap in this PR (https://github.com/rapidsai/kvikio/pull/1062/changes#diff-f0d7e2e881b74a4d4e8271608d4e404dcefb2d261d1939c8811ef90fb7d21fa0R175) to address the host + device mix cases, but the stall might still occur if device copy is significantly slower than network (not often to see but may happen).

I've updated the comment on this, and added a TODO for the "queue per context" solution.

cpp/include/kvikio/defaults.hpp
kingcrimsontianyu: Thanks. Fixed.

kingcrimsontianyu:
> Now that SHARED_QUEUE wins on the benchmarks, would it make sense to retire PER_CHUNK and PER_PREAD and keep a single dispatch path?

Yes. I plan to deprecate and remove the other dispatch options in a future PR. Probably the property setter/getter can all be gone as well.

kingcrimsontianyu:
> Maybe run test_http_io.py under SHARED_QUEUE in a subprocess with the env var set, since the pool captures the mode at first use

Done! Needed a small change to the multi-poll overflow error message.

madsbk (approved):
Thanks @kingcrimsontianyu

